PR #91, opened 2016-09-26
Tika pdf text

jbothma:
hackedly added in haste

deals with PDFs that pdfminer take forever on like http://code4sa-gazettes.s3.amazonaws.com/archive/ZA/2016/government-gazette-ZA-vol-614-no-40208-dated-2016-08-12.pdf

try with the following in docker-compose

```
  tika:
    image: logicalspark/docker-tikaserver

  worker:
      links:
        - tika
      environment:
        TIKA_URI: 'http://tika:9998/'
        PDF_TEXT_MODULE: 'tika'
```


jbothma:
seems to work well https://search.opengazettes.org.za/collections/24 (email me for access if you can't view it)

setup.py
```diff
@@ -31,11 +31,11 @@
             'html = aleph.ingest.html:HtmlIngestor',
             'djvu = aleph.ingest.djvu:DjVuIngestor',
             'img = aleph.ingest.image:ImageIngestor',
-            'email = aleph.ingest.email:EmailFileIngestor',
-            'mbox = aleph.ingest.email:MboxFileIngestor',
+            'email = aleph.ingest.email_:EmailFileIngestor',
```
jbothma: This is about requests looking for the builtin email module and finding the aleph email module. Apparently one doesn't name modules with the same name as builtin modules in python. Forget namespacing.

http://stackoverflow.com/questions/28685175/cannot-use-python-requests-module-if-file-is-named-email-py-or-logging-py

If I told you this was an adventure, I wouldn't be lying.

longhotsummer: Oh. My. God.

But why does this pull request need to fix this? Was it never working?

jbothma: I don't know what in aleph used requests before or whether it worked - I guess no ingestor was using `requests`, but now something in aleph.ingest is (the code I added) which means the email module is picked up incorrectly.

pudo: Perhaps this could be solved with ``absolute_imports``?

pudo:
Merging this for now, thanks for this important new feature. I wonder if we should consider making it it's own extractor that can override the native versions for all supported Tika data types?

PR #96, opened 2016-09-29
Match headers with rest of site

longhotsummer:
* also add SEO tags


longhotsummer:
Aaargh! Stupid default for merge requests.

PR #100, opened 2016-10-04
Indicate signing in will allow you to set an alert

longhotsummer:
Our users thought this functionality wasn't yet enabled.

This patch also prevents .swp files from being sucked in as assets.


longhotsummer:
Perhaps the "Sign in to set an alert" should behave just like the "sign in" link and take them through that flow. Then they'd come back to this page and can click 'set alert'.

pudo:
Cool PR. I'll add the link

longhotsummer:
If my Angular is crap, please change it. I'm not familiar with it at all.

PR #101, opened 2016-10-06
Support for Google Analytics

### LINKED ISSUE #98: Google analytics
Just wondering if we could get proper GA integration with Angular? Might be as simple as leaning on [Angulartics.](http://angulartics.github.io/)

longhotsummer:
We have to explicitly tell Angulartics to send a pageview when the search term changes in /search. I don't know if we should be doing the same thing in other views (eg. collections search?) or why we need to do that explicitly.

Closes #98.

### DIFF
--- a/aleph/templates/layout.html
+++ b/aleph/templates/layout.html
@@ -134,7 +134,6 @@
       })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
       ga('create', '{{ config.GA_TRACKING_ID }}', 'auto');
-      ga('send', 'pageview');
     </script>
     <!-- End Google Analytics -->
     {% endif %}
--- a/aleph/templates/layout.html
+++ b/aleph/templates/layout.html
@@ -125,6 +125,20 @@
       </div>
     </footer>
 
+    {% if config.GA_TRACKING_ID is defined %}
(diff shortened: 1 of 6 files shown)


pudo:
I'm not allowed to use GA in my env; is there any way you could also put this behind a config part, at least the template integration?

longhotsummer:
If you don't declare GA_TRACKING_ID in the config, the GA script is not injected at all (and the angular code just does nothing). That's why the example config file has it commented out. Is that sufficient?

longhotsummer:
Does this cover your use case, @pudo ?

pudo:
Sorry I still wanted to check out if there's a way this can be made to support Piwik also. I'm not allowed to use GA (or load their assets -- which this currently does?), but more fine-grained analytics would be great.

aleph/templates/layout.html
```diff
@@ -125,6 +125,19 @@
       </div>
     </footer>
 
+    {% if config.GA_TRACKING_ID is defined %}
```
longhotsummer: This only loads the GA assets if the tracking ID is defined.

longhotsummer:
This doesn't load GA assets unless a GA tracking ID is defined.  Don't know about Piwik.

Issue #52, opened 2016-07-07
User registration and password-based login


pudo:
@danohuiginn to submit PR :)

pudo:
hey @danohuiginn -- now that aleph.openoil.net has been born to great acclaim, any chance I can tease you into backporting this into master :)

longhotsummer:
Yes please!

pudo:
Note to self: folks at Süddeutsche Zeitung would like to see integration with AD/LDAP here, which should not be too hard once we've got a password prompt in the first place. 

/cc @vwormer

stas:
This should be ready for a review. I'll add LDAP support in a separate branch.

stas:
Both, support for stand-alone users and LDAP support was merged.
I'm closing this.

PR #106, opened 2016-10-10
Support for multiple OAUTH providers

longhotsummer:
This updates the OAUTH support to handle multiple providers. The login endpoint changes to `/api/1/sessions/login/<provider>` and the callback URL changes similarly. There is built-in support for Facebook and Google, and it's very easy to add support for other providers.

NB: the callback URL entered into the Google Developer Console must be updated.

When the user clicks the login dialog they get a modal popup giving them a choice of login providers.

This doesn't yet include full support for Twitter, because I don't yet have a functioning Twitter OAUTH setup.

sign-in

### DIFF
--- a/aleph/oauth.py
+++ b/aleph/oauth.py
@@ -1,0 +1,34 @@
+from flask_oauthlib.client import OAuth
+from flask import session
+
+oauth = OAuth()
+
+
+def get_oauth_token():
+    if 'oauth' in session:
+        sig = session.get('oauth')
+        return (sig.get('access_token'), '')
+
+
+def setup_providers(app):
+    providers = app.config.get('OAUTH', [])
+    if isinstance(providers, dict):
+        # support for legacy single provider
+        providers = [providers]
+
+    for provider in providers:
+        name = provider.pop('name')
+        # legacy provider
+        if name == 'provider':
+            name = 'google'
+        label = provider.pop('label', name.capitalize())
+
+        provider = oauth.remote_app(name, **provider)
+        provider.label = label
+        provider.tokengetter(get_oauth_token)
+
+
+def configure_oauth(app):
+    setup_providers(app)
+    oauth.init_app(app)
+    return oauth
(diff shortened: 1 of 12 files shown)


aleph/oauth.py
```diff
@@ -0,0 +1,65 @@
...
+        return
+    me = provider.get('userinfo')
+    user_id = 'google:%s' % me.data.get('id')
+    role = Role.load_or_create(user_id, Role.USER, me.data.get('name'),
+                               email=me.data.get('email'))
+    session['roles'].append(role.id)
+    session['user'] = role.id
+
+
+@signals.handle_oauth_session.connect
```
longhotsummer: We need the equivalent code for Twitter.

aleph/views/sessions_api.py
```diff
@@ -60,13 +62,23 @@ def status():
...
             'write': authz.collections(authz.WRITE)
         },
-        'login': url_for('.login'),
-        'logout': url_for('.logout')
+        'logout': url_for('.logout'),
+        'providers': providers,
     })
 
 
 @blueprint.route('/api/1/sessions/login')
```
longhotsummer: This isn't technically necessary, we could remove it.

pudo:
Looks very interesting, thanks much for the PR. Two thoughts:

a) Would it be possible to make it so the selector is skipped if there's only one provider configured (background: in our deployment here at OCCRP, the app runs against an internal OAuth provider and we can't allow people to pick), and

b) should the login perhaps be a proper route (e.g. ``/login``) so that we can put #52 there? On that note: @danohuiginn -- do you think it's worth back-porting the OpenOil code for password-based logins, too?

longhotsummer:
I've adjusted it so that if there's only on oauth provider, it goes there directly.

I now also use triggerLogin, rather than a new global method.

Regarding `/login`, I don't mind either way. I'm not sufficiently familiar with Angular to know about popping up modals based on routes.

longhotsummer:
The tests are failing because the app is being re-initialised for each test, but without the previous app being torn down correctly. Any thoughts?

longhotsummer:
We have passing tests! This is ready to merge from my side.

longhotsummer:
Can we merge?

pudo:
Thanks very much for this pull, looks excellent.

PR #121, opened 2016-12-12
Integrate graph dataset browser

### LINKED ISSUE #115: Integrate memorious spike into Aleph
A graph indexer prototype exists here: https://github.com/occrp/memorious 

After developing this for three weeks, it is obvious to me that it could work as part of aleph, adding an improved entities component based on bulk data. This requires integrating the existing code and removing duplicate components.

pudo:
Fixes #115. This is of course not complete, but keeping the branch open longer seems precarious. At the moment, the following large changes are included:

* Adopt the data model used in memorious for all entities.
* Make entity<->collection and document<->collection 1:1 relationships, no more sharing. 
* Support mapping-based loads of raw datasets from remote SQL
* Consolidate data parsing functions for domain data (e.g. phones, emails, etc.)
* Use role-based authorization for datasets.
* Display databases in UI upon demand (hidden by default): /datasets
* Fix up entity tagger.

Missing parts vs. memorious include: 

* Cross-referencing support (should this be vis-a-vis collection entities only?)
* Link browser doesn't have search, pagination or facets
* Duplicates display on entity pages
* Links to searches in entity properties display.


stas:
@pudo I'll update #112 since the schemata changed.
Btw, if we don't need this branch, we can delete it I guess.

pudo:
@stas thanks. There's a much larger documentation task around the graph modelling thing, but that's out of scope for your sprint, I think (unless you crave reverse-engineering the mapper code ;). I'll delete the memorious branch now.

I'm sorry for interfering with #112, if you were to pull it up I would owe you eternal cookies. Really messed up the merge order on that one.

stas:
It's all good, I was actually considering to hold that merge and add some tests once we front-end is split from the back-end and we could run front-end tests aside.

Issue #119, opened 2016-12-09
Make a management command for deleting pending entities


pudo:
Either per collection or for the whole system. Refs #118.

stas:
Handled by #122

PR #118, opened 2016-12-05
Sphinx docs

stas:
This is a WIP branch on the documentation review and migration to Sphinx.
I will take some more time to review what is currently in our wiki and add relevant pages to the `docs`.

A preview of this is currently available at:
http://aleph.readthedocs.io/en/latest/

I will let you know when a final-ish review is needed.

PR #122, opened 2016-12-14
Docs missing cmds

stas:
This needs to be merged only after #118 is closed.

It covers a note from the wiki related to #119 
https://github.com/pudo/aleph/wiki/Maintenance#deleting-all-pending-entities

I tried adding a test for this change and it looks like all the fixtures require manual composition.
I suggest using factories and fake generators to build these for further tests.
This is superior since it abstracts the interaction with models to the minimum necessity and brings a decent randomness aspect that can make the tests more solid.

@pudo let me know what are your thoughts on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,9 @@ addons:
 env:
   global:
     - NODE_VERSION=6.9.2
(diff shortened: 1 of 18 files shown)


stas:
Alright this was updated and can go through a second review.

PR #132, opened 2017-01-30
Credentials

stas:
Split some of the existing role data into credentials model.
This layout and authentication mechanism should be flexible enough for us to pull-in additional authentication _protocols_ (LDAP is a good example, #52).

I'll prepare a separate PR for the front-end functionality.

@longhotsummer if you have a minute, it would be great to have your :eyes: on this and let us know how these changes will work for you.

@pudo from #52 I understand we want user registration to be optional?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ install:
 before_script:
   - sleep 10
   - psql -c "create database ${ALEPH_DATABASE_NAME}_test;" -U postgres
+  - psql -c "create extension pgcrypto;" -U postgres ${ALEPH_DATABASE_NAME}_test
   - elasticsearch --version
   - nosetests --version
   - psql --version
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ assets-dev: assets
 test:
 	PGPASSWORD=aleph psql -h postgres -U aleph -c 'drop database if exists aleph_test;'
 	PGPASSWORD=aleph psql -h postgres -U aleph -c 'create database aleph_test;'
+	PGPASSWORD=aleph psql -h postgres -U aleph aleph_test -c 'create extension pgcrypto;'
 	nosetests --with-coverage --cover-package=aleph --cover-erase
 
 base:
--- a/aleph/migrate/versions/2d230e6ce46d_added_credentials.py
+++ b/aleph/migrate/versions/2d230e6ce46d_added_credentials.py
@@ -65,8 +65,9 @@ def upgrade():
     creds_table = meta.tables['credential']
 
     credentials = []
+    roles_query = sa.select([role_table]).where(role_table.c.type == 'user')
 
-    for role in bind.execute(sa.select([role_table])).fetchall():
+    for role in bind.execute(roles_query).fetchall():
         credentials.append(
             dict(
                 id=make_textid(),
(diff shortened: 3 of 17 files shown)


longhotsummer:
Looks fine to me, FWIW.

Does the UI around choosing a login provider need to change? Do you need to add a Password login provider, or hardcode that into the UI?

stas:
> Does the UI around choosing a login provider need to change? Do you need to add a Password login provider, or hardcode that into the UI?

I suggest making the password login as the default authentication strategy. But I think it should be pluggable/optional. There's also some work we'll be doing to provide the LDAP/AD support too.

pudo:
Having passwords by default would be cool, but it implies having quite a bunch of additional functions: registration, password reset, etc. I guess we need to build these out first before we can make the transition.

stas:
> Having passwords by default would be cool, but it implies having quite a bunch of additional functions: registration, password reset, etc. I guess we need to build these out first before we can make the transition.

There is #147 for that and also the LDAP in a separate branch.

pudo:
Ok. First off: I really fucked up here, should have not just read this but actually deployed it to a staging instance before merging. 

However, it seems to be the following needs to be fixed before we can revive this:

a) We need to make sure the system roles, and OAuth groups, are ported correctly. When running this on a migrated database, the system roles mechanism is currently completely broken, and groups just get deleted. Users have been disconnected from their OAuth identity, even though the code for that seems to exist. 

b) Let's not do DB crypto. Reliable code exists in werkzeug.security. 

c) And this is really late: do we actually, actually need the credentials table? Are we likely to every do that thing where you have N sign-in methods for a single account? Wouldn't two new fields (password, reset_token) on the role table be a simpler solution?

stas:
Alright, let me see what happened. It is important though to have the rest of the #147 merged or at least reviewed before we make final decisions.

So far **a)** I added migration support here:
https://github.com/pudo/aleph/pull/132/files#diff-01063ebbd4e63b60539db518f1b77a44R72
I'm not sure how exactly this would break the groups, though. I'll take another look.

In regards to **b)**, it's your call. Credentials do use bcrypt right now which is a solid algorithm, but I'll change it to werkzeug.security upon your advice.

**c)** we can simplify this down to a couple of columns, but this would mean that users logging in with an OAUTH and then with an LDAP account might end up with two separate accounts. Could you explain a bit how you see this? Should we just add extra columns for any potential authentication type to record foreign IDs or just set the user session and forget about the foreign IDs (except for the password type)?

pudo:
Thanks! Regarding b) I'd just prefer the werkzeug solution because it doesn't require the pgcrypto module (a potential error source), while we already have all of the Python things installed. 

c) I understand the benefit, but it just feels a little framework-without-usecase to me. I think sign-in methods are fairly limited, and I'm happy to say that an account is tied to a specific login method (plus perhaps password, which doesn't rely on foreign_id). From what I can tell, at the moment we don't actually have merging support, anyway, ie. if I sign into an account first with a registration and then via Oauth, the system wouldn't know how to reconcile them -- am I missing something here?

stas:
> Thanks! Regarding b) I'd just prefer the werkzeug solution because it doesn't require the pgcrypto module (a potential error source), while we already have all of the Python things installed.

Sounds good. I don't want to get into details, but to avoid any confusion, the PG implementation of bcrypt is actually very solid and more portable compared to what werkzeug offers.


> I understand the benefit, but it just feels a little framework-without-usecase to me.

My use-case was very simple: multiple logins from different sources, hence the one-to-many entity. Merging support is automatic and works as long as the email is the same. Since this is not the case, I'll try to come up with something more simple.

PR #150, opened 2017-02-15
Allow overriding celery options from config file

longhotsummer:
This also takes into account that Celery now wants its options to be lowercase, whereas the old `celery.conf.update(app.config)` gave it way too much info and tended to use uppercase.


pudo:
Looks good to me. Just out of interest: which overrides are you guys currently using? I've had so many in place now, but somehow most of them turned out to be blankets put in place to fix up architecture bonfires, so it'd be good to know!

Issue #163, opened 2017-03-06
aleph module is missing on local development with docker-compose


wpf500:
This normally first manifests itself as the error "No ingestors found" (e.g. https://github.com/pudo/aleph/issues/135)

The build process for the [pudo/aleph](https://hub.docker.com/r/pudo/aleph/) image clearly does [install the module](https://github.com/pudo/aleph/blob/master/Dockerfile#L12), the proof is:
```bash
$ docker run pudo/aleph ls /aleph/aleph.egg-info
PKG-INFO
SOURCES.txt
dependency_links.txt
entry_points.txt
namespace_packages.txt
not-zip-safe
top_level.txt
```

However, when you follow the `docker-compose` instructions, its no longer there, as shown:
```bash
$ docker-compose up
$ docker-compose run app ls /aleph/aleph.egg-info
ls: cannot access /aleph/aleph.egg-info: No such file or directory
```

This is because [docker-compose.yml](https://github.com/pudo/aleph/blob/master/docker-compose.yml) mounts your local repo to `/aleph` in the container, overwriting the `/aleph` path created in the [Dockerfile](https://github.com/pudo/aleph/blob/master/Dockerfile). The other build commands (various `pip installs` etc.) only have effect because they install files outside of that path. Obviously for developers who had previously run `pip install -e .` on their machines, this problem wouldn't show up.

Therefore, adding a step to run `pip install -e .` locally in the developer installation guide will fix this. It does however makes me wonder why Python requirements aren't just bundled into pudo/aleph-base and the containers then based off that? Or I guess the project should use a virtualenv and developers then install the requirements locally too.

wpf500:
Actually, you'd have to run `pip install -e .` inside the docker container, otherwise it'll try to install lots of things globally on the local machine

stas:
@wpf500 thanks for digging into this.

It looks like a very :snake: issue. Apparently the `.egg-info` is what the development mode does in order to tell python the package is installed. I went through their docs, and it looks like _this is the way it works_™.

I actually asked @pudo previously whenever we do actually need a `setup.py`, and this issue brings this question back on the table. Currently the `setup.py` is not actually used for it's original purpose (requirements.txt is still handling the deps). And I still think we have no reasons to keep the `setup.py`, @pudo if we get the ingestors plugged the other way around?

pudo:
[...]

Issue #146, opened 2017-02-12
Fully split development and deployment config


pudo:
At the moment, we're getting a bunch of issues resulting from bugs related to the semi-docker development setup. This means it's virtually impossible to actually deploy Aleph, either in production or in development. Here's what I'd like to do:

* Make sure the deployment configuration is separate from the development configuration, and make sure it is very stable. A deployment install should be as simple as pulling a particular github tag, and running ``docker-compose up``. 
* Go back to recommending developers perform a docker-less install. The semi-docker route has proven itself brittle, and breaks the Makefile and various other settings for those who do have a native install.

What do you think, @stas?

stas:
There was previously some related discussions in #130 about this.

I did a little bit of research on this topic, and Docker peeps suggest providing two separate compose files as the best practice (which kinda' makes sense).

What I am worried about though, is the `aleph-base` image that keeps growing and requires significant time to rebuild. It might be better if we keep maybe python apart from nodejs and have separate containers running one and the other. As for dev/prod, we already started splitting requirement files, and I think we could just leverage that to keep things prod-ready.

My experience so far with Docker is definitely not a :heart: story. But I try to imagine providing the cross-platform support for the installation procedures outside the Docker, and It gives me :cold_sweat:.

So I'd pick any solution that will take us as little time as possible and let us focus on our real tasks :)

pudo:
@stas -- I've finally merged docker_settings and default_settings in the commit above. Would appreciate a sanity check.

I'm still a little torn regarding the best route forward on the developer setup. Have been using a native install on OS X for a while, and while getting the dependencies installed is a good chunk of work, the resulting install is really smooth, and I don't have to block off 4GB of memory for docker (memory allocation on the Mac is static). On the other hand, I see the appeal of just putting it all in a container and being done with it.

Regarding a separate Node.js container: excellent idea. Let's do it as we get into the new UI.

stas:
@pudo is it possible we just get rid of the `default_settings` somehow?

`flask.config.get_namespace` somehow already does a part of the job. My biggest concern is that we no longer load the values from the `os.environ` directly. This means we need to copy environment directives for any flask plugin that might need some (my current use-case is related to `flask-simpleldap`).

stas:
A note on something worth keeping in mind, the `aleph.datasets` is using a configuration directive `SCHEMA_YAML` that uses physical file(s) to load dataset mappings. The way it is right now, requires probably mounting a volume with the relevant files.

[...]

Issue #130, opened 2017-01-27
requirements.txt doesn't pin all dependencies


sihil:
Trying to run `1.1` currently doesn't work as the Jinja2 `2.9.x` release has broken the asset management (see https://github.com/miracle2k/webassets/issues/477).

This seems to be a result of some of the dependencies (Jinja2 in this case) not being pinned.

stas:
@sihil we got rid of webassets in #123, but you're right we should pin all possible dependencies. :neutral_face:

sihil:
My colleague tried to bring up using docker-composer on the master branch a couple of weeks ago but that didn't work (for other reasons), so we reverted to using the most recent 'stable' release. Also thrown by two sets of, sometimes conflicting, documentation at http://aleph.readthedocs.io/en/latest/ and https://github.com/pudo/aleph/wiki.

stas:
@sihil http://aleph.readthedocs.io/en/latest/ is the latest documentation. We've migrated and updated most of the wiki (should be pretty much all in fact).

We received two issues recently regarding the installation procedures, but the authors never reported back #125 #126 

Could you tell us what was not working using the latest documentation?

@pudo do you think we could deactivate the wiki for now? The content should not be deleted afaik.

sihil:
The app container doesn't come up cleanly:
```
app_1            | touch aleph/static/style/_custom.scss;
app_1            | (test -f '' && cp -f  aleph/static/style/_custom.scss) || exit 0
app_1            | ./node_modules/webpack/bin/webpack.js --env.prod
app_1            | make: ./node_modules/webpack/bin/webpack.js: Command not found
app_1            | Makefile:14: recipe for target 'assets' failed
app_1            | make: *** [assets] Error 127
```

For users whoa re intending to experiment with aleph, it would be more robust to have a docker-compose that used completely built containers and did not depend on anything build built at startup. Obviously a different use case compared to development. But I'm a few days in and have yet to index any content (the default crawler configured in 1.1 doesn't seem to work and doesn't seem to display or log any errors).

stas:
@sihil I'm sorry you need to get through all of it. Let me see if I can help somehow.

So what you are suggesting is to have a docker-compose file similar to what we have now, but without the local code-base mounted as a [volume]?

@pudo I think this makes sense, I could review what we have now and come up with something.

[volume]: https://github.com/pudo/aleph/blob/master/docker-compose.yml#L50

sihil:
[...]

Issue #126, opened 2017-01-06
No ingestors found (fresh docker install)


cdharris:
Hi @pudo ! This software looks awesome and I've been trying to test it out, I've hit a problem after running a fresh `docker-compose up`. Would be very grateful for your time if you could point me in the right direction.

The app and services are started and I've edited the `aleph.env` with appropriate config and run the migrations as instructed. 

When I login and try and upload a pdf document to a collection, I receive the error
```
Traceback (most recent call last):
  File "/aleph/aleph/ingest/ingestor.py", line 109, in dispatch
    best_cls = cls.auction_file(meta, local_path)
  File "/aleph/aleph/ingest/ingestor.py", line 102, in auction_file
    meta.mime_type))
IngestorException: No ingestor found: u'test-document.pdf' (pdf, application/pdf)
```

it seems that `get_ingestors()` is returning empty `[]` even though I can see in `setup.py` the ingestors specified where they seem like they should be:
```
 entry_points={
        'aleph.ingestors': [
            'skip = aleph.ingest.dummy:SkipIngestor',
            'pdf = aleph.ingest.text:PDFIngestor',
            'doc = aleph.ingest.document:DocumentIngestor',
            ...
```

Q1. do you think this could be something to do with the application enviroment paths / the app not finding this list in the `setup.py` ?

Q2. Any advice on how to debug this app when running in Docker? I've tried editing the source on my host but it doesn't seem to be picked up by the app/worker running in the docker container, not sure if this is due to the mounted Docker volume or the setup of the image, or the python app running in memory, could you share any workflow tips here? 

Thanks a load,
Chris

stas:
Hey @cdharris,
thanks for the detailed report.
Could you confirm you did a fresh build of the docker images before you start the containers?
We recently did some changes to the way the installation is done, a full guide is available here:
http://aleph.readthedocs.io/en/latest/installation.html#development-installation-steps

In regards to the error you get, this happened to me before and it worked after removing any old containers and images and running a fresh build. In the meantime I could try to reproduce the error by running a fresh installation too if that helps.

cdharris:
Hi @stas I can confirm this issue was following a fresh build.

To make sure, I've just done a completely fresh build again (clean VM, clean docker installation, fresh git clone and following your build instructions)

To get the app and UI working there's two 'fixes' I had to include on both run throughs of the fresh install.

1. npm dependencies not available
```
aleph-docker-01 aleph # docker logs aleph_app_1
[...]

PR #167, opened 2017-03-09
[Do not merge] Initial build/docker changes

wpf500:
This PR is a work in progress, most things still don't work, but I'd love people's thoughts/comments.

There are currently at least three different places that do some sort of building (i.e. `pip install`, etc.):
1. Dockerfile: building `pudo/aleph`
1. travis.yml: building and running tests
1. Makefile: building for development

This PR is an attempt at simplification. All build related commands should be in the Makefile and other things use it. The aim is to have two Docker configurations:
- docker-compose.yml: "Just works" out-of-the-box, based on the `pudo/aleph` image. Users only need docker-compose.yml and an aleph.env, no building necessary (could even go in its own repo?)
- docker-compose.dev.yml: A developer's local checkout is built and mounted into `pudo/aleph-base`.

Currently, each service needs to install its dependencies separately. This is a problem because:
1. It takes ages
1. Other dependencies (Node etc.), are already local, so they end up overwriting each other.

Two possible solutions:
1. Have one service, run worker/beat/app inside it. Basically how it would be done if the developer chose not to use docker (my preferred option)
1. Build once on the developer's machine, then mount. We'd have to move to `virtualenv` for the Python dependencies, I think all other things are already locally installed.

Thoughts? @stas @pudo 

## TODO
- [ ] Use docker-compose extends functionality for shared services (elasticsearch, postgres, rabbitmq)
- [ ] Decide how to run developer builds
- [ ] Travis building/testing
- [ ] Test it works
- [ ] Docs

Makefiles support dependency rules, (i.e. `command: dependency`), but I don't know how feasible it is to make it recognise whether or not it needs to run the dependency, e.g. only pip install if the Python requirements aren't there

At the moment the README will say run `make install` on clone, but it would be nice if you could run `make app` and it automatically installs/updates dependencies.


stas:
@wpf500 thanks for taking a look at this.

I'll raise some concerns that might not have been documented/stated on how things work so far.

> Currently, each service needs to install its dependencies separately. This is a problem because:
> 
> It takes ages

This is indeed an issue. But this is how Docker works. The most we could do and already suggested is to try splitting the front-end dependencies into separate Docker containers (say provide an image for working on the front-end aka Node.js stuff, and one for working on the back-end aka Python stuff).

The reason splitting (say app from workers) services is not an option, is because the codebase requires the packages to be installed before starting any of the services. Some packages just refuse to install if the system libraries are not present on the system. This is another reason we started working on ingestors and making those more pluggable. See: https://github.com/alephdata/ingestors/pull/1

> Other dependencies (Node etc.), are already local, so they end up overwriting each other.

What do you mean by local?
The reason we added Node.js support to the image is to make sure we are all working with the same Node.js version, along with the same Webpack and etc. If we start mixing these we might end up breaking the builds for the front-end.

> Two possible solutions:
> 
> Have one service, run worker/beat/app inside it. Basically how it would be done if the developer chose not to use docker (my preferred option)

This was mentioned before, and I think this should not be our concern. I don't think we should provide support for every developer on any platform. With all the overhead of the Docker (I'm not saying it is a great experience so far), but at least we have one set of instructions we can follow.

Why local installation, sans-Docker is a better option for you?

> Build once on the developer's machine, then mount. We'd have to move to virtualenv for the Python dependencies, I think all other things are already locally installed.

Could you explain a bit more what do you mean by _build once .. then mount_?

I'll leave some comments on the changeset, to make sure which things can break.

Makefile
```diff
@@ -1,27 +1,28 @@
-egg-info:
-	pip install -q -e .
+app:
+	gunicorn -w 5 -b 0.0.0.0:8000 --log-level info --log-file /var/log/gunicorn.log aleph.manage:app
```
stas: The previous command was different in the `prod` file since we need support for hot-reload and other minor features when developing.

Makefile
```diff
@@ -1,27 +1,28 @@
...
 
-assets:
+install:
+	pip install -r requirements.txt
+	pip install -e .
+	npm install
 	touch aleph/static/style/_custom.scss;
 	(test -f '$(CUSTOM_SCSS_PATH)' && cp -f $(CUSTOM_SCSS_PATH) aleph/static/style/_custom.scss) || exit 0
-	/node_modules/webpack/bin/webpack.js --env.prod
+	./node_modules/webpack/bin/webpack.js --env.prod
```
stas: Using local `node_modules` will mess with the mounted repository when working in the development mode. We've had it before and were reported all kind of weird issues, see #125 #126 #130

docker-compose.prod.yml
```diff
@@ -1,95 +0,0 @@
-version: "2"
-
-services:
-  postgres:
-    image: postgres:9.4
-    volumes:
-      - "./build/data/postgres:/var/lib/postgresql/data"
-      - "./build/logs/postgres:/var/log"
```
stas: I suggest always providing the persistent volumes support, both in the development and production mode. Sometimes you generate data and local accounts and you don't want this to be gone after your battery dies or you forgot to update the production YAML file accidentally.

docker-compose.yml
```diff
@@ -1,100 +1,69 @@
 version: "2"
 
 services:
-  resolvable:
-    image: mgood/resolvable
-    volumes:
-      - "/var/run/docker.sock:/tmp/docker.sock"
-      - "/etc/resolv.conf:/tmp/resolv.conf"
-
```
stas: This is very handy in the development mode, since you don't have to search for the container IP address if you want to connect to the postgres database or point your browser to some endpoint.

stas:
> docker-compose.yml: "Just works" out-of-the-box, based on the pudo/aleph image. Users only need docker-compose.yml and an aleph.env, no building necessary (could even go in its own repo?)

One last thing. This is actually also an WIP. See #146 
As far as I know, editing `default_settings.py` might still be required for additional OAUTH support.

Makefile
wpf500: This pull request is about partly about fixing that problem so we don't have to use a hack to install node_modules globally

docker-compose.yml
wpf500: Sorry if it wasn't very clear, but in this PR docker-compose.yml is the production version, hence why I've removed these bits and they would be re-introduced in docker-compose.dev.yml. Also most of the code changes are just hacky at this point, I should have made that clearer in the PR

docker-compose.prod.yml
wpf500: :+1:

Makefile
wpf500: Yup I think the Makefile should be a collection of all our commands (sort of how npm scripts are for JS projects). There will be an equivalent for running Flask in debug mode.

wpf500:
Thanks for taking a look @stas

> This is indeed an issue. But this is how Docker works. The most we could do and already suggested is to try splitting the front-end dependencies into separate Docker containers (say provide an image for working on the front-end aka Node.js stuff, and one for working on the back-end aka Python stuff).

> The reason splitting (say app from workers) services is not an option, is because the codebase requires the packages to be installed before starting any of the services. Some packages just refuse to install if the system libraries are not present on the system. This is another reason we started working on ingestors and making those more pluggable. See: alephdata/ingestors#1

I agree that for the production build we should absolutely have the services in different containers, what I'm suggesting is for development they can all be hosted inside the same container. This is effectively exactly how it would have been pre-Docker (i.e. all services running on the same machine).

> What do you mean by local?
> The reason we added Node.js support to the image is to make sure we are all working with the same Node.js version, along with the same Webpack and etc. If we start mixing these we might end up breaking the builds for the front-end.

By local I mean the dependencies are created within the project (i.e. `./node_modules`), rather than globally as is the case for Python (`/usr/lib...`)

> Why local installation, sans-Docker is a better option for you?

I think Docker is definitely the way to go, as you've mentioned in locks the dependencies amongst other things. But at the moment the docker-compose.yml file doesn't actually work for new developers, this is about fixing that

wpf500:
Maybe it would help if I did a bit more work on the PR so you can see what I'm aiming for, admittedly the code changes are obscuring things a bit, sorry about that.

stas:
> I think Docker is definitely the way to go, as you've mentioned in locks the dependencies amongst other things. But at the moment the docker-compose.yml file doesn't actually work for new developers, this is about fixing that

Will, you didn't answer the question. :)
I'm still confused what exactly is _not_ working, besides the fact that it is sort of a slow process, which I'm not sure we should try to fix without addressing relevant existing issues mentioned above.

If your main frustration comes from the fact that you want to run the development of Aleph outside the Docker, I'm not sure this should become a priority right now.

wpf500:
I want to run aleph in Docker, but it doesn't work because the containers aren't set up correctly, as described in #163. All of the issues you've mentioned above are related to the way the docker-compose.yml currently half overwrites the installation, which is why the new docker-compose.dev.yml is based on `pudo/aleph-base`.

PR #169, opened 2017-03-10
Added CSV reader to aleph entities extractor

smmbllsm:


pudo:
Do you think it would be possible to make this into two subclasses -- one for CSV loads and one for SQL loads? I also would prefer to use a single unicodecsv iterator to load data from a CSV file, with no support for the SQL query manipulators (``join``, ``filters``, ``where``) whatsoever. That should make things much simpler. What do you think?

smmbllsm:
So the intention is just to do very quick extractions? That should be easy enough.

pudo:
I guess filters could also implemented on the stream, but a proof of concept does not need that. So yes, let's keep it ultra-simple :)

smmbllsm:
Yeah, I would've stuck with the csv iterator, but it was just the joins
that made it impossible.

smmbllsm:
@pudo Update

PR #171, opened 2017-03-11
Added CSV reader to aleph entities extractor

smmbllsm:


pudo:
Ha, this is getting a little simpler and more explicit! But I don't quite get what's going on in CSVQuery yet, it seems there's still some stuff we may not need: we really only need 1 table for CSV loading, not multiple files. So you probably don't actually need the CSVQueryTable (or, for that matter, the whole ``refs`` mechanism for table aliasing). That also means you don't have to open it twice (but you could, if you wanted to do explicit encoding detection!). 

Some random suggestions: 
1) Perhaps instead of ``table_location`` we can name it a little more explicitly, say ``csv_url``?
2) It seems you are opening the CSV both with requests (good!) and ``open()`` (weird). When you open it with ``open`` you do expandvars (which is ultra-cool for passwords) but you don't do it for ``requests``. Perhaps that could be more coherent?

smmbllsm:
Okay, so we just want to be able to point it at a single URL and load one
table? I was trying to provide the ability to load from local files as
well, that's why the call to open(). That's also why I made the path/uri
variable ambiguous :). But I'll update it according to your suggestions.

PR #172, opened 2017-03-14
AutomatonCache now always returns an Automaton

smmbllsm:


smmbllsm:
That's a good point.

pudo:
Have you tested this? I'm curious to see if the automaton supports calling ``.iter()`` (which uses the automaton) before calling ``.make_automaton()`` (which ... makes an automaton).

PR #176, opened 2017-03-29
Add case #, court, fileDate, closeDate to courtcase schema

smmbllsm:


pudo:
So with all the other dates (``startDate``, ``endDate``, ``incorporationDate``, ...) we have the Date part at the end. Would it be OK to do this here as well?

PR #189, opened 2017-05-06
Ingestors

stas:
WIP transition to the new ingestors.


pudo:
looks pretty good, light touch, thus far. nice!

aleph/ingest/all.py
```diff
@@ -0,0 +1,175 @@
...
+    def pdf_to_xml(self, *args, **kwargs):
+        """Archive the PDF before the XML converstion."""
+        document = Document.by_meta(self.collection_id, self.aleph_meta)
+
+        file_meta = archive.archive_file(
+            kwargs.get('file_path', args[1]), document.meta.pdf, move=False)
+
+        document._meta['pdf_version'] = file_meta.content_hash
+        # Weird SQLAlchemy stuff. The meta column is never updated otherwise.
+        document.meta = document.meta
```
stas: @pudo there's some weird issue with the document metadata. I couldn't figure out why the metadata would just reset after the `db.session.commit()`. It has something to do with the SA's `flag_modified(...)`... :hankey: 

If you know what could be the issue, I'd be happy to look at it again.
Also, I find the whole metdata over-engineered. The proxies to other attributes are not easy to follow and hardly work. I'd be happy to give up on these as we discussed on #aleph already. See below an example:
```
(Pdb) document.meta.pdf.content_hash
(Pdb) document.meta.pdf.content_hash = 'a'
(Pdb) document.meta.pdf.content_hash
(Pdb) document.meta.pdf_version
```

aleph/model/document.py
```diff
@@ -146,6 +153,26 @@ def is_crawler_active(cls, crawler_id):
...
+        else:
+            raise ValueError("No unique criterion for document: %s" % meta)
+
+        document = q.first()
+        if document is None:
+            document = Document()
+            document.collection_id = collection_id
+            document.foreign_id = meta.foreign_id
+            document.content_hash = meta.content_hash
+            document.meta = meta
```
stas: This changed, I think it is not ok to overwrite the metadata object every time when you can build it from the database metadata field.

https://github.com/alephdata/aleph/blob/master/aleph/ingest/ingestor.py#L61

aleph/ingest/all.py
pudo: Well that actually makes sense, it's unrealistic to expect SQLAlchemy to do dirty checking on the wrapper type ``Metadata``, which doesn't even use the dict given by the ORM internally. That just means ``meta`` updates on the document [need to be explicit](https://github.com/alephdata/aleph/blob/master/aleph/model/document.py#L76).

Regarding refactoring this: we can switch to using ``Document`` directly has the metadata holder object, which solves this SQLA issue. However, the attribute proxies are based on the actual *need* for data parsing and validation. We cannot just drop them without replacement. Using a ``dict`` for metadata has been tried, and it is *not* enough. We do not need to repeat this error.

stas: Alrighty. For now, it works and I'll leave it so that we can improve it later on. I've worked on a similar issue for the peewee ORM and it's not fun, but it's doable.

stas:
@pudo I added some integration tests to make sure paged and non-paged ingestion works properly.

Take a look when you have some time, I'd be happy to answer any questions.
Ty.

stas:
This is now rebased with the latest master changes.

/CC @pudo

pudo:
Close in favour of #193.

PR #190, opened 2017-05-14
Role password signup fixes

cdharris:
Was investigating using Aleph and wanted to use email and password signup rather than Oauth, I couldn't get the signup process to work. The user role would never end up in the database for the next session.

I think the current test  `RolesApiTestCase.test_create_success` passes because the database session is still around from which to retrieve the role from, but if this session is closed, as would happen for a new request, the role won't have been committed to the database yet.

I've included a change to the test to illustrate this and a fix in the API view.

I'm new to Aleph so maybe I've missed/overlooked something here, feedback welcome, maybe I'm just not doing something right.

Cheers!

### DIFF
--- a/aleph/tests/test_roles_api.py
+++ b/aleph/tests/test_roles_api.py
@@ -1,5 +1,6 @@
 import json
 
+from aleph.core import db
 from aleph.model import Role
 from aleph.tests.util import TestCase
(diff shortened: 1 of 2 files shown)


stas:
Thanks @cdharris 
I found the same issue in an upcoming PR (cefae5a), so I'll just merge yours.

PR #201, opened 2017-06-20
csv datasets: add support for file protocol

felixebert:


pudo:
Very cool, I've been meaning to have that! Do you also want to do one against develop? o/w I'll backport it.

PR #234, opened 2017-11-01
Rework role "invitation" flow

wpf500:
The invitation flow looks more like a standard register/activate flow

This PR mostly changes the naming, the process is the same except:

- No email required for activation
- Return error if user tries to create role with in-use email (is this a security leak?)


aleph/views/roles_api.py
```diff
@@ -59,28 +60,31 @@ def create():
...
+                foreign_id='password:{}'.format(email),
+                type=Role.USER,
+                name=data.get('name') or email,
+                email=email
+            )
+            role.set_password(data.get('password'))
+            db.session.add(role)
+            db.session.commit()
+            # Let the serializer return more info about this user
+            request.authz.id = role.id
```
wpf500: This feels like magic, should it be here? We need to return the email to the user and currently there is some logic [here](https://github.com/alephdata/aleph/blob/master/aleph/views/serializers.py#L79) which removes the `email` if they aren't authed.

This makes sense in theory, but the `foreign_id` also contains the email address and is always serialised. Should it be removed too?

PR #244, opened 2017-11-12
fix index mapping for es5

tpreusse:
Derived mappings from the master branch for es5 compatible. Might not be perfect but worked for me. Didn't encounter any more errors after this change.

While importing thousands of emails I also run into `index.mapping.total_fields.limit`. I ended up bumping the limit with:
```
PUT aleph/_settings
{
  "index.mapping.total_fields.limit": 10000
}
```
Have you considered bumping the limit in the index creation method?


pudo:
Hey Thomas, thanks for the PR! I have to figure out a little bit how to merge it, because changing the index types requires rebuilding the index, and doing that with the live OCCRP site would take weeks. We might need to remove automatic ES schema upgrades from the default upgrade path, so that people who start a new index get the new schema, but existing indexes aren't touched.

P.S. the max fields issue is due to the way that aleph 1 indexes spreadsheets. Basically, every spreadsheet column name ever seen becomes a schema component. It's fixed in ``master``, but really ugly behaviour. 

Note to self: https://github.com/alephdata/aleph/blob/master/aleph/manage.py#L153

tpreusse:
Removing the automatic ES schema upgrades sounds good.

Have you thought about the migration path to v2 for OCCRP? 

Using index names with versions and a separate [search and index aliases](https://www.elastic.co/blog/changing-mapping-with-zero-downtime#_using_aliases_without_reindexing) looks interesting. Then you could [reindex](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/docs-reindex.html) in the background while new documents already get added to the new index. But might not work if the schemas diverge too much and does add complexity.

PR #275, opened 2018-01-08
Documentation for ingest API endpoint

sunu:


pudo:
Looks good to me, thanks for writing it up!

pudo:
Actually, we should probably explain: 

Metadata is sent via a JSON-encoded POST variable, ``meta``, which can contain nested values and lists for the specified metadata fields.

sunu:
@pudo updated.

PR #278, opened 2018-01-10
rdfdump

rhiaro:


pudo:
Do you also want to get started with a page for our tech blog to talk about the RDF data release? I guess we'd first start with a few basic collections, like a few interesting leaks. We can probably publish the dump files on archive.org.

PR #310, opened 2018-02-21
Emina editing collection

emmina:
- made collection edit screen
- saves collection info and permissions for collection
- added toast boxes on save button on collection editing and profile screen
- added css for these kind of tables


pudo:
Can I make on super-annoying request? You're currently using the term "user" all over the code - which is understandable, but creates ambiguity with the term "role" which is used by the backend and also for groups. Would it be possible to change all of the "user" things to "role" things?

iaincollins:
Generally everything works fine for me!

Things that I think might be worth looking at in the UI:

- [ ] Padding above "You cannot access collection editing." error would be nice as the text clashes into the navbar at the moment. I think `style={{paddingTop: 50}}` is a decent workaround.
  
  We could also make an issue to improve error handling if we need a new (or better) re-useable control for displaying errors. I don't know how the existing one is used yet so unsure if refactoring it would be the best move.

- [ ] I would lean towards avoiding exclamation points in prompts to users, and just make a less urgent suggestion to them if we want to prompt to do something, unless it's something important (e.g. "This will delete your account!") .

- [ ] I think if the search place holder was "Search for user…" and the button said "Add user" it would more clear what this form was for. (I assume we don't want to say "Search roles…" in the user interface but maybe @pudo has a view on that.)

- [ ] I think the Search input and Add button are not aligned quite right.

- [ ] The Add button should probably be disabled if the search field is empty.

- [ ] After adding a user I think it it should automatically clear the search field.

(I haven't looked at the code much but will do that and maybe leave another comment :)

pudo:
on the "adding user" UX: do we want that to be up on top, or down where the users are listed?

PR #311, opened 2018-02-21
Improved views (toolbars, paging, width limits…)

iaincollins:
* Added a toolbar to every document view (PDF, Text, HTML, Email, Folder, Table).
* Main content in Entity and Document views is now constrrained with a max-width property.
* Both the above has necessitated some viewer refactoring and changing how content is are rendered with flexboxes so it was a bit more involved that just adding a new toolbar.
* Most of the views are still work-in-progress - all fully functional but not where we want them to be yet (e.g. not 'final' layouts, disabled document search or other toolbar options not yet implemented).
* The search method for folders/emails no longer works since the search refactoring so I have disabled it (as it just errors if you try to use it right now).
* Some issues with Blueprint SVGs causing rendering problems in the Sidebars are addressed in this update. I have rolled back to using the Blueprint font library as the JS API creates SVGs and they seem to have issues but the font is fine.

screen shot 2018-02-23 at 11 14 02
screen shot 2018-02-23 at 11 11 14


iaincollins:
**Known issues:**

* The email view is ugly.
  * I'm not making a statement with the tabs for attachments here, it's just a work in progress (but functional).
  * The attachments table is a search component, which is why it has column lines on it still.
* There are multiple existing issues with the table view, but to clarify these are not new (they existed long before I touched anything).
  * You can only scroll sideways with a track pad or if your mouse has horizontal scroll option as the the horizontal scrollbar is hidden when you scroll down.
  * A lot of tables - most of the ones I try - are not loading for me. Known good files (both Excel and CSV) are fine, but a lot in the current database just don't load.
 * The folder view is functional and a raw search component but we may want to render it differently (e.g. with padding on the container below the toolbar).

Lots of obvious improvements for each view (add tabs on spreadsheets, try out Blueprint components on them, improve email views, get search working in each view, handle loading errors better, etc).

ui/src/components/DocumentScreen/viewers/PdfViewer.jsx
```diff
@@ -66,22 +66,21 @@ class PdfViewer extends Component {
...
-          <button className="pt-button pt-fill" disabled={true}>{pageNumber} of {numPages}</button>
-          <Link to={{ hash: `page=${pageNumber+1}` }}><button className="pt-button pt-icon-arrow-right" disabled={pageNumber >= numPages}/></Link>
+      <React.Fragment>
+        <Toolbar document={document} pageNumber={pageNumber} pageTotal={numPages}/>
+        <div className="ContentPaneOuter">
+          <div className="ContentPaneInner PdfViewer max-width-content-limit">
+            <div className="document_pdf" ref={(ref) => this.pdfElement = ref}>
+              <Document renderAnnotations={true}
+                        file={url}
+                        onLoadSuccess={this.onDocumentLoad.bind(this)}
```
emmina: put binding in constructor of the class

ui/src/components/DocumentScreen/viewers/PdfViewer.jsx
```diff
@@ -66,22 +66,21 @@ class PdfViewer extends Component {
...
+                <Page pageNumber={pageNumber} className="page" width={width} />
+              </Document>
+            </div>
+          </div>
         </div>
-
-        <div className="document_pdf"  ref={(ref) => this.pdfElement = ref}>
-          <Document renderAnnotations={true}
-                    file={url}
-                    onLoadSuccess={this.onDocumentLoad.bind(this)}
```
emmina: put binding in constructor of the class

iaincollins:
I've addressed this in #7e8921 -  though it was existing code not directly part of the toolbar update, just moved around in the refactoring (I was reluctant to touch the PDF viewer invocation as it does some funky stuff).

Note we are also doing this here, but don't want to make updating auth code part of this PR:

````
$ git grep bind | grep '{'
src/components/auth/ActivateScreen.jsx:        <PasswordAuthActivate onSubmit={this.onActivate.bind(this)}/>
src/components/auth/LoginScreen.jsx:        {passwordLogin && <PasswordAuthLogin onSubmit={this.onLogin.bind(this)}/>}
src/components/auth/SignupScreen.jsx:            <PasswordAuthSignup onSubmit={this.onSignup.bind(this)}/>
````

I have also refactored some of Toolbar the old viewer CSS following the discussion earlier.

Though some of the views still differ in how they apply CSS, I think it makes more sense to do work on them when we come to do more work on each view - as this is a PR specifically to add the Toolbar (and minimal work to get away with that) as I've probably already put more unrelated changes in this PR than I should have.

PR #328, opened 2018-02-26
Entity.Load & Collection.Load components

benoccrp:
So we can reuse the same fetch-if-needed logic every time it is needed. They function somewhat like a higher-order component by requiring a function as its `children`.


benoccrp:
Merged in commit 4623f96, but somehow having lost the reference to these commits.. closing manually.

PR #341, opened 2018-03-06
Flush documents and entities from a collection separately

rhiaro:


aleph/index/collections.py
```diff
@@ -118,3 +118,23 @@ def delete_collection(collection_id, wait=True):
               doc_type='doc',
               id=collection_id,
               ignore=[404])
+
+
```
pudo: These two look correct, that's the best logic we have.

pudo:
Is there any point to us using these two methods in the collection deletion code and getting rid of the duplicate deletion logic? Otherwise let's merge.

PR #344, opened 2018-03-07
[Work In Progress] Preview sidebar

iaincollins:
**Work In Progress of Preview** 

Not intended for merging, just a PR so it's easy to see the current state and what's been done.

sidebar

Plenty of refactoring we can do inside Preview.jsx to make it more generic and to improve visual appreance when loading new contents, but the initial impmention should be pretty robust and worth playing with.

Features so far:

- [x] Enabled on Search and Collections views.
- [x] Loads and save state from contents of hash path (eg #preview:id=123).
- [x] Watches route for changes to hash path and updates contents accordingly (e.g. can just use <a href="#preview:id=123&preview:type=document">link</a> to show a preview).
- [x] Can be maxmised / minimised and hidden.
- [x] Keeps track of it’s state (e.g. if it was maximised by user, stays maxmised when new content is loaded, making it easy to stay in preview mode while going through a list of search results).
- [x] Uses a simple CSS transition animation when becoming visible, hiding and re-sizing.
- [x] Can show summary of entities, document and collections and (when maximised) shows previews of documents.

Known issues:

- [x] ~~Contents is blank while new content is being loaded.~ Now has a loading spinner.
- [ ] More explicit inline comments would be helpful for future developers.
- [ ] Could maybe handle scrolling long content in the sidebar better (fix the toolbar / heading?)


iaincollins:
Thanks for the feedback!

1. So the idea behind the defining defaultState like that is just so that there is a single definition for the state object, so that when we need to reset the state (e.g. line 86) it's always reset to the same initial state.

  This is a fairly common pattern for this sort of situation, just to avoid duplicating the code which resets the state with the intent of making it harder for bugs to creep in later (and done this way because it only needs to be an object, not a method), but I'm happy to make it a function inside the class that returns an object if we prefer that.

2. Sure, can do!

3. BTW I am totally happy to use classes for styling (and have avoided doing that that where possible - as there was pushback against using scoped JSX) but in this case it's dynamic so it's an instance where we do need to set the style attributes directly to have a pinned sidebar.

It can't be pre-calculated because the top position is based on the current scroll offset and dynamically takes account of the amount of visible Navbar. This is basically the 'least awful' way of doing it, as position sticky / fixed added more complications and required more watchers and more wrapper element sand more code.

4. Noted. For reference, the only px values it's using are:

* 1px for the border - this was hard coded already elsewhere so I just took a cue from that but I can refactor into something like aleph-border-width (so we can faff with toggling border in the UI if we want).

* 1024px for max-width of a maximised preview - this is an experimental value and probably wrong, and is probably going to be Preview class specific thing, but happy to set as a variable if it makes it easier to manage.

5. This was just to give the best of both worlds (short include paths _and_ files with explicit names), but I'm happy however we decide want to do it. :-)

I appreciate this approach also has a downside that it litters the place with small index.jsx files and am happy to remove the index.jsx from components that only have a single JS file.


Will loop you in again for another review when I've added entities and document previews in!

(Will see how that complicates things!)

pudo:
Just started playing with this branch, some comments which are entirely random:

* It's slightly odd to have two toolbars on top of each other when previewing a document.
* I find the contract/expand thing a weird logic for switching between document metadata/content preview. Perhaps a toggle with labels like "Document | Metadata" could be more clear?
* Would be cool to add an "active" CSS highlight to the result row for which preview is shown.

iaincollins:
1. Agree about the double toolbars. I wasn't sure what we wanted to do here before tackling it.

    If we don't want to show a second toolbar - we can move the paging buttons into the toolbar if we liked (though might get odd if they were below tabs?).

    Alternatively we could display the paging buttons above/below the document (and not in a toolbar).

2. Agree. It seems odd and am concerned that people wouldn't know they could show a preview without being shown it (or stumbling on it).

    a. Happy to try tabs (had forgotten we sketched that out!), but also wondering if resizing the pane when changing tabs might also be weird?

    b. If we had text next to the current maximise button that said something like "Show Preview" might that help?

3. Agree - on it!

    I'd like to include support keyboard navigation too, to allow quick paging through results, if that can be done well.

PR #374, opened 2018-04-01
Login and registration on one screen, dialog

emmina:


pudo:
Great PR! One thing I noticed: the ``/login`` route was used by OAuth to supply a JWT to the UI once the login was successful. This happens in componentWillMount where it's reading the token from the location fragment. We probably want to make this a separate route, like ``/oauth`` now that the login is a dialog. 

Also: have you checked that `/login` isn't referenced anywhere else in the code base any more?

Issue #334, opened 2018-03-01
[META] Case management


pudo:
We're planning to add functionality to Aleph 2 which would allow reporters to manage the people, companies, events and documents that constitute an investigation inside of the platform. This will include options for the editing and navigation of a particular case by means of tabular listings, using entity-related dossiers (wiki pages) and network diagrams (like in current VIS). Our goal is to make it easier to build and share a consolidated case file that includes all the atoms of a particular story.

### Overview

* Introduce a distinction in collection types, between a `Source` (e.g. a scraped or leaked dataset) and a `Case` (which is manually curated and assembled).
* Introduce a new domain object, `Dossier`, which describes a conceptual entity - such as a `Person` or `Company` - that is made up by linking together many `Entities` from different collections. Unlike with an `Entity`, the ultimate goal of a `Dossier` is to have only once instance of a particular concept (e.g. "Donald J. Trump") in an Aleph instance. A `Dossier` does not have a `collection_id`.
* A `SameAs` (name?) describes the link between an `Entity` (either bulk-loaded or not) and a `Dossier`. It can indicate a) the `Entity` being a constituent part of the `Dossier`, b) the `Entity` being definitely distinct from the `Dossier`, c) the identity has yet to be decided by a user.
* A `Link` (name?) describes the inclusion of a `Dossier` in a `Case`. Any `Dossier` can appear in numerous investigations. The `Link` will also store `notes` and `tags` detailing the research conducted on the thing described by the `Dossier` as part of the `Case`.

### Plan of Attack

Since this is a large feature, we cannot do it all at once. Here's some possible entry ways we have discussed:

* **Bookmarking entities and documents while searching.** This is a much-requested feature, and it can be developed using a combination of `Link`, `Dossier` and `Case`. It will essentially leave the user with a set of their own investigations (`Cases`) and the documents and entities they have chosen to file into each of them.
* **Porting Oligrapher to run on Aleph.** Since we want to support network canvases (and eventually replace VIS), we could leapfrog the development by making it a milestone to build an API capable of running the Oligrapher tool on top of it. For this to work, we need to map the RDF-style entity data model of Aleph into a property graph-style model (similar to Neo4j) for use in the network tool.

### Links

* https://github.com/public-accountability/oligrapher - Oligrapher, from LittleSis. We'll use this to prototype the network diagramming functionality in Aleph, possibly to refactor or replace it later.
[...]

Issue #291, opened 2018-01-17
Creating collections via the UI


pudo:
A simple pop-up to make a new collection. Only need to define label, category and optionally countries/languages/summary.

PR #377, opened 2018-04-03
Data model for case management

### LINKED ISSUE #245: Improve collection provenance
* Who is the publisher behind a collection? (`publisher`, `publisher_url`)
* Further links: `info_url`, `data_url`
* Who handled the information in the interim?

And, for entities: 

* How does the entity relate to the structure of the source data?
* How can we make the source data accessible?

pudo:
This PR contains the basics for implementing a dossier and case management system in Aleph, based on the discussion in #334. The idea is to merge this as invisible functionality relatively soon, and start exposing it API by API as we implement functionality like #291, and entity/document bookmarking into a case file.

Here's some of the changes in this: 

* Fixes #245 by introducing new collection fields: ``publisher``, ``publisher_url``, ``data_url``, ``info_url``. 
* Introduces ``DossierMatch`` and ``DossierLink`` to generate dossiers. 
* Renames collection ``managed`` flag to ``casefile`` and inverts meaning. 

Still pending (not all of this needs to be done before this can be merged): 

* Generate and index dossiers into ElasticSearch
* Adapt relational query endpoints to expand on entity IDs (or more generally, handle the case in which we now have dossiers which have multiple entity IDs that they may store relations on). 
* Figure out how to merge two dossiers properly, and, based on that, how to cleanly add an entity to a dossier. 
* Figure out document dossiers -- what do they mean?
* Make a ``/api/2/dossiers`` endpoint.
* Make AhoCorasick extractor use dossiers for dictionary generation.


pudo:
So --- looks like we couldn't come up with a good editorial model that would allow Dossiers to be both per-collection and global, i.e. where changing the dedupes of a case-specific dossier also change the dossier in other collections. We might have to go with a simpler model, where Dossiers exist within a particular case, rather than globally.

While we figure this out, I'll close and split up this PR so that we can at least get the collections model changes into master asap.

PR #383, opened 2018-04-04
ip & iban regex analyzers

jcshea:
needs corresponding exactitude push as well as some tag-creation magick ;)


pudo:
Some feedback:

* I wonder if it would make sense to combine the emitted IPv4 and IPv6 addresses into a single field, "ip". In terms of cross-referencing and UI that seems more desirable.
* Not sold on the IBAN regex, did you check that covers all possible country IBAN formats? There's [some discussion](https://stackoverflow.com/questions/44656264/iban-regex-design) on SO and it seems they're settling on a more liberal format. 

Cleanups:

* Cannot leave ``print`` in production code, it's fine for local debug
* Checked in some temp files in ``test_data``.

PR #421, opened 2018-05-23
ingest directories async

felixebert:
... for faster traversal through a lot of nested directories.

### DIFF
--- a/aleph/logic/documents/ingest.py
+++ b/aleph/logic/documents/ingest.py
@@ -35,18 +35,12 @@ def ingest_document(document, file_path, role_id=None):
     if role_id is not None:
         document.uploader_id = role_id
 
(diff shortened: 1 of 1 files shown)


pudo:
I'm a little bit torn on this: the implication of this is that the source data being ingested has to be present in the container which executes the worker that will eventually do the processing. That can be made true as long as you're running the app on a single server, but it falls apart once you're using multiple servers. OTOH, at this point probably most of the data loading that happens on a production cluster is via the API, and not via command-line tools.

pudo:
Ok I've played with this a bit now, and it has knock-on effects against the API not working properly any more and imports on clustered installs becoming impossible. Sorry, but I'm gonna reject this.

PR #428, opened 2018-06-19
Navbar on mobile versions and medium size devices

emmina:


pudo:
I think we need to get rid of some of the use of `!important` in the CSS by making more specific selectors, otherwise this'll have unintended consequences. Here's an article about it: 

https://css-tricks.com/when-using-important-is-the-right-choice/

pudo:
OK, more feedback: 

screenshot 2018-06-20 14 58 57

There's a big gray bar around the menu, is that intentional? Is there any way to make it show up underneath the navbar, full-width, and without overlaying anything? 

b) When the window is wider, it renders as this on my browser -- assume that's an error? Also, there's still a white box inside the gray hover area when you mouse over elements. 

screenshot 2018-06-20 15 00 21

Let's get this really polished and ready to merge before we continue work on the facets!

PR #431, opened 2018-06-28
Entity extractor refactoring

sunu:


sunu:
Tests should pass after https://github.com/alephdata/alephclient/pull/4 is merged and a new version of alephclient is pushed out.

PR #434, opened 2018-07-02
Aleph on small and medium devices

emmina:
- fixed facets (on small devices facets are on dropdown, on medium facets are 250px)
- fixed preview (on small devices preview is from left to right, on medium preview is smaller width)
- fixed info bar and side pane (on small devices they are full width and on top of the screen, on medium they are 300px)
- fixed showing tables (on smaller devices there are only 2 columns)
- fixed footer(removed breadcrumbs on smaller and medium screens)


ui/src/components/EntityTable/EntityTable.scss
```diff
@@ -8,4 +8,30 @@
   .updating {
     opacity: 0.4;
   }
+}
+
+//small screen
```
pudo: This looks messy. Do we need to do it separately for medium and small? Why do we have two levels of invisibility going on?

ui/src/components/Navbar/Navbar.scss
```diff
@@ -31,7 +31,7 @@
     }
 
     .heading-title {
-      padding-right: 18.5em;
+      width: 21.5em;
```
pudo: Is this regarding the line break in the current name? Is there a solution that doesn't hard-code a width at all?

ui/src/components/Navbar/Navbar.scss
```diff
@@ -109,7 +113,7 @@
 }
 
 //small screen
-@media screen and (max-width: $aleph-screen-xs-max-width) {
+@media screen and (max-width: $aleph-screen-sm-max-width) {
```
pudo: In general: I think it's much more readable if we keep the media queries close to where the objects are declared, not as a separate secondary structure. Instead of: 

```
.Navbar {
    .h2 { 
       color: red;
    }
}

@media (...) {
   .Navbar { 
        h2 { color: blue; }
   }
}
```

It's much cleaner to do: 

```
.Navbar {
  h2 {
     color: red;
     @media () {
          color: blue
       }
    }
}
```

See also: http://thesassway.com/intermediate/responsive-web-design-in-sass-using-media-queries-in-sass-32

ui/src/components/common/DualPane.scss
```diff
@@ -68,20 +68,35 @@
   }
 }
 
-
-@media screen and (max-width: $aleph-sidebar-width * 2) {
+//small screen
```
pudo: Same as above: can we nest the media queries directly into the object?

ui/src/screens/SearchScreen/SearchScreen.jsx
```diff
@@ -129,10 +131,11 @@ class SearchScreen extends React.Component {
         icon: 'person'
       }
     ];
-    this.state = {facets: facets};
+    this.state = {facets: facets, toggle: false};
```
pudo: can we rename toggle -> hideFacets or something? Toggle is too generic as a name.

ui/src/components/Navbar/Navbar.scss
emmina: @pudo i get what you are saying, i saw these kind of media queries and how they write it, but in one file (for example Navbar.scss) i have to change more than 15 classes, it will get really messy if i add for all of the classes media query and there will be a lot more lines of code then when you put it in one place. But I can do that in smaller scss files, but then it would not be consistent.

Issue #437, opened 2018-07-06
Add support for authentication via Azure Active Directory


Ueland:
As we need to get some AD/LDAP authenticaton into Aleph in order to use it, i will start off contributing to this project by removing some dust from this PR, and send it back inn after seeing that it works.

https://github.com/alephdata/aleph/pull/162

Feel free to assign this issue to me

Ueland:
As @pudo earlier mentioned that the previous LDAP support was not used by anyone, i will take a look on that code and see how it can be rewritten to support Azure Active Directory instead. That should make it possible to simply plug Aleph into any organization that uses Office65/Exchange for authenticaton.

pudo:
Sounds great. I'm curious, though, if Office365 could be linked on with OAuth, it strikes me as a somewhat cleaner solution...

Ueland:
I do believe that it is just that which happens actually, the OAuth request goes to "https://login.microsoftonline.com/<login type>/oauth2/v2.0/authorize". Our organization uses office365 too. So not quite sure if we should call this OAuth via office65 or Oauth via Azure AD.

I have a fork where i am working on it, will do a PR when i have it working.

Ueland:
Just adding some notes here, the following are required for OAuth 2.0 auth against Azure/AD, and should  be added to the Wiki after merge:

> Create a new app over at https://apps.dev.microsoft.com, make a note of the KEY and secret you generate there. Callback URL should be as follows: https://<hostname>/api/2/sessions/callback .
Then add the following to aleph.env, remember to update KEY and SECRET with your values.

>ALEPH_OAUTH=true
ALEPH_OAUTH_KEY=***************************
ALEPH_OAUTH_SECRET=***************************
ALEPH_OAUTH_NAME=Azure
ALEPH_OAUTH_BASE_URL=https://login.microsoftonline.com/organizations/oauth2/v2.0/
ALEPH_OAUTH_AUTHORIZE_URL=https://login.microsoftonline.com/organizations/oauth2/v2.0/authorize
ALEPH_OAUTH_TOKEN_URL=https://login.microsoftonline.com/common/oauth2/token
ALEPH_OAUTH_SCOPE=openid profile email user.read
ALEPH_OAUTH_TOKEN_METHOD=POST
`

pudo:
Are you using a custom handler in `aleph/oauth.py` for handling the returned token?

Ueland:
[...]

PR #451, opened 2018-07-25
Added support for authentication via Azure AD (OAuth)

Ueland:
This implements #437 

If accepted, the following wiki page should be updated: https://github.com/alephdata/aleph/wiki/System-configuration, with the comment here: https://github.com/alephdata/aleph/issues/437#issuecomment-407038206


pudo:
woow, some crazy shit on the cert parsing :) looks good to me, though.

PR #458, opened 2018-08-01
Log user queries, document views etc

sunu:


aleph/logic/audit.py
```diff
@@ -0,0 +1,26 @@
...
+ACTIVITY_SCHEMA = {
+    "USER.LOGIN": [],
+    "USER.SEARCH": ["query_string", "filters"],
+    "USER.VIEW_COLLECTION": ["collection_id", ],
+    "USER.VIEW_DOCUMENT": ["document_id", ]
+}
+
+
+@celery.task(priority=1)
+def record_audit(activity_type, activity_data, authz):
```
pudo: I'm pretty convinced that we cannot put an `authz` object on the queue (it'd need to be JSON serialisable). So I think we need a record_audit function (which takes the authz) and which in turn calls record_audit_delayed (which would be on the queue).

aleph/migrate/versions/a1a4f4eccae5_add_audit_table.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def upgrade():
+    op.create_table(
+        'audit',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('activity_type', sa.Unicode(), nullable=True),
+        sa.Column('activity_metadata', postgresql.JSONB(astext_type=sa.Text()),
+                  nullable=True),
+        sa.Column('created_at', sa.DateTime(), nullable=True),
```
pudo: Should this have an updated_at column?

aleph/search/query.py
```diff
@@ -205,6 +206,16 @@ def scan(self):
     @classmethod
     def handle(cls, request, limit=None, schema=None, **kwargs):
         parser = SearchQueryParser(request.args, request.authz, limit=limit)
+        # Log the search
+        filters = []
+        for field, values in parser.filters.items():
+            if field not in parser.facet_names:
+                filters.append(field_filter_query(field, values))
+        record_audit.delay(
```
pudo: There's also prefix searches etc. etc. Would it perhaps be easier to somehow serialise the whole thing? Or at least have a testable `to_dict` method for the parser that generates a full state dump?

aleph/views/sessions_api.py
```diff
@@ -32,6 +37,18 @@ def _get_credential_authz(credential):
         return Authz.from_role(role=role)
 
 
+def _get_session_id(request):
```
pudo: I feel this could be done in the `record_audit` function mentioned above -- it doesn't really have anything to do with Authz (although authz may influence it on the front-end, but that's more ... coincidence).

ui/src/reducers/util.js
```diff
@@ -67,3 +67,10 @@ export function resultObjects(state, result) {
   }
   return state;
 }
+
+export function generateUUID() {
+  // js black magic from https://stackoverflow.com/a/2117523
```
pudo: that's some hard-core shit ;)

PR #464, opened 2018-08-13
Redis backed graph API based on FTM schema

sunu:


pudo:
This is now in ``develop``, closing.

PR #471, opened 2018-08-20
Move regex extractors to entity extractor microservice

pudo:
Just creating a PR to comment


pudo:
Some random-ass comments:

* For the regex patterns, can we pre-compile them and find a way of making ``REGEX_TYPES`` more readable?
* Let's have tests for the other types, e.g. IBANs, IPs, etc.
* I wonder if it would be worth trying once more to fix up the ``TextIterator`` in aleph proper, e.g. by passing in the document ID and then having it do a fresh SELECT for the doc, and iterating over contents.

PR #474, opened 2018-09-06
[WIP] Add message_id and in_reply_to properties to document

sunu:


sunu:
Tests should pass if we bump up the version of ingestors.

PR #497, opened 2018-09-25
API endpoint to delete audit log for a query

sunu:


pudo:
Looks really good, but I'd add a single check in the blueprint function to see if `require(request.authz.logged_in)`) and test for that with a 403.

pudo:
One more point that I realised just now: should we be using query arguments for the delete API? I feel like it's a little bit inconsistent with the API, which otherwise uses form/json data for POST data. Perhaps we could pull the text out of that instead of using the query URL param?

davo11122:
@sunu Could you please provide a simple example of API request

sunu:
@davo11122 to delete the search logs for the term "foo", the request will look something like
```
fetch("/api/2/querylog?query=foo", {
        method: "DELETE",
        headers: {...},
    }); 
```

PR #498, opened 2018-09-25
WIP: Query logs visualisation and management  

### LINKED ISSUE #479: Show query logs to users.
We have an API, ``/api/2/querylog``. This returns the search queries the user has run, aggregated by how often the term has been searched. We should have a way for the user to see this analysis, with the following actions for each queried term: 

* Create an alert from the query term (existing API)
* Re-do the search now
* Delete the query log for the term (API not yet implemented).

davo11122:
Resolves #479


ui/src/components/Activity/ActivityItem.jsx
```diff
@@ -0,0 +1,21 @@
+import React, {Component} from 'react';
+import ActivitySearch from "./ActivityItems/ActivitySearch";
+
+class ActivityItem extends Component {
```
emmina: I'm not sure what does this component do?

davo11122: In the feature we must include more activity types(not only search results) in the `/activity` screen, so `ActivityItem` component is a layer for switching a different types of activities.

PS: At the current moment this component is useless and can be removed, but in long term perspective we need it.

pudo:
Some feedback on this: 

* I think the auto-complete pop-up is too aggressive at the moment. could we make it come up only when a user has started typing a few characters? It's just subjectively intrusive at the moment. Also: do we need the arrows at the beginning of each line in the pop-up? What are they meant to convey? 
* Should the link to the activity log perhaps be in the main search bar, like the alerts button? That way it might be clearer and more readily accessed.
* Title: "All activities" -> "Your search history"? 
* The query log screen has a lot of space at the moment, feels almost a bit awkward. Are you super-opposed to using modal dialogs? I just feel that having that visual pop out might make it a more lightweight interaction ....

davo11122:
Most of the features implemented in PR #568, but some useful code still committed in the PR

pudo:
Closing in favor of #568.

PR #502, opened 2018-10-05
[WIP] Logic fixes for views menu refactor

pudo:


ui/src/components/Entity/EntityScreenContext.jsx
```diff
@@ -10,11 +10,14 @@ import LoadingScreen from 'src/components/Screen/LoadingScreen';
 import ErrorScreen from 'src/components/Screen/ErrorScreen';
 import { DualPane, Breadcrumbs, Entity } from 'src/components/common';
 import { selectEntity } from 'src/selectors';
+import { queryEntitySimilar } from "../../queries";
```
davo11122: As far as I notice we use relative paths for imports only in case of subfolders otherwise we are using absolute paths referring to `src` alias

PR #518, opened 2018-10-16
Kubefiles with healthchecks added to sidecars

Ueland:


pudo:
We still need to do work on this, but I'm merging so it doesn't get lost (it's not in the way of anything)

PR #520, opened 2018-10-17
NER wants to be a service after all

pudo:


services/extract-entities/entityextractor/service.py
```diff
@@ -0,0 +1,109 @@
...
+        for result in self.extract_all(request.text):
+            if result is None:
+                continue
+            yield result
+
+
+def serve(port):
+    executor = futures.ThreadPoolExecutor(max_workers=3)
+    server = grpc.server(executor)
+    add_EntityExtractServicer_to_server(EntityServicer(), server)
```
sunu: This will create a single EntityServicer that the threadpool will share?

aleph/logic/extractors/extract.py
```diff
@@ -1,71 +1,51 @@
...
-            log.debug("Loading spaCy NER model...")
-            settings._spacy_model = spacy.load('xx')
-        doc = settings._spacy_model(text)
-        for ent in doc.ents:
-            clazz = SPACY_TYPES.get(ent.label_)
-            if clazz is not None:
-                yield clazz.create(agg, ent.text, ent.start, ent.end)
-    except Exception:
-        log.exception("spaCy failed")
+        text = text[:self.MAX_LENGTH]
```
sunu: Why do we need to trim the text? Just to keep the memory footprint smaller or is there any other reason?

pudo: spaCy has a limit, Polyglot doesn't. gRPC also limits out around about 4MiB. Do you think this should be done in the server?

services/extract-entities/entityextractor/service.py
pudo: That's in the examples -- I'm also a bit "curious" if spaCy is thread-safe....

aleph/logic/extractors/extract.py
sunu: If grpc has a limit, then we have to enforce that limit before shoving it down the wire though. How about sending it in chunks instead of throwing the rest of the text away though?

services/extract-entities/entityextractor/service.py
sunu: From comments from this thread https://github.com/explosion/spaCy/issues/172, it seems spacy does alright.

aleph/logic/extractors/extract.py
```diff
@@ -1,71 +1,54 @@
...
+            for attempt in range(10):
+                try:
+                    service = EntityExtractStub(self.channel)
+                    req = Text(text=text, languages=languages)
+                    for res in service.Extract(req):
+                        clazz = self.TYPES.get(res.type)
+                        yield (res.text, clazz, res.start, res.end)
+                    break
+                except self.Error as e:
+                    if e.code() == self.Status.RESOURCE_EXHAUSTED:
```
sunu: Shouldn't we wait a bit before retrying for this error too?

aleph/logic/extractors/extract.py
```diff
@@ -1,71 +1,54 @@
...
-            clazz = SPACY_TYPES.get(ent.label_)
-            if clazz is not None:
-                yield clazz.create(agg, ent.text, ent.start, ent.end)
-    except Exception:
-        log.exception("spaCy failed")
+        texts = textwrap.wrap(text, self.MAX_LENGTH)
+        for text in texts:
+            for attempt in range(10):
+                try:
+                    service = EntityExtractStub(self.channel)
```
sunu: This will end up recreating the service for each chunk. But I assume the cost is small enough? Didn't want to have multiple try/except blocks.

PR #521, opened 2018-10-17
Small ui fixes

emmina:
- added custom icons 
- put inline text inside of InfoPane
- small css fixes


ui/src/components/Entity/EntityViews.jsx
```diff
@@ -54,8 +52,10 @@ class EntityViews extends React.Component {
           <Tab id="info"
                title={
                   <React.Fragment>
-                    <i className='fa fa-fw fa-info-circle'/>
-                    <FormattedMessage id="entity.info.info" defaultMessage="Info"/>
+                    <Icon name='info' iconSize='14px' className='entity-icon'/>
+                    <span className='tab-padding'>
+                      <FormattedMessage id="entity.info.info" defaultMessage="Info" className='test'/>
```
davo11122: I'ld suggest to change className from `test` to something which will better describe what it is for.

PR #526, opened 2018-10-22
K8s: tag all images with Aleph-version to avoid surprise-upgrades

Ueland:


pudo:
We can add this to [bumpversion.cfg](https://github.com/alephdata/aleph/blob/master/.bumpversion.cfg) so that it automatically gets updated with the latest version. Would that be a good thing (TM)?

Issue #513, opened 2018-10-13
Proposal for UI updates


davo11122:
Hey all.

I've got an improvement ideas which I'ld like to discuss and realize if they make sense.

* [ ] 1. [Search Screen] Side panel seems too big so width can be decreased.
* [x] 2. [Search Screen] Show where search term exactly matches with search result. [react-highlight-words](https://github.com/bvaughn/react-highlight-words) module can be used for frontend-only implementation. - cf. #494 
* [x] 3. Make sure cross-ref pages are being indexed by search engines.
* [ ] 4. [Search Screen] Do not show empty pages like "No search result" or "This folder is empty"(for Cases) instead offer users to "Set an Alert" and "Upload files"
* [ ] 5. [CASES] Extract CROSS-REF, SIMILAR and TAGS for uploaded files
* [ ] 6. [Home Screen] Add a hint which tells to sign in to get more access to more sources.
* [x] 7. [Search Screen] We need preview for photos, may be we can embed photos in tooltip
* [x] 8. Change date format, for example `2016-07-14` becomes `14 July 2017`

Ueland:
Hey, some quick thoughts.

Point 2, see my issue #494 as that has been noted by "my" users.
Point 8, it is important to assume that dates can be incomplete, as in only year, only year+month etc.

pudo:
1. Let's play with the size, but keep it consistent with the other places where DualPane is used.
2. Yes! Let's discuss in #494, but we just need to add `?highlight=true` to the search URL and we'll get snippets from the backend. We need to play with where we're going to fit this in the backend.
3. Let's make a list of SEO problems we have, this may not be the most important one. 
4. Yep. Can we make a SuggestAlert 
5. Bigger issue - we need to figure out how to perform xref for documents in the backend.
6. Yes.
7. I'd rank that low-pri. But keyboard nav on search results would be really cool.
8. What Tor said. Partial dates are important.

pudo:
Can I propose we close this an turn the remaining issues into separate tickets, @davo11122?

PR #530, opened 2018-10-25
Keyboard navigation on search screen

davo11122:
Implements basic keyboard bindings for preview in search screen.

`/` - focus on search box
`j` - preview for next entity 
`k` - preview for previous entity 

Resolves **7.** point in https://github.com/alephdata/aleph/issues/513#issuecomment-429727406


davo11122:
@pudo Thanks for the feedback. I missed that point and fixed in c93031f commit.

PR #537, opened 2018-10-29
Bulk indexing API

pudo:
This is an API suggested in conversation with @Ueland - for writing bulk entities using the REST API. Here's the mechanism:

``POST /api/2/collections/453/_bulk``

The expected body is a JSON array of FollowTheMoney entities with the following fields: 

* ``id`` - an ID, unique to the current collection (best practice is to put the collections foreign_id) into key values as a SHA1.
* ``schema`` - name of an FTM schema, e.g. ``LegalEntity``.
* ``properties`` - a dict with property values


Ueland:
Looking forward to this getting into the master release! Thanks!

Issue #494, opened 2018-09-20
Better search result snippets


Ueland:
Search results now shows some of the text before a search term in the search result snippet, but it stops right there. The result should be in the middle of a snippet to easier understand the context of the search result  for a given document.

pudo:
Now in ``develop``, closing.

PR #540, opened 2018-10-30
Highlighted sentences on search

emmina:
#494

### DIFF
--- a/ui/src/app/App.scss
+++ b/ui/src/app/App.scss
@@ -117,6 +117,10 @@ th.clickable {
   margin-top: $aleph-grid-size * 5;
 }
 
+.collection {
+  padding-top: $aleph-grid-size * 1.5 !important;
+}
+
 //small screen
 @media screen and (max-width: $aleph-screen-sm-max-width) {
   .visible-sm-block {
--- a/ui/src/components/EntityTable/EntityTable.jsx
+++ b/ui/src/components/EntityTable/EntityTable.jsx
@@ -112,17 +112,15 @@ class EntityTable extends Component {
           </tr>
         </thead>
         <tbody className={c({'updating': isLoading})}>
-          {result.results !== undefined && result.results.map(entity =>
-            <React.Fragment>
-              <EntityTableRow key={entity.id}
-                              entity={entity}
-                              location={location}
-                              hideCollection={hideCollection}
-                              showPreview={showPreview}
-                              documentMode={documentMode}
-                              updateSelection={updateSelection}
-                              selection={selection} />
-            </React.Fragment>
+          {result.results !== undefined && result.results.map(entity =>            
+            <EntityTableRow key={entity.id}
+                            entity={entity}
+                            location={location}
+                            hideCollection={hideCollection}
+                            showPreview={showPreview}
+                            documentMode={documentMode}
+                            updateSelection={updateSelection}
+                            selection={selection} />
           )}
         </tbody>
       </table>
--- a/ui/src/components/EntityTable/EntityTable.jsx
+++ b/ui/src/components/EntityTable/EntityTable.jsx
@@ -42,8 +42,6 @@ class EntityTable extends Component {
     this.state = {
       result: props.result
     };
-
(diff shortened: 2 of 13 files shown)


ui/src/app/App.scss
```diff
@@ -118,6 +117,10 @@ th.clickable {
   margin-top: $aleph-grid-size * 5;
 }
 
+.collection {
+  padding-top: $aleph-grid-size * 1.5 !important;
```
davo11122: I'm not sure we need `!important` here, would like to get more details about

ui/src/components/EntityTable/EntityTable.jsx
```diff
@@ -42,6 +42,8 @@ class EntityTable extends Component {
     this.state = {
       result: props.result
     };
+
+    this.getHighlight = this.getHighlight.bind(this);
```
davo11122: No need to bind class methods; instead use I suggest to use `class properties` to attach persist the context.
For more details [see](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties)
Note: By default our babel configuration supports this plugin

ui/src/components/EntityTable/EntityTable.jsx
```diff
@@ -64,12 +66,29 @@ class EntityTable extends Component {
     }
   }
 
+  splitHighlight(split, array) {
```
davo11122: the name `array` doesn't tell anything

ui/src/components/EntityTable/EntityTable.jsx
```diff
@@ -64,12 +66,29 @@ class EntityTable extends Component {
...
+    let highlightArray = array.split(split);
```
davo11122: Consider using `const` for highlightArray

ui/src/components/EntityTable/EntityTable.jsx
```diff
@@ -64,12 +66,29 @@ class EntityTable extends Component {
...
+    let highlights = '';
+    for(let i = 0; i < highlightArray.length; i++) {
```
davo11122: Consider using `Array.prototype.join` method instead of iteration

ui/src/components/EntityTable/EntityTable.jsx
```diff
@@ -113,14 +132,25 @@ class EntityTable extends Component {
...
+            <React.Fragment>
+              <EntityTableRow key={entity.id}
+                              entity={entity}
+                              location={location}
+                              hideCollection={hideCollection}
+                              showPreview={showPreview}
+                              documentMode={documentMode}
+                              updateSelection={updateSelection}
+                              selection={selection} />
+              {entity.highlight !== undefined && entity.highlight.map(function(sentence, index){
```
davo11122: I think the highlight could be part of `<EntityTableRow/>` component;

ui/src/components/EntityTable/EntityTable.jsx
```diff
@@ -113,14 +132,25 @@ class EntityTable extends Component {
...
+                return entity.name !== getHighlight(sentence) && <tr>
+                <td key={entity.id} colSpan="5" className={index !== entity.highlight.length - 1 ? 'highlighted-words' : ''}>
```
davo11122: `key` property must be set to parent element in iteration _(in our case it is `<tr>`)_

ui/src/components/EntityTable/EntityTableRow.jsx
```diff
@@ -27,33 +27,33 @@ class EntityTableRow extends Component {
     }
     
     return (
-      <tr className={rowClassName}>
-        {updateSelection && <td className="select">
+      <tr className={rowClassName} key={entity.id}>
+        {updateSelection && <td className="select" key={entity.id +1}>
```
davo11122: Why do we need `key+n` property on `<td/>`'s

ui/src/components/EntityTable/EntityTable.jsx
pudo: Strongly agree.

PR #544, opened 2018-10-31
UI Performance 

davo11122:
* 1d49a6e Upgraded to react-scripts@2.x 
* c2c0061 8234011 d1c52cb Lazy load for all screens _In a purpose to decrees main bundle size and prevent useless code from loading_
* c0507dc 9be9654 69e4f13 Ejecting `react-scripts`
* d6e4e06 Configured splitChunk plugin for 'commons' and 'vendors' cacheGroups


davo11122:
Thanks for feedback @pudo!
This is a result of using `React.lazy` feature with `react-route`1. See this https://github.com/ReactTraining/react-router/issues/6420 issue  for more details. 
If app works as expected I would suggest to merge with this error for now. 
I think library owners will fix this later.

pudo:
Argh, gotta love upstream issues. Ok, I'll take it :)

PR #559, opened 2018-11-13
FtM implementation

davo11122:


davo11122:
@pudo Thanks very much for your feedback!

>I'd actually instantiate the model object in the metadata reducer and then make the schemata section of the metadata unavailable. Then you'd get a schema object from model.get(schemaName).

Strongly agree! at this moment it implemented like that, `model` is stored under `schemeta` key in metadata reducer, plus instead of using `model.get(schemaName)` I suggest using `model[schemaName]` which implemented via Proxy API so no regression or major refactoring needed.

PR #568, opened 2018-11-27
Search history

davo11122:
Created auto-suggest on search history to easy access 
Created popup on search history to manage
Changed theme `visited-link` color
Removed some death code.
Fixed some logical issues.

### DIFF
--- a/ui/src/actions/queryLogsActions.js
+++ b/ui/src/actions/queryLogsActions.js
@@ -1,0 +1,14 @@
+import asyncActionCreator from "./asyncActionCreator";
+import {queryEndpoint} from "./util";
+import {endpoint} from "src/app/api";
+
+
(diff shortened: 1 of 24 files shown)


davo11122:
1. Completely agree!
2. This is not a real auto-suggest, we just show the history to users. For implementing auto-suggest we need additional endpoint. For now we can just disable the popover to not embarrass users

pudo:
Re 2) My bad, I didn't think it through. Do you think we should do the suggest? As a history log, I think the auto-complete makes less sense, but the suggest thing might be a slow API a bit.

davo11122:
Re: Re: 2 All the majors(e.g FB, google...) do auto-suggest on search history, and I think we can do it as well, but with one update, we can add `Remove` button next to each search term, so users will have have sense of that things are under their control and we are not tracking but just trying to save their time.

PR #580, opened 2018-12-06
Make UI languages configurable

DavidLemayian:
Let users make a choice of UI language configuration similar to Languages.

### DIFF
--- a/aleph/settings.py
+++ b/aleph/settings.py
@@ -145,6 +145,8 @@ LANGUAGES = [l.lower().strip() for l in LANGUAGES]
 
 # User interface
 UI_LANGUAGES = ['ru', 'es', 'de', 'bs', 'en']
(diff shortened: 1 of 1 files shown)


pudo:
I'm not sure this makes sense. Adding a new language to the UI required actual code changes here: 

https://github.com/alephdata/aleph/blob/master/ui/src/app/Translator.jsx#L16

So adding to this list in the ``settings.py`` would just add stuff to the menu that crashes the front-end. 

Do you have specific languages you're looking into translating?

DavidLemayian:
@pudo No, I want to remove some of the languages that show up in the dropdown. Russian doesn't make sense as an option in my side of the world. ;)

pudo:
Aaaah I didn't think about it that way around. Sorry. Then it's useful. 

Also: https://www.ft.com/content/ce7d0fdc-2a7d-11e7-bc4b-5528796fe35c - don't say never :)

PR #582, opened 2018-12-08
Tracing with Opencensus and Stackdriver Trace

sunu:


pudo:
This looks good. Is there any way we can trace ES on a lower level, so that all calls to the index are traced?

sunu:
@pudo Couldn't find anything for es or celery. So I've added some tracing hooks of our own. Should work alright until we find something better.

PR #589, opened 2018-12-14
Login Error Handling

felixebert:
show an error message when user enters an invalid username or password


felixebert:
401 should be appropiate when the request is wellformed and the credentials are wrong, see RFC: https://tools.ietf.org/html/rfc7235#section-3.1
"If the request included authentication credentials, then the 401
   response indicates that authorization has been refused for those
   credentials."

pudo:
Fair enough :)

PR #599, opened 2018-12-27
[Hotfix] Collection toolbar critical bug

davo11122:
### Bug 
The issue was reproducible only in `data.occrp.org`
When the user clicks on `Manage...` button in the collection page
then the whole app crashes into a white page.

### Reason
Debugging showed `Menu.Item` variable as `undefined` which supposed to be an alias for `MenuItem` component `blueprint@3.x.x`. Most probably this is because of missing `package-lock.json` file.

### Resolution
Instead of using `Menu.Item` alias use `MenuItem` component directly.

### for @pudo
 I think event we have a fix for now but we still shall dig deeper into this issue to find the origins.

Big thanks to @uhhhuh for reporting about the bug


pudo:
Looks like they just removed the Menu.Item shortcut. Thanks for the fix.

PR #600, opened 2019-01-08
SEO improvements

davo11122:
* d659e6b Collection pages will be indexed as Dataset
* 965d27d Updated Icon meta description
* 8706bc7 Entity properties are keywords for page
* e4f5baa Property `description` represented as description meta-tag


davo11122:
This is needed to make microdata`itemProp`'s  available in DOM.
I could pass `itemProps`'s manually but as `itemProps` are browser default properties/attributes so for me that feels weird))

PR #631, opened 2019-03-06
 API endpoint for exporting entities and documents from search result

### LINKED ISSUE #584: Bulk export results
User request: the option to download all results from a search, including both structured data entities and source documents. I imagine it would be good to flatten the entities into a set of CSVs or into a multi-sheet Excel document, while source documents should be stored in folders named according to the document ID.

sunu:
Fixes #584


sunu:
https://github.com/alephdata/followthemoney/pull/37 needs to be merged for the tests to proceed

aleph/logic/export.py
```diff
@@ -0,0 +1,89 @@
...
+def export_entities(entities, format):
+    assert format in (FORMAT_CSV, FORMAT_EXCEL)
+    zip_archive = zipstream.ZipFile()
+
+    if format == FORMAT_EXCEL:
+        workbook = get_workbook()
+        for entity in entities:
+            export_entity_excel(workbook, entity)
+            if entity.schema.is_a('Document'):
+                write_document(zip_archive, entity)
```
pudo: Looking at the code for ZipStream, I feel we could basically unwind here and make it so that each document is yielded individually, so that we never keep more than one document in memory. Check this loop, which we could essentially plunder (and then emit the metadata last):

see https://github.com/SpiderOak/ZipStream/blob/master/zipstream.py#L200

aleph/views/entities_api.py
```diff
@@ -33,6 +35,20 @@ def index():
     return EntitySerializer.jsonify_result(result)
 
 
+@blueprint.route('/api/2/search/export/<any(csv, excel):format>', methods=['GET'])  # noqa
+@blueprint.route('/api/2/entities/export/<any(csv, excel):format>', methods=['GET'])  # noqa
+def export(format):
+    parser = SearchQueryParser(request.args, request.authz)
```
pudo: I think this should be restricted to signed-in users.

aleph/views/entities_api.py
```diff
@@ -33,6 +35,20 @@ def index():
     return EntitySerializer.jsonify_result(result)
 
 
+@blueprint.route('/api/2/search/export/<any(csv, excel):format>', methods=['GET'])  # noqa
```
pudo: Let's make just one API endpoint, I'd go for `/api/2/entities/...`.

aleph/views/entities_api.py
```diff
@@ -33,6 +35,20 @@ def index():
...
+@blueprint.route('/api/2/entities/export/<any(csv, excel):format>', methods=['GET'])  # noqa
+def export(format):
+    parser = SearchQueryParser(request.args, request.authz)
+    result = EntitiesQuery.handle(request, parser=parser)
+    results = result.to_dict(serializer=EntitySerializer)['results']
```
pudo: At the moment, this will return only the first `limit` results, but max out at 9999. I think this is unexpected behaviour, we should return a fixed number of max results (e.g. 1000?) and disable the option in the UI if the result total is larger than that. 

/cc @dkhurshudian

aleph/logic/export.py
```diff
@@ -0,0 +1,89 @@
...
+    if 'inode/directory' in filetypes:
+        return
+    ext = mimetypes.guess_extension(filetypes[0], strict=True)
+    if ext:
+        name = name + ext
+    path = os.path.join(parent, name)
+
+    file_url = entity.context['links'].get('file')
+    if file_url:
+        stream = requests.get(file_url, stream=True)
```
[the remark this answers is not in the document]
sunu: Yeah, true. I'll refactor the current archive serving code a bit so that we can reuse it here as well.

aleph/logic/export.py
sunu: We are using this: https://github.com/allanlei/python-zipstream/blob/master/zipstream/__init__.py#L182. Yielding data from individual documents one at a time is already implemented internally.

pudo:
Note to self: https://gist.github.com/chipx86/9598b1e4a9a1a7831054 -

PR #666, opened 2019-05-23
UI improvements 

dkhurshudian:


ui/src/components/Entity/EntityTagsMode.jsx
```diff
@@ -33,6 +33,12 @@ class EntityTagsMode extends React.Component {
   render() {
     const { intl, entity, tags } = this.props;
 
+    if (tags.shouldLoad || tags.isLoading) {
```
pudo: I think this is the wrong place to implement this, because it's still going to show the top menu in an incomplete way. Can we do the section loading thing in `EntityViews` and make it dependent on references, tags etc.?

ui/src/components/common/Entity.jsx
```diff
@@ -30,10 +30,10 @@ class EntityLabel extends Component {
     if (!entity) {
       return null;
     }
-    const title = entity.getFirst('title') || entity.getCaption();
-    const fileName = entity.getFirst('fileName') || title;
+    const title = entity.getCaption();
+    const fileName = entity.getFirst('fileName') || entity.getFirst('title') || title;
```
pudo: this can just be `fileName || getCaption()` if we're using a current version of ftm.js.

ui/src/components/common/Language.jsx
```diff
@@ -28,13 +30,108 @@ class List extends Component {
...
+
+  render() {
+    const { codes } = this.state;
+    const { languages } = this.props;
+    const items = Object.keys(languages);
+
+    return (
+      <BlueprintMultiSelect
+        initialContent={
+          <MenuItem disabled text="asd" />
```
pudo: ??

ui/src/components/common/Language.jsx
```diff
@@ -28,13 +30,108 @@ class List extends Component {
...
+        }
+        itemPredicate={this.itemFilter}
+        noResults={
+          <MenuItem disabled text="sd" />
```
pudo: ??

ui/src/components/common/Language.jsx
dkhurshudian: dumb moment

PR #670, opened 2019-06-06
add http header X-Content-Type-Options: nosniff

zrthstr:
X-Content-Type-Options: nosniff


pudo:
Maybe we want to add it here, too? https://github.com/alephdata/aleph/blob/master/ui/nginx.conf

PR #680, opened 2019-06-26
Streaming CSV viewer

### LINKED ISSUE #677: Whats our plan for rendering CSV in the UI?
So we're switching the way that tabular data is stored in aleph from `followthemoney` entities to having a CSV file that is output and stored. That means we need to be able to preview tabular data on the basis of the CSV file. Two methods come to mind: 

* Parse in JavaScript from URL data. There seem to good CSV parsers for JS, e.g. [PapaParse](https://www.papaparse.com/), but their default premise is that we can load the full file. We don't want that, at most we want to load maybe 1000-10000 rows of data for a preview.
* Or, we could write a backend proxy that reads the CSV from archive storage and converts it into a paged pseudo-API that produces a JSON representation. 

We also need to make a call on whether we treat tables as all having headers, or headers as some sort of optional extra.

sunu:
fixes #677 .

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -31,7 +31,8 @@
     "query-string": "^6.2.0",
     "jwt-decode": "^2.2.0",
     "lodash": "^4.17.11",
-    "numeral": "^2.0.6"
+    "numeral": "^2.0.6",
+    "papaparse": "^5.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


sunu:
Search inside document doesn't work right now.

pudo:
> Search inside document doesn't work right now.

That's absolutely expected and the trade-off we entered into.

PR #729, opened 2019-08-02
Detect content language and use language specific model for NER

### LINKED ISSUE #725: Named Entity Recognition should take collection language and document language into account.
Right now, we use SpaCy's multilingual `xx` model to do all our NER tasks. We should be able to use a language specific model by detecting the language of the text and by taking a hint from the collection languages.

sunu:
Fixes #725. Uses FastText for language identification (https://fasttext.cc/docs/en/language-identification.html)


pudo:
Nice work! What's the overall size for all of the models?

pudo:
Ok, I'm a little bit confused by the actual code, too: how did you come to the conclusion that language would change line by line? Is it efficient for us to do language detection once per text snippet (e.g. document page), rather than for the whole document - perhaps resulting in an array of active languages?

sunu:
`fasttext` is extremely fast. It does line by line language detection on 1000 full length wikipedia articles at 460 ms on average. So that should not be a problem. The collection I downloaded had documents where there were some Arabic paragraphs among mostly English text. So my assumption is if there are text in multiple languages in a doc, we can detect it. If not, the detection is fast enough. So we don't lose much time.

Memory footprint is kinda bad though. Memory used by the fasttext model is ~210 MB. Combined memory footprint of the SpaCy models is ~540 MB.

pudo:
I don't think we need per-line language detection; it's not a unit that we do anything else with. I'd propose we move the language detection up to the document level and also add it as a property to the result (i.e. `entity.add('language', detected)`). Of course we can tag multiple languages for a single document if many are detected between text snippets, but pages seem plenty precise for this. 

I'm not sure how I feel about tripling the memory footprint of aleph this way. Are you sure the fasttext thing is that much better than `langdetect`? Do we need all the spaCy models, or should we just do `en` and `xx`? I just don't want to bloat things out super much over marginal gains.

sunu:
> I don't think we need per-line language detection; it's not a unit that we do anything else with. I'd propose we move the language detection up to the document level and also add it as a property to the result (i.e. `entity.add('language', detected)`). Of course we can tag multiple languages for a single document if many are detected between text snippets, but pages seem plenty precise for this.

Alright. So we go the language detection on`Page` content and set the predominant languages detected on the `Pages` objects?
 
> I'm not sure how I feel about tripling the memory footprint of aleph this way. Are you sure the fasttext thing is that much better than `langdetect`? Do we need all the spaCy models, or should we just do `en` and `xx`? I just don't want to bloat things out super much over marginal gains.

`fasttext` is definitely better than `langdetect` overall. It gives more accurate results, it's from FB, it's well maintained, quite popular and a lot faster. `langdetect` on the other hand is abandoned and can't do multithreading out of the box.

Re: NER models, if we don't have significant percentage of docs in other langs, then `en` and `xx` is fine of course. May be we can add `ru` when we have it.

sunu:
Btw the model size of the `fasttext` model on disk is under 1 MB. I'm not sure if I measured the memory usage wrong or it's extremely well compressed. I'll try to check again.

aleph/analysis/extract.py
```diff
@@ -42,13 +44,30 @@ def location_country(location):
...
 
-def extract_entities(text):
+def get_language(text):
+    """Given a list of lines, return a list of (line, lang)"""
+    if not hasattr(settings, '_lang_detector'):
+        lid_model = FastText()
+        lid_model.load_model(settings.LID_MODEL_PATH)
+        settings._lang_detector = lid_model
+    langs = settings._lang_detector.predict([text])
+    return langs[0]
```
pudo: Are we sure this array always has a length?

sunu: Yes, the result when it doesn't detect anything is a list of empty lists; an empty list for each snippet of text passed in.

PR #733, opened 2019-08-19
WIP: Admin dashboard APIs

sunu:
Depends on https://github.com/alephdata/servicelayer/pull/10


aleph/views/dashboard_api.py
```diff
@@ -0,0 +1,14 @@
...
+
+from aleph.logic.dashboard import get_active_collection_status
+from aleph.views.util import jsonify
+
+blueprint = Blueprint('dashboard_api', __name__)
+
+
+@blueprint.route('/api/2/dashboard', methods=['GET'])
+def dashboard():
+    if request.authz.is_admin:
```
pudo: Alternatively, we could make this read `request.authz.collections` and intersect the collections the user is able to see with the status board. We're gonna need to serve out collection metadata (like their label) anyway to render this as a GUI.

sunu:
> Do we really need to cache stuff from redis into redis? I understand there's a bunch of operations happening in the other thing, but this way we have a dashboard that only refreshes every two minutes..?

The queue code is doing status code for each run. So I was thinking that's a bit too many reads if we have multiple big collections in the queue. But I guess Redis is fast enough. I'll change it back and remove the caching.

pudo:
Nice progress: could we go from `/api/2/dashboard` to `/api/2/status`? Dashboard is a bit of a useless term in the API, it could mean any number of things.

PR #737, opened 2019-08-26
Ignore empty directories

milo-trujillo:
Currently if the secrets directory is empty (as is the case with a brand-new installation), elasticsearch crashes because it is passed a literal asterisk with `elasticsearch-keystore add-file * /secrets/*`.

Desired behavior is to either call elasticsearch-keystore with each keyfile (like `elasticsearch-keystore add-file foo /secrets/foo`), or, if there are no keyfiles, to not call `elasticsearch-keystore add-file` at all.

This change skips the for-loop if the secrets directory is empty.


pudo:
Thanks!

Issue #750, opened 2019-09-05
Bug in language identification


moreymat:
I have a collection of monolingual documents for which the identified language is always wrong (but always the same language).
I was able to reproduce this behavior locally with pyfasttext, providing the same text and getting the same wrong language.

The official fasttext package (https://pypi.org/project/fasttext/) won't accept this text as is though because it contains newlines.
When I replace newlines with a regular space, I get the right language.
When I split the text on newlines, the first line is predicted to be of the same wrong language pyfasttext was returning.

I see at least two possible fixes :
1. Keep pyfasttext and replace newlines, for example with whitespaces:
```python
langs = settings._lang_detector.predict([text.replace('\n', ' ')])
```
2. Or drop pyfasttext that is no longer maintained in favor of the official fasttext package, and replace newlines with an almost identical call:
```python
langs = settings._lang_detector.predict(text.replace('\n', ' '))
```

sunu:
Thanks for filing the issue. I hadn't realized pyfasttext was not the official package. Switching to the official package definitely makes more sense to me.

pudo:
Fixed.

PR #751, opened 2019-09-06
use fasttext, bugfix lang detection

moreymat:
This PR uses the official fasttext package for language detection and fixes bug #750 .


sunu:
Hi @moreymat, thanks for the PR! Looks great. I just changed the base to `develop` branch since that's where we keep the commits before releasing. Would you mind resolving the conflicts? Or else I can resolve and merge. Thanks.

PR #766, opened 2019-10-02
WIP: Mappings API

### LINKED ISSUE #762: Refactor the mappings API
We want to enable our users to define mappings for `Table` entities. The defined mappings can be used to generate FtM entities out of the table content and users can easily cross reference them against other collections.

- Each `MappingQuery` will be against a single `Table` entity. No joins allowed.
- Each `MappingQuery` created by the user will be stored in the db
- We need to assign an ID to each saved `MappingQuery`, and allow users to retrieve all the mappings for a `Collection`, and all the mappings for a `Table`; and perform CRUD operations. API endpoints will look something like:
`/api/2/collections/XXX/mappings`
`/api/2/collections/XXX/mappings?filter:table=YYY`
- Needs a “trigger” operation to start loading entities.
  - Need to be able to stream from own archive
  - Does the table have headers (csv headers)? (Let a UI flag tell us -- is first row a header?)
- Needs a “unload” operation to delete all the loaded entities
- Each FtM entity generated by the mapping gets a `proof` property pointing to the `Table`.
- We will need a new database table, a new view controller. Database schema will look roughly like: `mapping_id`, `collection_id`, `table_id`, `entities_query`: JSON BLOB

We may get rid of the existing mapping API afterwards.

sunu:
Fixes #762

### DIFF
--- a/aleph/model/mapping.py
+++ b/aleph/model/mapping.py
@@ -1,0 +1,87 @@
+from datetime import datetime
+import logging
+
+from normality import stringify
+from sqlalchemy.dialects.postgresql import JSONB
+
+from aleph.core import db
+from aleph.model import Role, Collection
+from aleph.model.common import SoftDeleteModel, ENTITY_ID_LEN
+
+
+log = logging.getLogger(__name__)
+
+
+class Mapping(db.Model, SoftDeleteModel):
+    """A mapping to load entities from a table"""
+    __tablename__ = 'mapping'
+
+    id = db.Column(db.Integer, primary_key=True)
+    entities_query = db.Column('query', JSONB)
+
+    role_id = db.Column(db.Integer, db.ForeignKey('role.id'), index=True)
+    role = db.relationship(Role, backref=db.backref('mappings', lazy='dynamic'))  # noqa
+
+    collection_id = db.Column(db.Integer, db.ForeignKey('collection.id'), index=True)  # noqa
+    collection = db.relationship(Collection, backref=db.backref('mappings', lazy='dynamic'))  # noqa
+
+    table_id = db.Column(db.String(ENTITY_ID_LEN), index=True)
+
+    def update(self):
+        self.updated_at = datetime.utcnow()
+        db.session.add(self)
+        db.session.commit()
+
+    def delete(self, deleted_at=None):
+        self.deleted_at = deleted_at or datetime.utcnow()
+        db.session.add(self)
+        db.session.commit()
+
+    def to_dict(self):
+        data = self.to_dict_dates()
+        data.update({
+            'id': stringify(self.id),
+            'entities_query': dict(self.entities_query),
+            'role_id': stringify(self.role_id),
+            'collection_id': stringify(self.collection_id),
+            'table_id': self.table_id,
+        })
+        return data
+
+    @classmethod
+    def by_id(cls, id, role_id=None):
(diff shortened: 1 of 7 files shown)


aleph/model/mapping.py
```diff
@@ -0,0 +1,87 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class Mapping(db.Model, SoftDeleteModel):
+    """A mapping to load entities from a table"""
+    __tablename__ = 'mapping'
+
+    id = db.Column(db.Integer, primary_key=True)
+    entities_query = db.Column('query', JSONB)
```
pudo: Why the name shift? Is query overloaded somehow?

aleph/views/mappings_api.py
```diff
@@ -0,0 +1,123 @@
...
+def index():
+    require(request.authz.logged_in)
+    parser = QueryParser(request.args, request.authz)
+    q = obj_or_404(Mapping.by_role_id(request.authz.id))
+    result = DatabaseQueryResult(request, q, parser=parser)
+    return MappingSerializer.jsonify_result(result)
+
+
+@blueprint.route('/api/2/mappings', methods=['POST', 'PUT'])
+def create():
```
pudo: All of these actions need to authenticate on the collection. I'm wondering if it wouldn't be a good idea to prefix the API with `collections/xx` for that reason alone? At the moment, the security on this makes no sense.

aleph/views/mappings_api.py
```diff
@@ -0,0 +1,123 @@
...
+@blueprint.route('/api/2/mappings/<int:mapping_id>', methods=['GET'])
+def view(mapping_id):
+    require(request.authz.logged_in)
+    mapping = obj_or_404(Mapping.by_id(mapping_id, role_id=request.authz.id))
+    return MappingSerializer.jsonify(mapping)
+
+
+@blueprint.route('/api/2/mappings/<int:mapping_id>', methods=['DELETE'])
+def delete(mapping_id):
+    require(request.authz.session_write)
```
pudo: This gives anonymous users write access. It just checks if the system is in write mode.

aleph/views/mappings_api.py
```diff
@@ -0,0 +1,123 @@
...
+    mapping = obj_or_404(Mapping.by_id(mapping_id, role_id=request.authz.id))
+    mapping.delete()
+    db.session.commit()
+    return ('', 204)
+
+
+def _get_mapping_query(mapping):
```
pudo: Shouldn't this live in `aleph.logic` somewhere?

aleph/views/mappings_api.py
```diff
@@ -0,0 +1,123 @@
...
+    if 'table' in parser.filters:
+        q = Mapping.by_table(parser.filters['table'], q)
+    result = DatabaseQueryResult(request, q, parser=parser)
+    return MappingSerializer.jsonify_result(result)
+
+
+def _load_query():
+    try:
+        query = request.json.get('mapping_query', '{}')
+        # TODO: validate query
```
pudo: I wonder if we should do a refactor in `followthemoney` that makes it validate both mappings and schemata against a written-out JSONSchema spec. At the moment, the validation is a little ad-hoc and procedural, and if we have JSONSchema files for these formats they might also come in handy with the swagger docs. WDYT?

PR #768, opened 2019-10-02
Mapping UI + API

pudo:


aleph/logic/mapping.py
```diff
@@ -0,0 +1,113 @@
...
+    queries = keys_values(config, 'queries', 'query')
+    mapping_id = config.get('mapping_id')
+    mapping = Mapping.by_id(mapping_id)
+    config['entity_ids'] = []
+    for query in queries:
+        try:
+            entity_ids = load_mapping_query(stage, collection, hash_data(query), query)  # noqa
+            config['entity_ids'].extend(entity_ids)
+        except Exception as exc:
+            if mapping:
```
pudo: What is our strategy here for removing the old "bulkload" - are we doing it or not? Seems like the API is gone but the CLI command and logic layer support is still there...

aleph/logic/mapping.py
```diff
@@ -0,0 +1,113 @@
+import logging
+from banal import first
+from flask import request
+from werkzeug.exceptions import BadRequest
+from followthemoney import model
+from banal import hash_data, keys_values
+
+from aleph.core import archive, db
+from aleph.views.util import get_index_entity, get_session_id
```
pudo: Logic layer should never import from `from aleph.views`, that's a sign of something being in the wrong place. In this case I'd say the `flush_mapping` and `load_mapping` functions want to live in the API.

aleph/views/mappings_api.py
```diff
@@ -0,0 +1,100 @@
...
+@blueprint.route('/api/2/collections/<int:collection_id>/mappings/<int:mapping_id>', methods=['POST', 'PUT'])  # noqa
+def update(collection_id, mapping_id):
+    collection = get_db_collection(collection_id, request.authz.WRITE)
+    mapping = obj_or_404(Mapping.by_id(mapping_id))
+    data = parse_request(MappingSchema)
+    entity_id = data.get('table_id')
+    query = load_query()
+    entity = get_index_entity(entity_id, request.authz.READ)
+    mapping.update(query=query, table_id=entity.get('id'))
+    if request.args.get('flush') == 'true':
```
pudo: There are two ways to trigger flush and load, both as a query argument and a separate endpoint. I prefer the separate endpoint and would get rid of the flags. Then we can also move the logic of `flush_mapping` and `load_mapping` into the view functions more easily.

aleph/index/entities.py
```diff
@@ -201,3 +206,32 @@ def delete_entity(entity_id, exclude=None, sync=False):
...
+        }
+    }
+    query = {
+        'query': query,
+        '_source': _source_spec([], []),
+        'size': MAX_PAGE
+    }
+    result = es.search(index=index, body=query)
+    # log.debug(result)
+    for doc in result.get('hits', {}).get('hits', []):
```
pudo: I think this could easily use `aleph.index.collections:delete_entities` via an additional argument for `mapping_id`. That would still leave the aggregator, which we may want to drop entirely....

aleph/logic/mapping.py
sunu: I think the CLI command is still handy for loading mappings files directly and we should keep it. We're using it in tests in a couple of places to quickly load entities without creating a Mapping object which is also somewhat useful.

PR #776, opened 2019-10-04
API documentation, api input validation and api response validation in tests based on OpenAPI 3 / JSONSchema

sunu:
Some notes about implementation:
- [`api-spec.yml`](https://github.com/alephdata/aleph/blob/sunu/api-docs/api-spec.yml) contains OpenAPI 3 specs for Aleph's API
- We use [`openapi2jsonschema` tool](https://github.com/alephdata/openapi2jsonschema) to pick out schema definitions for api request and response data and convert them into JSONSchema. These generated JSONSchema files live in https://github.com/alephdata/aleph/tree/sunu/api-docs/aleph/schemas. The generation can be done by running `make build-schema`
- Then [`jsonref`](https://pypi.org/project/jsonref/) is used to use the generated JSONSchema to do api input validation in the view functions and api output validation in tests.
- The conversion tool doesn't cover all edge case differences between OpenAPI 3 and JSONSchema. So we have to use the subset of syntax that is recognized by both the schema standards.


aleph/tests/test_ingest_api.py
```diff
@@ -31,6 +31,7 @@ def test_upload_no_meta(self):
     def test_upload_csv_doc(self):
         _, headers = self.login(is_admin=True)
         meta = {
+            'collection_id': self.col.id,
```
pudo: Isn't this already in the URL?

pudo:
The simplification in #862 worked, closing.

PR #821, opened 2019-10-23
Compute name frequency distribution

pudo:
The goal of this PR is to compute a frequency score for a composite name (like "Vladimir Vladimirovitch Putin") that would indicate how frequent that name is in the full corpus of the system. This score can then be used by the cross-referencer as an input to its comparison function to predict the quality of an entity match.

The first idea was to just compute the frequency of a name token (i.e. occurrences of "Vladimir" vs. total names) but that isn't a useful thing to normalise over. Instead, this PR tries to compute a frequency distribution for all names in aleph proactively, and then finds the frequency percentile for each name. This gives a more useful number, but it currently requires keeping all name tokens in RAM. At the very least this should happen via redis, but even for that it's unclear if we're meeting the cost/benefit ratio here.

There's also an open issue around how we combine the percentiles for multiple name tokens in a name. For example "Vladimir Vladimirovitch Putin" would decompose into three tokens, each of which has a frequency. Do we add them up? Multiply? Average?


pudo:
Manually "rebased" this, now in core.

Issue #886, opened 2019-11-25
VIS Integration


kjacks:
The primary goal for the first iteration of the integration of VIS into Aleph is to provide the basic infrastructure to support users' ability to create/edit/delete VIS diagrams within the context of a personal collection.  The diagrams will reference entities stored in the db, and any edits to the diagram should persist these changes to the entities themselves.

#### This iteration will explicitly not cover
* Editing of VIS diagrams for non-personal collections
* Canonical entity related features (integration of entities from across Aleph, merging, etc.)

#### Access Control
By storing the diagram as a 'Diagram' entity member of an existing collection, read/write access will be inherited from the collection permissions, so we shouldn't have to implement any additional permissions management.

### API
* Creation of 'Diagram' model to store JSON representation of VIS diagrams
* CRUD API endpoints for Diagrams
* JSON Schema to validate Diagram objects
* Diagram creation will involve:
  - Creating entities in the database; will reuse the `create_entity` function
  - Replacing entity ids with the signed version.
  - Storing only references to the created entities in the JSON blob
* Serializing Diagrams will involve:
  - De-referencing entity ids to include them inline in the JSON data being returned

#### Screens

* __*Collection Views > Diagrams Index Screen*__
  * create/edit/delete of diagrams for the collection
* __*Personal Settings Area > My Diagrams Index Screen*__
  * create/edit/delete of diagrams for the given user
  * (all diagrams must be tied to an existing collection)
* __*Diagram Edit Screen*__
  * full screen VIS window below the Navbar
  * visibility of toolbars/editing components controlled by user's write permssions
  * existing collection entities passed to vislib as datasource
  * delete button, .vis download button

#### Aleph <-> Vislib Message Passing
Vislib modifications required: See [alephdata/vislib - #120](https://github.com/alephdata/vislib/issues/120) 

##### Instantiation
Diagram view component in Aleph must pass the following parameters to instantiate a VisGraph component:
* `config`: GraphConfig object (can be default config params for now)
* `layout`: GraphLayout object reference
* `viewport`: Viewport object reference (could mayyyybe be made optional/ created internally within vislib)
* `updateLayout`: function called whenever layout is updated
* `exportSvg`: function to handle writing svg to file 
* `isEditable`: boolean (to be implemented)
* `datasource`: Datasource object (to be implemented)

##### Update frequency
* should follow Google Docs style delayed (5s ?) autosave
* whenever updateLayout function is called, sends delayed update API request

pudo:
[...]

PR #891, opened 2019-11-27
VIS Diagrams in API & UI

sunu:
Refs #886, also alephdata/vislib#120.


aleph/model/diagram.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+class Diagram(db.Model, SoftDeleteModel):
+    """A mapping to load entities from a table"""
+    __tablename__ = 'diagram'
+
+    id = db.Column(db.Integer, primary_key=True)
```
pudo: Can we also have a `label`?

aleph/validation/schema/diagram.yml
```diff
@@ -0,0 +1,115 @@
+DiagramCreate:
+  type: object
+  required: ["data"]
+  properties:
+    data:
+      type: object
+      allOf:
+      - $ref: '#/components/schemas/DatedModel'
```
pudo: I think that wants to move up one level.

aleph/logic/diagram.py
```diff
@@ -0,0 +1,105 @@
...
+        proxy = obj.to_proxy()
+        obj.delete()
+        index.delete_entity(entity_id)
+        refresh_entity(proxy)
+
+
+def delete_diagram(diagram):
+    existing_ids = diagram.entity_ids
+    # Clean up created entities
+    for ent_id in existing_ids:
```
pudo: I reckon this should either be a flag, or maybe not even exist at all :/

sunu:
> Question: what's the up- and down-side of making `entities` an ARRAY on the table? I assume we can query within the data blob, too, but this might be more convenient.
> 
I have this property function to access the entity ids in a diagram: https://github.com/alephdata/aleph/blob/0f7784de36b4a60ba45e8958d62bb4b0605dfc40/aleph/model/diagram.py#L64 My reasoning behind storing all the data in one place is just to keep the data as close to the original VIS format as possible. I don't have any strong reasons for or against it.

> There's something icky about letting VIS choose database entity IDs. These need to be globally unique (they're the table's primary key), and we could provoke conflicts here easily (i.e. it's an attack vector). I'm almost curious if VIS should negotiate this with Aleph via a hook in VIS' entity creation routines...

Right now, we are throwing away the id given by VIS and making our own id: https://github.com/alephdata/aleph/blob/0f7784de36b4a60ba45e8958d62bb4b0605dfc40/aleph/model/entity.py#L81 So the blob returned by the backend after creation has completely different ids that the input blob.

aleph/validation/schema/diagram.yml
sunu: It's a type. Should be `DiagramData` instead.

aleph/model/diagram.py
sunu: Yes, makes sense

aleph/logic/diagram.py
sunu: Ok, I'll make it a flag for now. If we don't have it at all, is there any other user-facing way to delete stray entities later?

aleph/validation/schema/diagram.yml
```diff
@@ -0,0 +1,115 @@
...
+              position:
+                type: object
+                properties:
+                  x:
+                    type: number
+                  y:
+                    type: number
+              entityId:
+                type: string
+        edged:
```
kjacks: edged => edges

aleph/validation/schema/diagram.yml
```diff
@@ -0,0 +1,115 @@
...
+                properties:
+                  x:
+                    type: number
+                  y:
+                    type: number
+        groupings:
+          type: array
+        selection:
+          type: array
+    viewport:
```
kjacks: @sunu I'm thinking to actually only store the layout information, not the viewport.  But curious if you have any thoughts on this.  The viewport is only storing the bounding box/zoom level so would only be necessary to store if we wanted to show a user the same view of the diagram that they had left previously

sunu: In that case, omitting viewport seems reasonable. Do you think there's anything else we should not store? May be selection (I assume that's storing the current selection)?

aleph/model/diagram.py
kjacks: And maybe a description?

PR #899, opened 2019-12-02
Improve SMTP settings

jpilet:
- Add ALEPH_MAIL_TLS option;
- Remove unused ALEPH_MAIL_ADMIN option from aleph.env.tmpl;
- Set default in aleph.env.tmpl to TLS, not SSL.
- add ALEPH_MAIL_DEBUG option


pudo:
Makes perfect sense. Welcome to the contributors!

PR #915, opened 2019-12-24
Add CORS support.

### LINKED ISSUE #914: CORS support for reconciliation endpoints
The reconciliation endpoint is only available via JSONP at the moment. It would be great to enable it for CORS too.

We are currently planning to phase out JSONP in the reconciliation API: see https://github.com/reconciliation-api/specs/issues/19.

Support for CORS will be added in OpenRefine: https://github.com/OpenRefine/OpenRefine/issues/2260

wetneb:
Fixes #914.

This adds CORS for all origins on all endpoints which use the `jsonify` helper. Given that JSONP support was already implemented there, these endpoints were already accessible from any origin via JSONP, so this should not change the policy.

### DIFF
--- a/aleph/views/util.py
+++ b/aleph/views/util.py
@@ -172,6 +172,11 @@ def jsonify(obj, status=200, headers=None, encoder=JSONEncoder):
         data = '%s && %s(%s)' % (cb, cb, data)
         # mime cf. https://stackoverflow.com/questions/24528211/
         mimetype = 'application/javascript'
+    headers = headers or {}
+    # CORS support
+    headers['Access-Control-Allow-Methods'] = 'GET, POST, PUT, DELETE, OPTIONS'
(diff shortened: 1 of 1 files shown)


pudo:
Hey! Thanks for submitting the CORS headers. I'm a bit confused by the whole issue with the reconciliation test bench: we do have CORS support (cf. `settings.py`, `core.py`) and I'm suspicious that it's actually our CSP rules that prevent the request. Are you aware of any tooling more verbose than Chrome to debug the reasons the requests are being aborted?

wetneb:
Thanks! The reason is that although `aleph.occrp.org` does have CORS enabled, `data.occrp.org` redirects to it without having the appropriate headers, so the request is blocked before being redirected. I have updated Wikidata with the new endpoints and the testbench now confirms that CORS is supported.

So this PR is not needed!

Issue #953, opened 2020-01-23
Logging: Configure level & handler in aleph.env


simonwoerpel:
_As an admin for production environments, i want to access logfiles on disk with all the errors in it._

- set up new environment variables and their defaults in aleph.env to configure the python-logging (level & handler) of the different services:
`ALEPH_LOGGING_LEVEL='error'`
`ALEPH_LOGGING_HANDLER='file'`
`ALEPH_LOGGING_PATH='/var/logs/aleph/'`
- adapt these settings in the services for the already existing logging mechanism:
  - worker
  - ingestor

simonwoerpel:
decided to only set the log level & format via env vars, the handler (console, file, whatever) should be configured on docker level

PR #960, opened 2020-01-24
Make log level & format configurable via env vars

simonwoerpel:
implements this: https://github.com/alephdata/aleph/issues/953


pudo:
Thanks for the PR. At the moment this sort of erases all the overrides for the logger, which were set specifically to keep the log output readable. I'd propose we keep them in place unless there is a super strong reason the print every SQL query, ES response etc...

simonwoerpel:
thx for the comment. you're totally right, i commited now without the level overrides. we could discuss about setting the different specific loggers via env vars itself, but for now i'll keep it simple

pudo:
`servicelayer` is now building 1.9.8 which contains the logging config handler.

PR #967, opened 2020-01-28
Diagrams UI + API

kjacks:
Refs #886

Initial integration with @alephdata/vislib

Known issues:
- [ ] inconsistent entity property ordering (ftm issue)
- [ ] diagram import currently only reliably imports vertices and their associated Thing entities, associated entities are not handled (awaiting backend fix)
- [ ] scoped search on Diagram Editor should probably clear its input when the diagram selection is changed
- [ ] better indication of search results / lack of results for a diagram  
- [ ] table editor allows creation of entities with empty name properties
- [ ] internationalization of vislib 
- [ ] better property type checking within vislib for numeric inputs, etc.


ui/src/reducers/entities.js
```diff
@@ -20,6 +21,8 @@ export default createReducer({
 
   [queryEntities.COMPLETE]: (state, { result }) => resultObjects(state, result),
 
+  [undeleteEntity.COMPLETE]: (state, { id, data }) => objectLoadComplete(state, id, data),
+
   [deleteEntity.COMPLETE]: (state, { id }) => objectDelete(state, id),
 
 }, initialState);
```
pudo: I think we may need to also hook this into the `mutation` reducer to flush front-end cache....

ui/src/screens/DiagramIndexScreen/DiagramIndexScreen.jsx
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  componentDidMount() {
+    this.fetchIfNeeded();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { result } = this.props;
+
+    if (result.shouldLoad && !prevProps.result.shouldLoad) {
```
pudo: I'd move that into fetchIfNeeded ;)

ui/src/selectors.js
```diff
@@ -207,6 +219,10 @@ export function selectCollectionMappings(state, collectionId) {
   return selectObject(state, state.collectionMappings, collectionId);
 }
 
+export function selectCollectionDiagrams(state, collectionId) {
+  return selectObject(state.collectionDiagrams, collectionId);
```
pudo: Why is that an object, does that not hard-code a pagination limit?

ui/src/components/Diagram/DiagramListItem.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+import {
+  Collection, Date, Diagram,
+} from 'src/components/common';
+
+import './DiagramListItem.scss';
+
+const DiagramListItem = ({ diagram, showCollection }) => (
+  <div className="DiagramListItem" key={diagram.id}>
+    <Link className="DiagramListItem__link" to={`/diagrams/${diagram.id}`}>
```
pudo: getDiagramLink?

ui/src/components/common/Diagram.jsx
```diff
@@ -0,0 +1,71 @@
...
+      {!truncate && content}
+    </div>
+  );
+};
+
+
+class DiagramLink extends PureComponent {
+  render() {
+    const { diagram, className } = this.props;
+    const link = `/diagrams/${diagram.id}`;
```
pudo: getDiagramLink?

aleph/index/entities.py
```diff
@@ -201,7 +202,16 @@ def delete_entity(entity_id, exclude=None, sync=False):
...
-        es.delete(index=index, id=entity_id,
-                  refresh=refresh_sync(sync))
-        q = {'term': {'entities': entity_id}}
-        query_delete(entities_read_index(), q, sync=sync)
+        try:
+            es.delete(index=index, id=entity_id,
+                      refresh=refresh_sync(sync))
+            q = {'term': {'entities': entity_id}}
+            query_delete(entities_read_index(), q, sync=sync)
+        except NotFoundError:
```
pudo: For me to understand, the goal here is explicitness?

aleph/logic/collections.py
```diff
@@ -102,7 +102,7 @@ def delete_collection(collection, keep_metadata=False, sync=False):
     deleted_at = collection.deleted_at or datetime.utcnow()
     Entity.delete_by_collection(collection.id, deleted_at=deleted_at)
     Document.delete_by_collection(collection.id)
-    Mapping.delete_by_collection(collection.id)
+    Diagram.delete_by_collection(collection.id)
```
pudo: Did you mean to get rid of the mapping delete?

aleph/model/diagram.py
```diff
@@ -0,0 +1,91 @@
...
+    def update(self, data=None):
+        self.updated_at = datetime.utcnow()
+        if data:
+            self.label = data.get('label', self.label)
+            self.summary = data.get('summary', self.summary)
+            self.entities = data.get('entities', self.entities)
+            self.layout = data.get('layout', self.layout)
+        collection = Collection.by_id(self.collection_id)
+        for ent_id in self.entities:
+            collection.ns.verify(ent_id)
```
pudo: I think you could just do `.sign` here?

aleph/model/entity.py
```diff
@@ -71,8 +75,9 @@ def to_proxy(self):
             'schema': self.schema,
             'properties': self.data
         })
-        proxy.add('name', self.name)
-        proxy.set('indexUpdatedAt', self.updated_at)
+        if proxy.schema.is_a('Thing'):
+            proxy.add('name', self.name)
+            proxy.set('indexUpdatedAt', self.updated_at)
```
pudo: I think there are more entity types with indexUpdatedAt. Let's just pass quiet=True to avoid errors.

aleph/tests/test_diagrams_api.py
```diff
@@ -0,0 +1,186 @@
...
+
+from aleph.tests.util import TestCase
+from aleph.views.util import validate
+from aleph.logic.entities import create_entity
+from aleph.logic.collections import delete_collection
+
+log = logging.getLogger(__name__)
+
+
+def _normalize_data(data):
```
pudo: Almost wonder if we want something like that in ftm. `proxy.cleanup()`?

aleph/tests/test_entities_api.py
```diff
@@ -351,3 +350,204 @@ def test_entity_tags(self):
         assert len(results) == 1, results
         assert results[0]['value'] == '+491769817271', results
         validate(res.json['results'][0], 'EntityTag')
+
+    def test_undelete(self):
```
pudo: Looking at it now: is undelete the same as update? Should we calm down and make update do an undelete by default?

aleph/validation/schema/alert.yml
```diff
@@ -15,7 +15,7 @@ Alert:
     normalized:
       type: string
       readOnly: true
-    writable:
+    writeable:
```
pudo: outch. That was me I think.

aleph/validation/schema/diagram.yml
```diff
@@ -0,0 +1,132 @@
...
+  required: ["collection", "role_id", "label", "layout", "entities", "writeable"]
+
+Layout:
+  type: object
+  required: ["vertices", "edges"]
+  properties:
+    vertices:
+      type: array
+      items:
+        type: object
```
pudo: Should we give these names? DiagramVertex, DiagramEdge, etc.?

aleph/views/diagrams_api.py
```diff
@@ -0,0 +1,178 @@
...
+      tags:
+        - Diagram
+    """
+    require(request.authz.logged_in)
+    parser = QueryParser(request.args, request.authz)
+    collection_id = first(parser.filters.get('collection_id'))
+    if collection_id:
+        get_db_collection(collection_id)
+        q = Diagram.by_collection(collection_id).filter(Diagram.role_id == request.authz.id)  # noqa
+    else:
```
pudo: That seems like a false A or B to me. Can people either see their own diagrams, or those in a collection? I think this might be cleaner if we prepare a SQLA q object that filters for the user's collections, and then drill it down with collection and role filters.

aleph/views/serializers.py
```diff
@@ -312,3 +313,60 @@ def _serialize(self, obj):
 
 class MappingSerializer(Serializer):
     pass
+
+
+class VisDiagramEntitySerializer(EntitySerializer):
```
pudo: Let's not use the term Vis in the backend. DiagramEntity.. is fine :)

pudo:
Another thought: let's make sure we can package all the diagram things into a separate build bundle, so it gets lazy-loaded the first time a user opens a diagram-related screen.

ui/src/components/Diagram/DiagramEditor.jsx
```diff
@@ -0,0 +1,197 @@
...
+      this.downloadDiagram();
+    }
+  }
+
+  updateLayout(layout, propagateUpdate = false) {
+    const { diagram, onStatusChange } = this.props;
+    this.setState({ layout });
+
+    if (propagateUpdate) {
+      onStatusChange(updateStates.IN_PROGRESS);
```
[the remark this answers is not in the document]
kjacks: @pudo  As in, could it gather the status from the global state/ reducers instead of storing its own internal status state?

ui/src/components/Diagram/DiagramImport.jsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  render() {
+    const { importedFile } = this.props;
+
+    return (
+      <Dropzone
+        accept=".vis"
+        onDrop={acceptedFiles => (
+          acceptedFiles && acceptedFiles.length ? this.onDrop(acceptedFiles) : null
```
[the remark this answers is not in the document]
kjacks: @pudo Definitely

ui/src/components/Diagram/DiagramEditor.jsx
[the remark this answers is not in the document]
pudo: I think I just struggle to understand it -- it's sort of a meta-variable expressing the overall status of multiple backend actions? That seems reasonable...

ui/src/components/Diagram/DiagramListItem.scss
```diff
@@ -0,0 +1,62 @@
...
+    color: $blue2;
+    text-decoration: none !important;
+  }
+
+  &__content.bp3-card {
+    height: 100%;
+    padding: 35px $aleph-content-padding;
+    position: relative;
+    cursor: pointer;
+    box-shadow: 0 0 0 1px $aleph-border-color,
```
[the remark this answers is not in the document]
kjacks: Yaa I don't love this, but the bp folks use a different grey for the elevations that doesn't exist in the rest of their color palette, so it results in borders different from the rest of the site. :(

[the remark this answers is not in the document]
kjacks: I'll at least make it more reusable though

ui/src/components/Diagram/util.js
```diff
@@ -0,0 +1,64 @@
...
+  return { generatedEntities, generatedLayout };
+};
+
+// convert any entity references contained in entity properties to entityIds instead of
+//  full Entity objects
+const processApiEntity = (entity) => {
+  const { properties } = entity;
+  if (properties) {
+    Object.entries(properties).forEach(([key, values]) => {
+      properties[key] = values.map((value) => (value?.id ? value.id : value));
```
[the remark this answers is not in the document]
kjacks: @pudo I couldn't find one, but would be happy to add it to ftm if you think it'd be better living there

ui/src/components/common/Diagram.jsx
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      <span className="DiagramLabel" title={diagram.label}>
+        {icon && <Icon icon="graph" className="left-icon" />}
+        <span>{diagram.label}</span>
+      </span>
+    );
+  }
+}
+
+const DiagramSummary = ({ className, diagram, truncate }) => {
```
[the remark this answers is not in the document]
kjacks: Absolutely - will change

ui/src/selectors.js
kjacks: Good catch, forgot to remove this selector when we changed the api

aleph/index/entities.py
sunu: Yes, for explicitness and we also want to skip the `query_delete` step which is slow and doesn't do anything where the entity is already deleted once

aleph/logic/collections.py
sunu: Nope, mistake while merging I guess

aleph/model/diagram.py
sunu: Done

aleph/model/entity.py
sunu: The problem I was trying to avoid is that `name` is not a property for entities that are not `Things`. Added a check to see if `name` is defined first. Or may be I should check if the `Schema` has a `name` property? What's the best way to do that in FtM?

pudo: Just adding `, quiet=True` to the add and set calls will make it set the value iif the prop exists.

aleph/tests/test_entities_api.py
sunu: if we only `undeleted` the entity then I agree that it could be done inside `update` endpoint. But we create new entities too. So undelete is more of an `upsert`. It seems very random to do that in `update` if we're doing CRUD according to REST in all other places.

aleph/validation/schema/diagram.yml
sunu: Done

aleph/views/diagrams_api.py
sunu: done

aleph/views/serializers.py
sunu: done

PR #969, opened 2020-01-28
Document PDF Viewer - Rotation & Page Input Controls

### LINKED ISSUE #943: Rotate document
I would like to be able to rotate document pages when reading documents in the "View" mode.

kjacks:
Resolves #943 

- Rotation buttons for PDF Viewer
- Page numeric input for PDF Viewer

Screenshot 2020-01-28 at 14 54 14


pudo:
Looks like the linter is rebelling :/

I sort of wonder if it would make sense to move the next page thing behind the rotate buttons, so it's at either extreme. What do you think?

PR #1004, opened 2020-02-11
Fix sorting by date

### LINKED ISSUE #961: Sorting by date broken
Sez Aubrey:

>  I don't know if you noticed this, but Aleph is not arranging things by date like it used to. (Maybe other ways of sorting stuff are also broken I don't know)
>
> I wanted to demonstrate to a conference how we used Aleph on the XXX dataset. One useful thing we did was to select only images, then use the "date" tab to sort them chronologically. That allowed us to easily make I timeline. I tried to do it again just now and it doesn't work. It's just blank. Not sure what's up.

sunu:
Fixes #961

### DIFF
--- a/aleph/index/entities.py
+++ b/aleph/index/entities.py
@@ -1,6 +1,6 @@
 import logging
 import fingerprints
-from pprint import pprint  # noqa
+from pprint import pprint, pformat  # noqa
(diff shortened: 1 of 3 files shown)


pudo:
Ha, that's a legendary find! Very cool.

PR #1005, opened 2020-02-12
Disable google stackdriver stuff if `GOOGLE_APPLICATION_CREDENTIALS` …

simonwoerpel:
hey, I am not familiar with this google stuff, so I couldn't test if anything breaks with this change if you want to use it. But from the code logic, it shouldn't :see_no_evil: 

But this fix speeds up the development server reload on code changes A LOT :laughing:


pudo:
Fixed via #1022

PR #1030, opened 2020-02-27
Collection Index Sorting

### LINKED ISSUE #966: Sort collections by different criteria
As a user I want to pick what to order collections by in a listing in the UI: 

* Number of entities (current default)
* Newest
* Oldest 
* Search result score

kjacks:
Resolves #966 

Implements sorting of collections on the collection index screen, cases index screen, and group screen based on creation date, update date, title, and size.

@pudo I ended up removing the sort by search relevance option for the moment.
Because it is implemented currently as the default filter on the request it was causing two issues:
- no support for ascending/descending option (why anyone would want their least relevant search results first is a mystery to me, but for consistency / not to break the ui it probably still needs to be an option.
- the code gets a bit messy if we want the default filter to be creation date, but the null filter option sorts by relevance

Happy to talk about this / wait until we find a solution to merge


pudo:
Just played with it, I think the implementation is beautiful. Let's call it a fix :)

PR #1040, opened 2020-03-04
use pandoc over openoffice/http to convert documents into pdf's

critocrito:
This pull request is a humble attempt to use pandoc over convert-document/openoffice for document conversion to pdf. Using pandoc would get rid of the convert-document service as a whole, document conversion happens as a sub- process of the file-ingestors themselves.

Unfortunately, I couldn't test the code since I didn't manage to get the docker dev environment to run. I thought to share the changes despite the lack of testing to see if there is principal interest from your side.

`pandoc` has to be installed and in path. The `PYPANDOC_PANDOC` environment variable can be used to point to the location of the `pandoc` binary.

### DIFF
--- a/services/ingest-file/Dockerfile
+++ b/services/ingest-file/Dockerfile
@@ -93,6 +93,8 @@ RUN apt-get -qq -y update \
         libtiff-tools ghostscript librsvg2-bin \
         # pdf processing toolkit
         poppler-utils poppler-data pst-utils \
+        # document conversion
+        pandoc texlive \
     && apt-get -qq -y autoremove \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
--- a/services/ingest-file/ingestors/support/convert.py
+++ b/services/ingest-file/ingestors/support/convert.py
@@ -1,11 +1,10 @@
 import math
 import logging
 import requests
-from requests import RequestException, HTTPError
+import pypandoc
 from servicelayer.util import backoff, service_retries
 from followthemoney.helpers import entity_filename
 
-from ingestors.settings import CONVERT_URL
 from ingestors.support.cache import CacheSupport
 from ingestors.support.temp import TempFileSupport
 from ingestors.exc import ProcessingException
@@ -14,7 +13,7 @@ log = logging.getLogger(__name__)
 
 
 class DocumentConvertSupport(CacheSupport, TempFileSupport):
-    """Provides helpers for UNO document conversion via HTTP."""
+    """Provides helpers for document conversion via pandoc."""
 
     def document_to_pdf(self, file_path, entity):
         key = self.cache_key('pdf', entity.first('contentHash'))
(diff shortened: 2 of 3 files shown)


pudo:
I'm a little bit torn on this: while it's probably more stable than the current solution using convert-document, it does involve translating any incoming documents to LaTeX, so most of the contained layout information is going to be lost. People will see a preview of a LaTeX doc, not the original document as seen by others.

I'm curious what @simonwoerpel @felixebert @sunu think about this trade-off...

simonwoerpel:
> I'm a little bit torn on this: while it's probably more stable than the current solution using convert-document, it does involve translating any incoming documents to LaTeX, so most of the contained layout information is going to be lost. People will see a preview of a LaTeX doc, not the original document as seen by others.
> 
> I'm curious what @simonwoerpel @felixebert @sunu think about this trade-off...

I'd prefer to see the documents in their original layout (or at least as close to it as possible)

But there might be different cases / needs for different users / scenarios: Sometimes it's not so important to have the original layout (e.g. when you know your documents are mainly just text in word and no crazy ppt...) 

So what about using pandoc as an (optional) fallback? Like after 3 failed retries for the uno service?

One could control this via settings / env vars, so let the users decide the trade-off based on ther specific needs :slightly_smiling_face:

critocrito:
I understand the reservation regarding layout. I did a quick test and pandoc indeed drops a lot of the layout, I tried other engines than LaTeX as well (khtml, ConTeXt) but the results were even worse. 

The current approach is very unreliable. So I played with a CLI usage of LibreOffice to do conversion (`soffice --headless --convert-to pdf pandoc-test.docx`). I attached the source docx and the converted PDF. This converts documents as a regular process and doesn't require any socket to listen on. So it could be done inside the ingestors, the same way as the above patch does. Have you tried this approach in the past?

[pandoc-test-soffice.pdf](https://github.com/alephdata/aleph/files/4338399/pandoc-test-soffice.pdf)
[pandoc-test.docx](https://github.com/alephdata/aleph/files/4338400/pandoc-test.docx)

sunu:
> The current approach is very unreliable. So I played with a CLI usage of LibreOffice to do conversion (`soffice --headless --convert-to pdf pandoc-test.docx`). I attached the source docx and the converted PDF. This converts documents as a regular process and doesn't require any socket to listen on. So it could be done inside the ingestors, the same way as the above patch does. Have you tried this approach in the past?
> 
I assume the downside to this approach is that starting and shutting down soffice for every single conversion will add quite some overhead when converting large numbers of documents. With the current approach, we keep the soffice service alive so we don't have to start it for each single request.

critocrito:
Yeah that's true. I timed the conversion of the test file. It was a second slower to convert the file using the CLI approach over the current `convert-document` approach:

```sh
root@fish ~ # time soffice --headless --convert-to pdf pandoc-test.docx
convert /root/pandoc-test.docx -> /root/pandoc-test.pdf using filter : writer_pdf_Export

real	0m1.217s
user	0m0.844s
sys	0m0.212s
root@fish ~ # time curl -o /dev/null -w "%{http_code}" -F format=pdf -F 'file=@pandoc-test.docx' http://localhost:3000/convert 2> /dev/null
200
real	0m0.219s
user	0m0.006s
sys	0m0.004s
root@fish ~ #
```

pudo:
We've got another round of feedback from @felixebert, indicating that the set of document types supported by `pandoc` is much more limited than with LibreOffice. We're therefore looking at an alternative solution of running LO fresh for each document as an option. 

Sorry for not merging, and thanks for writing this PR, @critocrito.

PR #1045, opened 2020-03-12
Include table csv link in serialized mapping object

sunu:
cc @kjacks 

Some notes:
- The `csv_url` in the generated mapping yaml is a signed Google Storage URL and it expires after a while (24 hours by default). So the mapping when run with ftm cli will not work after that period.
- When used in dev mode, local filesystem is used for blob storage. So the csv_url is a file path inside the container. So if the user wants to run the mapping with ftm cli, they'll have to do it inside a container (`make shell` will give you one)
- I'm not quite sure about the return type. Should I directly return the yaml content or should I return it as json? or is it fine as it is?
- Also, the only additional data the endpoint is returning is the signed url for the csv file. May be if we include that in the mapping view response, we won't need a separate endpoint for export? Wdyt @kjacks?

### DIFF
--- a/aleph/logic/mapping.py
+++ b/aleph/logic/mapping.py
@@ -1,5 +1,4 @@
 import logging
-import yaml
 from banal import first
 from followthemoney import model
 
@@ -15,8 +14,8 @@ from aleph.logic.notifications import publish
 log = logging.getLogger(__name__)
 
 
-def make_mapper(collection, mapping):
-    table = get_entity(mapping.table_id)
+def get_table_csv_link(table_id):
+    table = get_entity(table_id)
     properties = table.get('properties', {})
     csv_hash = first(properties.get('csvHash'))
     if csv_hash is None:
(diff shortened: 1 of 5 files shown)


kjacks:
@sunu This looks great!

- As long as we say the url will expire somewhere in the ui I think that should be fine
- The parsing from json -> yaml is a pretty trivial one-line function so whichever is easiest to return works for me.  Json seems more consistent maybe?
- Yeh, it seems like, like in other endpoints where there's a 'links' key with any associated exports for the entity, a similar pattern could be used for the mappings?  So the export wouldn't have to be a separate endpoint

sunu:
The links under links key often point to other api endpoints or pre-computed converted formats already stored on blob storage (csv from excel files etc). I'm not sure we should pre-compute and store the yaml for mappings.

If it's easy enough to convert json to yaml on the frontend, I propose we just stick the `table_csv_url` in links and then the frontend will have all the data it needs to make the mapping yaml.

The entities endpoint already gives out the csv url for the source table but it includes the user's api key. So I think it's better to use the direct blob storage url instead.

PR #1058, opened 2020-03-18
Entity expand api

### LINKED ISSUE #1002: Networks: expand node to show relationships
Use case: As a user drawing a VIS diagram, I want to be able to double-click an entity and expand it to see all of the connections it has in the collection backend but that may not yet be shown in the diagram. 

Problems:
a. Entities can have many (thousands) links, so we may need a pre-filter dialog
b. We need to make an API that returns links, but links aren't really a thing in Aleph's data model
c. Theoretically, we could even support property-value based expansions.

As an attempt to deal with b., I added a graph backend to FtM over Christmas: `followthemoney.graph:Graph`. This could be sub-classed by Aleph to implement entity and adjacency retrieval from ElasticSearch, possibly with a generous dose of redis caching in the middle. 

We should go about designing this by getting a good understanding of our desired user experience, and then making a modest API to serve it. Otherwise this is gonna be full GraphQL support quick :)

sunu:
Fixes #1002 

ToDo:

- [x] Return expanded entities as serialised FtM Graph object
- [x] Use fixed set of types for property value based expansion (names , address, identifier , email, phone, iban, checksum, url)


sunu:
The endpoints don't handle expanding property value based nodes. Should we handle those as well? If we do, then we should probably move the endpoints to something like `/diagrams/node/<node_id>/expand/` or something. Thoughts?

aleph/logic/graph.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class AlephGraph(Graph):
+    def queue(self, id_, proxy=None):
+        if id_ not in self.proxies:
+            if proxy is None:
+                entity = index.get_entity(id_)
+                proxy = model.get_proxy(entity)
+            if proxy is not None:
+                self.proxies[id_] = proxy
```
pudo: I'm curious if the actual retrieval for this could not be done in `resolve` with `entities_by_id` so that we do only one query instead of N?

aleph/logic/graph.py
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+
+def expand_entity_graph(entity, properties=None, authz=None):
+    edge_types = [registry.name.name, registry.email.name,
+                  registry.identifier.name, registry.iban.name,
+                  registry.phone.name, registry.address.name,
+                  registry.url.name, registry.checksum.name,
+                  registry.entity.name]
+    graph = AlephGraph(edge_types=edge_types)
```
pudo: Does edge_types want to be an object instead of a string?

aleph/validation/schema/entity.yml
```diff
@@ -137,3 +137,58 @@ EntityTag:
...
+  type: object
+  properties:
+    nodes:
+      type: array
+      items:
+        $ref: '#/components/schemas/EntityGraphNode'
+    edges:
+      type: array
+      items:
+        $ref: '#/components/schemas/EntityGraphEdge'
```
pudo: Maybe the response type to this should be some extended form of  EntityExpandStats, where things are grouped by property? Otherwise a straight list of entities could also work...

aleph/logic/entities.py
```diff
@@ -113,12 +118,76 @@ def entity_tags(entity, authz=None):
...
+def entity_expand_nodes(entity, properties=None, include_entities=False, authz=None):  # noqa
+    proxy = model.get_proxy(entity)
+    schema = proxy.schema
+    facets = []
+    reversed_properties = []
+    literal_value_properties = []
+    matchable_prop_types = [
+        registry.name, registry.email, registry.identifier, registry.iban,
+        registry.phone, registry.address, registry.url, registry.checksum,
+    ]
```
pudo: We have to many of these lists. Let's discuss this and come up with a clear plan.

PR #1065, opened 2020-03-21
Re-adopt create-react-app

pudo:
We ejected CRA for unknown reasons. This is creating a new CRA project, which should mitigate the risk of running out of sync with their tooling.


pudo:
Yay, looks like CRA itself is a bit broken now: https://github.com/facebook/create-react-app/issues/8688

PR #1081, opened 2020-03-27
Network diagrams: Entity suggest

### LINKED ISSUE #1001: Networks: Suggest entities to add to diagram
When using the dialog to add an entity to a network diagram in Aleph, I want the tool to suggest entities to me which already exist in the dataset but not on the diagram. The API for this should already exist but we need the UI to prompt it for suggestions.

We may also want to make suggestions search only the entity name, not other fields -- for the results to be less magical.

kjacks:
Resolves #1001


pudo:
One feedback: should this send in a list of the entity IDs already on the diagram to the API to exclude them from potential matches?

PR #1125, opened 2020-04-27
API rate limiting

pudo:
Current plan is to limit anonymous users who overly stress the API by rejecting their requests with an HTTP error.

### DIFF
--- a/aleph/views/context.py
+++ b/aleph/views/context.py
@@ -6,11 +6,11 @@ from banal import hash_data
 from datetime import datetime
 from flask_babel import get_locale
 from flask import request, Response, Blueprint
-from servicelayer.rate_limit import RateLimit
 from werkzeug.exceptions import TooManyRequests
 
 from aleph import signals, __version__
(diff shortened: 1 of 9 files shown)


aleph/views/context.py
```diff
@@ -66,11 +116,23 @@ def setup_request():
...
 @blueprint.after_app_request
 def finalize_response(resp):
     """Post-request processing to set cache parameters."""
-    generate_request_log(resp)
+    # Compute overall request duration:
+    took = time.time() - request._begin_time
+
+    # Finalize reporting of the rate limiter:
+    if hasattr(request, 'rate_limit') and request.rate_limit is not None:
+        usage = request.rate_limit.update(amount=math.ceil(took))
```
sunu, line 131: Is `amount=took` there to charge users more for heavier requests?

pudo: Yes -- the specific issue is the `/tags` API still, which sometimes takes more than a second to execute.

PR #1126, opened 2020-04-28
Single user mode

zrthstr:
Add SINGLE_USER=true to aleph.env.
when True, authentication&authorization are disabled, any anonymous user become admin.

### DIFF
--- a/aleph/logic/notifications.py
+++ b/aleph/logic/notifications.py
@@ -45,10 +45,6 @@ def flush_notifications(obj, clazz=None, sync=False):
 def get_role_channels(role):
     """Generate the set of notification channels that the current
     user should listen to."""
-    # role is None in singel user mode
-    # TODO: is this the best way to solve this?
-    if role is None:
-        return [GLOBAL]
     key = cache.object_key(Role, role.id, 'channels')
     channels = cache.get_list(key)
     if len(channels):
@@ -70,12 +66,7 @@ def get_notifications(role, since=None, parser=None):
     filters = [{'terms': {'channels': channels}}]
     if since is not None:
         filters.append({'range': {'created_at': {'gt': since}}})
-    ### TODO: this is a workaround for the notifictations http/500 on in single user mode.
-    ### Is this to ugly to exist?
-    if role is None:
-        must_not = []
-    else:
-        must_not = [{'term': {'actor_id': role.id}}]
+    must_not = [{'term': {'actor_id': role.id}}]
     query = {
         'size': 30,
         'query': {'bool': {'filter': filters, 'must_not': must_not}},
--- a/aleph/logic/notifications.py
+++ b/aleph/logic/notifications.py
@@ -45,6 +45,10 @@ def flush_notifications(obj, clazz=None, sync=False):
 def get_role_channels(role):
     """Generate the set of notification channels that the current
     user should listen to."""
+    # role is None in singel user mode
+    # TODO: is this the best way to solve this?
+    if role is None:
+        return [GLOBAL]
     key = cache.object_key(Role, role.id, 'channels')
     channels = cache.get_list(key)
     if len(channels):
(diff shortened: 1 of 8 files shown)


aleph/settings.py
```diff
@@ -81,6 +81,8 @@
 OAUTH_AUTHORIZE_URL = env.get('ALEPH_OAUTH_AUTHORIZE_URL')
 OAUTH_UI_CALLBACK = env.get('ALEPH_OAUTH_UI_CALLBACK', '/oauth')
 
+# No authentication. Everyone is admin.
+SINGLE_USER = env.to_bool('SINGLE_USER')
```
pudo, line 85: Ok, so I'd prefix the actual env var with `ALEPH_`, just not the variable derived from it in `settings` (like all the other vars in this file).

aleph/views/base_api.py
```diff
@@ -72,6 +73,13 @@ def metadata():
         'model': model,
         'auth': auth
     }
+
+    if settings.SINGLE_USER:
+        role = Role.load_cli_user()
+        authz = Authz.from_role(role)
+        single_user_token = authz.to_token(role=role)
+        data.update({'token':single_user_token})
+
```
pudo, line 82: Good, that's perfect.

ui/src/reducers/roles.js
```diff
@@ -23,5 +24,7 @@ export default createReducer({
 
   [updateRole.COMPLETE]: (state, { id, data }) => objectLoadComplete(state, id, data),
 
+  [fetchMetadata.COMPLETE]: (state, { metadata }) => {  if (!metadata?.token) { return state } else { return handleLogin(state, metadata.token);}},
```
pudo, line 27: Is this duplicative with `session.js`?

ui/src/reducers/session.js
```diff
@@ -29,4 +31,9 @@ const handleLogout = state => ({
 export default createReducer({
   [loginWithToken]: handleLogin,
   [logout]: handleLogout,
+  [fetchMetadata.COMPLETE]: (state, {metadata}) => {
+    if (!metadata?.token) {return state}
```
pudo, line 35: Nice. You could also build a check into handleLogin to  check if it gets a token, keep the code in there rather than here.

aleph/logic/notifications.py
```diff
@@ -66,7 +70,12 @@ def get_notifications(role, since=None, parser=None):
     filters = [{'terms': {'channels': channels}}]
     if since is not None:
         filters.append({'range': {'created_at': {'gt': since}}})
-    must_not = [{'term': {'actor_id': role.id}}]
+    ### TODO: this is a workaround for the notifictations http/500 on in single user mode.
+    ### Is this to ugly to exist?
```
pudo, line 74: These work-arounds feel like they're addressing a secondary problem of some sort, rather than the root cause. What is the scenario where a user hits `/notifications` without the token being set?

zrthstr: Agree that it feels wrong. Still unsure what would be correct.

>> What is the scenario where a user hits /notifications without the token being set?
Any request in SINGLE_USER mode ends in http/500

/ failed with status code 500 and:
|   File "/aleph/aleph/logic/notifications.py", line 78, in get_notifications
api_1               |     must_not = [{'term': {'actor_id': role.id}}]
api_1               | AttributeError: 'NoneType' object has no attribute 'id'

/search ends with:
DETAIL:  Key (role_id)=(26) is not present in table "role".
api_1               | [SQL: INSERT INTO query_log (query, session_id, role_id, created_at) VALUES (%(query)s, %(session_id)s, %(role_id)s, %(created_at)s) RETURNING query_lo

pudo: Oooh I have a suspicion where this is going. Check what `Role.load_cli_user()` does, you may find that while it returns a valid user role it does not in fact commit the SQL transaction and thus the row is deleted after the `metadata` call returns. Maybe you just need to add `db.session.commit()` to the section where you add the token to the metadata.

ui/src/reducers/roles.js
zrthstr: yes, this is duplicate because `handleLogin` is duplicative.

aleph/logic/notifications.py
zrthstr: done

ui/src/reducers/roles.js
zrthstr: deduplicated

ui/src/reducers/session.js
zrthstr: done

PR #1130, opened 2020-04-29
Table Editor

kjacks:


pudo:
Impressive work, @kjacks! I'm collecting some random feedback here: 

## Table feedback: 

* Top-align cell contents and/or truncate long text fields to a pop-over. This may also be a problem for cells with many values, so the truncation should be based on their combined length, not the individual? 
* When I'm in the menu for adding a property it gets reset with each collection status refresh.
* Entity suggest filters for `schema`, should be `schemata`. This was I cannot easily add people to a property that is for `LegalEntity`. 
* Entities get sorted after creation, lose focus (because listing is reloaded?).
* I'm wondering if invalidating all of the redux cache after each property edit is probably evil. Maybe we can make an exception here? Can we batch/debounce edits a bit? The whole thing is crazy sluggish because of the reloads. 

## More broad considerations

* I'm still a fan of untangling the document and entity tabs
* At some point we need to think about how to do fewer calls to `/status`.

kjacks:
@pudo ok just pushed a bunch of changes here. Those that specifically address the feedback above are:

1. Cell values now have a max-height and scroll overflow within the cell (I implemented a truncate-to-popup version as well but then opted for the scroll instead, let me know what you think though...)

2. I refactored the underlying data structure of the table, so now the rendering should be sooo much faster, and it will only re-render what is necessary on updates, so status polling shouldn't cause re-renders anymore

3.  I changed the vislib suggestions function to allow for sending schemata instead of schema, but I'm still a bit confused about how, given a property, to best get its children schema types (i.e. to get People/Companies from a LegalEntity).  Currently FTM schemata only have the extends prop, but this is the opposite of what we need in this case right?

4. Entities now aren't re-sorted after editing, but still are re-shuffled after being created. This is due to the fact that the table editor has to re-render on other changes in entity length (deletion, progressive loading), and there is no reliable way to tell that a change in the incoming entities array is due to adding a new row. I think I have a solution to this, but it needs another hour or so to get home.

5. (see #2 above) should be fixed now :)

6. Documents and Entities tabs are now separated

PR #1132, opened 2020-04-30
Keycloak dev

dschulz-pnnl:
A docker compose for development that also loads a Keycloak docker and configures Aleph to use Keycloak for OAuth authentication.

For the UI, use a manually configured proxy (ui/src/setupProxy.js) that proxies all **/api** requests to http://api:5000 instead of just non _text/html_ requests, which is the behavior of the _**"proxy": "http://api:5000/"**_ option.


pudo:
Thank you for writing this up and sharing your compose file, Doug!

PR #1150, opened 2020-05-07
CSV Parser Refactor

kjacks:
- separates csv loading from view components
- catches cases where documents do not have csv/file links or where parser returns empty data


pudo:
Nice, thanks for looking into it!

PR #1204, opened 2020-06-15
most pages are rtl compatible

lepstep:


ui/src/app/blueprint-overrides.scss
```diff
@@ -35,3 +36,69 @@
...
+
+table {
+  tr td, tr th {
+    @include rtl(text-align, left, right);
+    &.numeric, &.number {
+      @include rtl(text-align, right, left);
+    }
+    // Float on the right, so that numbers overflow to the left.
+    // (FormattedNumber creates a <span>)
+    & > span {
```
kjacks, line 59: I think this float rule needs to only apply to `.numeric` and `.number`, not to all nested `span` elements

ui/src/components/Navbar/Navbar.scss
```diff
@@ -51,7 +52,7 @@
   }
 
   &__left-group {
-    padding-right: $aleph-grid-size;
+    @include rtlSupportInvertedProp(padding, right, $aleph-grid-size, inherit);
```
kjacks, line 55: For the LTR value I think this should be 0, not inherit.  Right now it's adding extra left padding in the LTR view

ui/src/components/Dashboard/Dashboard.scss
```diff
@@ -14,9 +15,9 @@ $dashboard-padding: $aleph-grid-size * 2.5;
         flex-direction: column;
         flex-grow: 0;
         width: 20%;
-        border-right: 1px solid $aleph-border-color;
-        margin-right: $dashboard-padding;
-        padding-right: $dashboard-padding;
+        @include rtlSupportInvertedProp(border, right, 1px solid $aleph-border-color, inherit);
+        @include rtlSupportInvertedProp(margin, right, $dashboard-padding, inherit);
+        @include rtlSupportInvertedProp(padding, right, $dashboard-padding, inherit);
```
kjacks, line 20: All of these values should be 0 instead of inherit for their default value :)

kjacks:
Hey @lepstep - this is looking great so far! 
I like the general structure and how you designed the mixins for sure :)

- I noticed that you're setting some CSS props to "inherit" where they were previously not set before, instead of 0 or "unset" -> which is causing a bit padding and margins to be added to the LTR view where they weren't set before.
- There seem to be some little issues with icon margins in various sections the site, but I'm guessing that won't be too crazy to set via the mixins :)
- There's a strange error with the icon rendering when in RTL mode (TypeError: this.isInternal is not a function) that I will take a look at and try to debug! (it's really quite weird though...)

ui/src/app/blueprint-overrides.scss
lepstep: Right!
this basically meant all nested span elements within a td. This was an already existing rule we had.
I moved it back to tables.scss now and I specify a span with a class for the <FormattedNumber> case

PR #1229, opened 2020-06-23
Filter and facet on date values

### LINKED ISSUE #721: Date range API spec
For an upcoming interactive, we need to support the filtering and faceting of date values. This ticket will serve as a place to discuss the API for this, which might also serve as a template of similar advanced filters for other data types (i.e. numbers). 

Current filters:

`?filter:collection_id=2000`

Date fields: 

`dates`, `properties.startDate`, `properties.date`, `properties.birthDate`  etc. 

Filter pptions:

* `?filter:ge:properties.startDate=2001&filter:le:properties.startDate=2007`
* `?filter@ge:properties.startDate=2001&filter@le:properties.startDate=2007`
* `?filter!ge:properties.startDate=2001&filter!le:properties.startDate=2007`

Facet options:

* `?facet=properties.startDate&facet_interval.properties.startDate=month` or `?facet=properties.startDate&facet_interval.properties.startDate=day`

We already have similar facet options, like `facet_size:schema=20` and `facet_total:schema=true`.

What does the most extendable and intuitive approach for this look like?

Refs #288.

sunu:
fixes #721 

The url looks like `/entities?filter:schema=Message&filter:gte:properties.date=2018-03-01&filter:lte:properties.date=2018-03-05&facet=properties.date&facet_interval:properties.date=day`

Uses `date_histogram` aggregation with (calendar intervals)[https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#calendar_intervals] for faceting. So we can facet on minutes up to years.

### DIFF
--- a/aleph/index/util.py
+++ b/aleph/index/util.py
@@ -125,6 +125,9 @@ def field_filter_query(field, values):
         return {'ids': {'values': values}}
     if field in ['names']:
         field = 'fingerprints'
+    if field.startswith(('gt:', 'gte:', 'lt:', 'lte:')):
+        op, field = field.split(':', 1)
+        return {'range': {field: {op: values[0]}}}
     if len(values) == 1:
         # if field in ['addresses']:
         #     field = '%s.text' % field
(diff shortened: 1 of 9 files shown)


pudo:
Wow, that was a fast bit of hacking, @sunu. Looks nice, too :) @kjacks - can you hear the call of d3 in the distance??

PR #1234, opened 2020-06-26
Simplify casefile handling

pudo:


ui/src/components/Dashboard/Dashboard.jsx
```diff
@@ -178,8 +178,9 @@ class Dashboard extends React.Component {
 const mapStateToProps = (state, ownProps) => {
   const { location } = ownProps;
   const groupsQuery = queryGroups(location);
+  const caseFilter = {'filter:category': 'casefile',};
```
kjacks, line 181: Not important at all, but that trailing comma is weird right?

PR #1236, opened 2020-06-29
Added Cognito as option for OAuth

bakrale:
This adds Amazon Cognito as an identity provider for OAuth2. I look forward to any feedback you may have.


aleph/oauth.py
```diff
@@ -7,9 +7,14 @@
 from cryptography.x509 import load_pem_x509_certificate
 from cryptography.hazmat.backends import default_backend
 from servicelayer.extensions import get_entry_point
+from jose import jwt as jwt_jose
+from jose import jwk
+from jose.utils import base64url_decode
```
pudo, line 12: I think we already have two implementations of JWT in use, I'd like to avoid a third. Can you check out https://docs.authlib.org/en/latest/jose/index.html instead? I'm pretty sure authlib also has solutions for some of the key parsing issues you're solving below.

aleph/oauth.py
```diff
@@ -77,6 +82,65 @@ def handle_google_oauth(provider, oauth_token):
...
+        if kid == keys[i]['kid']:
+            key_index = i
+            break
+    if key_index == -1:
+        log.warning('Public key not found in jwks.json')
+        return False
+    #Construct Public Key
+    public_key = jwk.construct(keys[key_index])
+    message, encoded_signature = str(token).rsplit('.', 1)
+    decoded_signature = base64url_decode(encoded_signature.encode('utf-8'))
```
pudo, line 99: Are you sure there isn't library support for doing this?

aleph/oauth.py
```diff
@@ -77,6 +82,65 @@ def handle_google_oauth(provider, oauth_token):
...
+    from aleph.model import Role
+    superuser_group = 'Administrators'
+    #Get keys
+    keys = json.loads(urlopen(settings.OAUTH_CERT_URL).read())['keys']
+    #Verify tokens
+    id_token = get_claims_cognito(oauth_token.get('id_token'),keys,verify_aud=True)
+    access_token = get_claims_cognito(oauth_token.get('access_token'),keys)
+
+    groups = set(access_token.get('cognito:groups',[]))
+    user_id = 'cognito:{}'.format(id_token.get('cognito:username'))
```
pudo, line 131: Can you even get an ID from cogito? We messed this up in Keycloak, but if there's an ID it's strongly recommended to use it.

aleph/oauth.py
```diff
@@ -77,6 +82,65 @@ def handle_google_oauth(provider, oauth_token):
...
+    #Construct Public Key
+    public_key = jwk.construct(keys[key_index])
+    message, encoded_signature = str(token).rsplit('.', 1)
+    decoded_signature = base64url_decode(encoded_signature.encode('utf-8'))
+
+    if not public_key.verify(message.encode("utf8"), decoded_signature):
+        log.warning('Signature verification failed')
+        return False
+
+    claims = jwt_jose.get_unverified_claims(token)
```
pudo, line 105: Can this use: https://docs.authlib.org/en/latest/jose/jwt.html#jwt-payload-claims-validation ?

bakrale:
Thanks for your feedback, I reworked the function to use authlib as suggested, which actually simplified the code a lot! I also added one more config option for specifying the administrator group. I didn't modify keycloak, but that may also be able to take advantage of that settings variable instead of hardcoding 'superuser'. Let me know what you think!

pudo:
Yes I really like that additional config option. Wanna quickly throw it into KC and maybe shorten the name from ADMINISTRATOR to ADMIN? Otherwise I'm happy to do it. 

Thanks a lot for the nice re-work, ready to merge.

bakrale:
Done!

PR #1249, opened 2020-07-01
converted entity delete processing to sequential instead of parallel

### LINKED ISSUE #1238: Delete UI should delete documents in sequence, not in parallel
The backend here sucks, so if we could chain the promises, rather than running them in parallel, it would make this much more stable.

kjacks:
Fixes #1238 

Also added some progress reporting as it's deleting...


pudo:
Thanks a lot :)

PR #1278, opened 2020-07-20
Support translation of document text and display of the translation in the UI

dschulz-pnnl:


pudo:
Sorry for the slow feedback on this. I really like the way it's integrated in the UI, but I wonder about the benefit/cost of merging it all into the `ingest` stage (vs. the other proposed translation implementation: https://github.com/alephdata/aleph/tree/main/services/translate ).

While doing it all in one go guarantees some performance gain, it makes the whole thing a bit monolithic and forces us to fold language detection into the `ingest` in a few different places, rather than running it later in the `analyze` step. On the minus side, if we had translation as a separate stage, we'd probably have to do some magic related to individual `Page` objects (e.g. by emitting `detectedLanguage` for them as well in the `analyze` stage...).

dschulz-pnnl:
The property names have been changed from "altBodyText" to "translatedText".

Issue #959, opened 2020-01-23
Delayed data exports


pudo:
Exporting bulk files, such as Xref Excels, search result bundles etc. takes an unrealistically long time at the moment. Instead, we should generate those files offline and send an email to the user once the process has completed, including a download link.

sunu:
We decided that:
1. We will generate these exports in the background and upload them to cloud storage
2. We'll store export related data - cloud storage hash, user, export type, export time, file size, deleted or not etc in a database table
3.  If the file is unreasonably big (more that 10GB), we'll abort after export and send an email explaining that to the user
4. If the export process is successful, we'll send an email to the user with a link to download the file.
5. The download link will be behind auth
6. The file is auto-deleted from cloud storage after a while (30 days)

pudo:
Ups, fixed in 3.9.0, closing :)

PR #1282, opened 2020-07-22
Background export

sunu:
refs #959


pudo (changes requested):
This looks amazing, overall! Let me suggest some points:

* I think we ought to emit signed export URLs (like with `archive_url`, especially in the email. Lets have a helper function, `export_url` in logic.util.

* I think "PUBLISH_EXPORT" is a misnomer, given that we'll want to implement actual publication at some stage later. Maybe "COMPLETE_EXPORT"?

* The ROLE_PREFIX thing: would it be nicer to make both prefixed: collections and roles? This way SL 'datasets' clearly can be both. 

* I think `convertUTCDateToLocalDate` wants to live in a util helper file :) Should we use it for notifications or is that somehow not an issue there?

* Kudos on the ref-counting in should_delete :) 

* Can you merge the migrations (just turn into one file and use the max() and min() hashes) and get rid of the `downgrade()` contents? 

Really impressive work.

sunu:
ingestors tests keep breaking in the CI for some reason. I don't know why yet.

pudo:
That's setuptools, I suspect

pudo (approved):

Issue #1267, opened 2020-07-13
Document Entities and Profiles


mynameisfiber:
This issue proposes a solution for two very linked feature requests in aleph: document entities and entity profiles.

Document entities encompasses the desire to have any outputs of the **document NER analysis create real Aleph entities**. Currently, the tokens extracted from the document are stored as text attached to the `Document` entity and are second-class citizens. By turning them into true entities, they can be x-ref'd, used in diagrams and more easily accessed by users. In order to accomplish this, a new type of entity, the `DocumentExtract` entity, is going to be created that can be xref'd on but otherwise will be invisible to the system. This means that the entity could be linked with another entity and shown in the profile view, described below, but otherwise won't clutter the system. The reason for doing this is that any NER extraction is bound to be noisy and create many pathological entities. By making them invisible unless explicitly linked to an existing entity, we allow users to control when these extracted entities are considered "relevant" to their Aleph use-case.

Entity profiles is a way of **clustering entities which have been linked together through the xref** system (and subsequently merged through the Linkage system by a user). By having clusters of entities we have confirmation represent the same physical thing, we have the ability to show in the UI a unified profile representing all the information we know. This is useful in the case that we have multiple entities for, say, Putin, but each entity contains different information. The merged profile will show _all_ of the information in one place and make available a single entity for easier analysis.

An additional requirement of the profile system is that **profiles are namespaced to a dataset**. A dataset is considered a "workspace" for a particular project and thus can have it's own requirements of the considerations when determining whether two entities are "the same". For example, one project may think that every branch of a bank should be merged into the same profile while another project wants to consider each branch separately. By namespacing the profiles to a dataset, we can still reference any entity (from all of Aleph) without potentially stepping on other projects notion of similarity.

To implement this, we've identified the following steps:

- Merge `Linkage` data model into `EntitySet` and `EntitySetItem` models, adding a `score` and `added_by` field to the `EntitySetItem`
- Update `Linkage` API to use the new models
- Add profile membership information into Aleph entity
- Update FTM to add `DocumentExcerpt` family of entities to represent extracted tokens from a document
- Create `DocumentExcerpt` entities from tokens extracted from a document in `ingest-file`
- Ensure `DocumentExcerpt` entities are xref'd on
- Update xref UI to display profile data when the xref target entity is already part of a profile

In the near future, we can also add the following features using the above system:

- Use profile data in the xref calculation
- Use profile data to boost search relevancy
- Create profile view to replace or enhance the single-entity view

[...]

PR #1283, opened 2020-07-22
Migration of former "Linkage" system to EntitySets

mynameisfiber:
Implementing the first parts of #1267.

In particular, this pull request:

- [x] Migrates the `decide_xref` method to use EntitySets
- [x] Update EntitySets to have extra fields to support profiles
- [x] Extend EntitySet API
- [x] Finalize modified xref api response
- [ ] ~Introduce profile-specific API endpoints~
- [ ] ~Introduce profile reference in entity index? Maybe save this for another PR~


pudo:
I've pushed some little commits here: https://github.com/alephdata/aleph/tree/feat/entity-linkage 

My main concern at this stage is having multiple active (set_id, entity_id) rows at the same time. While this can of course be resolved via time, it feels like too much of a mind-bender to me, to the point where I cannot tell after an hour of staring at it if this line is correct (I don't think so):

https://github.com/alephdata/aleph/pull/1283/files#diff-c0250c321eeaf3f8a8353bffb063846dR35

I'm wondering if we shouldn't just humble out and mark any previous judgements `deleted_at` (and thus making sure that there can only be one active (set, entity) pair at any time.

mynameisfiber:
@pudo ah yes, this is the curse of the `SoftDelete` model. We do indeed "soft delete" old (set_id, entity_id) entries before adding new ones (https://github.com/alephdata/aleph/pull/1283/files#diff-c0250c321eeaf3f8a8353bffb063846dR42-R48), but that bit of code you references should be using the `SoftDelete.all()` method (or explicitly making sure `deleted_at == None`).

PR #1303, opened 2020-07-31
Expand entityset API

pudo:
This expands the entity set API like this:

* `GET /api/2/entitysets/YADA/items` - returns a database-backed, paginated index of the entity set items, including judgement and authorship metadata
* `POST /api/2/entitysets/YADA/items` - lets you create or update one item, for example to modify the judgement. If the judgement is set to `no_judgement`, that implements item deletion. (is that too weird??)
* `GET /api/2/entitysets/YADA/entities` - returns a search-backed entities endpoint for search on all items with judgement `positive`. 
* `POST /api/2/entitysets/YADA/entities` - this is where it gets weird. This actually implements full entity upsert, but the entities added here will be added to the entity set, too. 

I haven't currently implemented an explicit `DELETE` API. Deletion can be performed by doing `POST /api/2/entitysets/YADA/items` with a JSON body like this:

```
{"entity_id": "banana", "judgement": "no_judgement"}
```

### DIFF
--- a/aleph/model/common.py
+++ b/aleph/model/common.py
@@ -1,6 +1,7 @@
 import uuid
 import logging
 from datetime import datetime, date
+from sqlalchemy import false
 
 from aleph.core import db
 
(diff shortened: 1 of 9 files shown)


pudo:
@kjacks this is ready for you.

PR #1314, opened 2020-08-09
Fix for Docker on Windows

nabla-c0d3:
This allows the elasticsearch container to start on Windows. Without this fix, it fails on startup due to k8s-entrypoint.sh having Windows-style line endings (which the bash running in the container doesn't understand). 

Most of the project then seems to run fine (including the API), except for the UI which "fails to compile".


pudo:
Thanks, I'll merge now. Kinda surprised this is working on Windows at all - definitely not a target environment we've ever tried for... :)

PR #1318, opened 2020-08-12
Add tests and type annotations to authentication logic

nabla-c0d3:
Hello, this PR:

* Adds type annotations to the authentication logic.
  * This allows for gradually adding more type annotations in the rest of the code.
  * This allows finding typing bugs in any code that interacts with the authentication logic right now.
* Enables mypy in `test` step/script of the Makefile so it gets run in CI/CD.
* Adds a test suite dedicated to the authentication logic.
* Refactors some of the tricky parts of the authentication logic.

My main goal was to just add type annotations to authz.py, but I stumbled upon a few security gaps and ended up refactoring some of the code:

* Never-expiring JWT tokens (ie. no exp field) are allowed by the backend code. I didn't find an obvious way to create such tokens (besides manually with the secret key) but it seemed like there is no use case for this. This PR makes the exp field required on incoming tokens (as they are used like session tokens).
* The backend logic does not specify which algorithm to use when validating an incoming JWT token, so an attacker could craft a JWT token that uses the `none` algorithm (ie. no secret key). At this point I thought I was going to compromise the app, but the JWT library does throw an exception if the algorithm in the token is `none` while a key was supplied by the backend code; neat. Regardless, this PR explicitly specifies the algorithm to be used for JWT tokens (`H256`).
* An aleph JWT token has a field "a" that is a bool that determines whether the "role" associated with the token is an admin. This setting retrieved from the token takes precedence over the is_admin field in the DB, which, I think, is unexpected/surprising. As far as I know, this means that you can craft a token that can turn any non-admin role into an admin. I haven't changed this behavior but I think this is dangerous and should be removed, so that only the is_admin field in the DB determines whether a role is an admin or not.
  * There is the same behavior with the "b" field in the tokens, which takes precedence over is_blocked, but it doesn't seem as dangerous.
* When parsing an Authorization header from an incoming request, the backend has to "guess" what the data is: an API key or a JWT token. That kind of "ambiguity" can sometimes lead to security bugs and I think it would be useful to require a "hint" in the Authorization header, for example `ApiKey <the api key>` VS `JwtToken <the token>`. In the current behavior there is no strict rule. Adding one would be an API-breaking change so I didn't do it.  
  * It also was possible to pass a JWT token via the `api_key` URL/GET parameter; this PR removes this behavior as it seems like it wasn't expected (based on the documentation at https://github.com/alephdata/aleph/blob/main/aleph/validation/spec.py ).

Some additional comments: 

* This adds mypy, lxml, pydantic as dependencies.
  * If we don't pydantic I can rewrite the code to not use it, but pydantic is a very useful library for parsing arbitrary data and could simplify code in other locations.
  * lxml was already needed to run aleph but it was missing from the requirements file.
* One thing I did (but out of scope for this PR) is to rewrite the Makefile into Python invoke tasks (it looks like this: https://gist.github.com/nabla-c0d3/1606993b9ea4a0b97a3b02eaf05f41ba) so I could run them on Windows (where `make` is not available). If it's something you'd like to see in the project I can make a PR for it, but my guess is that keeping the Makefile is fine.

### DIFF
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -27,10 +27,6 @@ class Authz:
     e.g. in the context of notifications.
     """
 
-    # TODO(AD): Remove after switching all code to ActionEnum
-    READ = "read"
-    WRITE = "write"
-
     PREFIX = "aauthz"
 
     _JWT_TOKENS_VALIDITY_DURATION = timedelta(days=1)
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -31,7 +31,9 @@ class Authz:
     WRITE = "write"
 
     PREFIX = "aauthz"
+
     _JWT_TOKENS_VALIDITY_DURATION = timedelta(days=1)
+    _JWT_TOKENS_ALGORITHM = "HS256"
 
     def __init__(self, role_id: Optional[int], roles: Set[int], is_admin: bool = False, is_blocked: bool = False):
         self.id = role_id
@@ -121,7 +123,7 @@ class Authz:
             return set()
         return self.roles.difference(Role.public_roles())
 
-    def to_token(self, scope: Optional[str] = None) -> bytes:
+    def to_token(self, scope: Optional[str] = None) -> str:
         if self.id is None:
             raise ValueError("Cannot create a JWT token for a None role_id")
 
@@ -133,7 +135,7 @@ class Authz:
             b=self.is_blocked,
             s=scope,
         )
-        return token.to_bytes()
+        return token.to_str()
 
     @classmethod
     def from_token(cls, jwt_token: str, request_path: str) -> "Authz":
@@ -184,14 +186,19 @@ class _JtwToken:
     @classmethod
     def from_str(cls, token_as_str: str) -> "_JtwToken":
         try:
-            token_as_dict = jwt.decode(jwt=token_as_str, key=settings.SECRET_KEY, verify=True, algorithms=["H256"])
+            token_as_dict = jwt.decode(
+                jwt=token_as_str, key=settings.SECRET_KEY, verify=True, algorithms=[Authz._JWT_TOKENS_ALGORITHM]
+            )
             # TODO(AD): Validate the type of each field at runtime as this is attacker-controlled data, or consider
             # using pydantic?
             token = cls(**token_as_dict)
             return token
 
         except (jwt.InvalidTokenError, TypeError):
+            #raise
             raise InvalidJwtToken()
 
-    def to_bytes(self) -> bytes:
-        return jwt.encode(payload=asdict(self), key=settings.SECRET_KEY, algorithm="H256")
+    def to_str(self) -> str:
+        return jwt.encode(
+            payload=asdict(self), key=settings.SECRET_KEY, algorithm=Authz._JWT_TOKENS_ALGORITHM
+        ).decode("utf-8")
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -184,10 +184,13 @@ class _JtwToken:
     @classmethod
     def from_str(cls, token_as_str: str) -> "_JtwToken":
         try:
-            data = jwt.decode(token_as_str, key=settings.SECRET_KEY, verify=True)
-            return cls(**data)
+            token_as_dict = jwt.decode(token_as_str, key=settings.SECRET_KEY, verify=True)
+            # TODO(AD): Validate the type of each field at runtime as this is attacker-controlled data, or consider
+            # using pydantic?
+            token = cls(**token_as_dict)
+            return token
+
         except (jwt.InvalidTokenError, TypeError):
-            log.exception("LOL")
             raise InvalidJwtToken()
 
     def to_bytes(self) -> bytes:
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -142,7 +142,9 @@ class Authz:
             raise Unauthorized()
 
         # TODO(AD): is_admin and is_blocked are already in the DB so it seems dangerous to take them from a JWT token
-        return cls(role_id=parsed_token.u, roles=set(parsed_token.r), is_admin=parsed_token.a, is_blocked=parsed_token.b,)
+        return cls(
+            role_id=parsed_token.u, roles=set(parsed_token.r), is_admin=parsed_token.a, is_blocked=parsed_token.b,
+        )
 
     def __repr__(self) -> str:
         return "<Authz(%s)>" % self.id
@@ -203,6 +205,6 @@ class _JtwToken(pydantic.BaseModel):
 
     def to_str(self) -> str:
         token_as_dict = self.dict(exclude_none=True)
-        return jwt.encode(
-            payload=token_as_dict, key=settings.SECRET_KEY, algorithm=Authz._JWT_TOKENS_ALGORITHM
-        ).decode("utf-8")
+        return jwt.encode(payload=token_as_dict, key=settings.SECRET_KEY, algorithm=Authz._JWT_TOKENS_ALGORITHM).decode(
+            "utf-8"
+        )
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, asdict
 from enum import Enum
-from typing import List, Set, Dict, Optional
+from typing import List, Set, Dict, Optional, Union
 import pydantic
 
 import jwt
@@ -98,17 +98,17 @@ class Authz:
             return False
         return self.logged_in
 
-    def can_write_role(self, role_id: int) -> bool:
+    def can_write_role(self, role_id: Union[int, str]) -> bool:
         if not self.session_write:
             return False
         if self.is_admin:
             return True
-        return role_id in self.private_roles
+        return int(role_id) in self.private_roles
 
-    def can_read_role(self, role_id: int) -> bool:
+    def can_read_role(self, role_id: Union[int, str]) -> bool:
         if self.is_admin:
             return True
-        return role_id in self.roles
+        return int(role_id) in self.roles
 
     @property
     def role(self) -> Role:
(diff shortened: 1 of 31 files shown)


aleph/authz.py
```diff
@@ -82,69 +90,63 @@ def can(self, collection, action):
...
-            return False
+        return role_id in self.private_roles
 
-    def can_read_role(self, role_id):
+    def can_read_role(self, role_id: int) -> bool:
         if self.is_admin:
             return True
-        return int(role_id) in self.roles
-
-    def match(self, roles):
```
nabla-c0d3, line 108: This wasn't used anywhere

aleph/authz.py
```diff
@@ -156,24 +158,49 @@ def from_role(cls, role):
...
+    a: bool = False  # is_admin
+    b: bool = False  # is_blocked
+    s: Optional[str] = None  # Scope ie. request path the token is valid for; None means all paths are authorized
+
+    @classmethod
+    def from_str(cls, token_as_str: str) -> "_JtwToken":
+        # First validate the token's signature and extract its payload
+        try:
+            token_as_dict = jwt.decode(
+                jwt=token_as_str, key=settings.SECRET_KEY, verify=True, algorithms=[Authz._JWT_TOKENS_ALGORITHM]
```
nabla-c0d3, line 187: The algorithm is now configured, when parsing incoming tokens.

aleph/search/__init__.py
```diff
@@ -37,7 +39,7 @@ class EntitiesQuery(Query):
     PREFIX_FIELD = "names.text"
     SKIP_FILTERS = ["schema", "schemata"]
     INCLUDE_FIELDS = PROXY_INCLUDES
-    SORT_DEFAULT = []
+    SORT_DEFAULT: ClassVar[List[Union[str, Dict[str, str]]]] = []
```
nabla-c0d3, line 42: Needed for mpyp to succeed.

aleph/tests/test_sessions_api.py
```diff
@@ -59,4 +59,4 @@ def test_password_login_post_good_email_and_password(self):
         res = self.client.post("/api/2/sessions/login", data=data)
         assert res.status_code == 200, res
         data = jwt.decode(res.json["token"], verify=False)
-        assert data["role"]["id"] == str(self.role.id), res
```
nabla-c0d3, line 62: It looked like having all the "role" fields in the token (as opposed to just the role ID) was never used anywhere so I removed it.

aleph/authz.py
```diff
@@ -156,24 +158,49 @@ def from_role(cls, role):
...
         cache.kv.hdel(cls.PREFIX, *keys)
+
+
+class InvalidJwtToken(Exception):
+    pass
+
+
+class _JtwToken(pydantic.BaseModel):
+    u: int  # role ID
+    r: List[int]  # role IDs  # TODO(AD): Is this required or optional?
```
nabla-c0d3, line 176: I wasn't sure what this field was supposed to contain or what it meant but I haven't changed the behavior.

aleph/authz.py
```diff
@@ -82,69 +90,63 @@ def can(self, collection, action):
...
-        payload = {
-            "u": self.id,
-            "exp": exp,
-            "r": list(self.roles),
-            "a": self.is_admin,
-            "b": self.is_blocked,
-        }
-        if scope is not None:
-            payload["s"] = scope
-        if role is not None:
```
nabla-c0d3, line 136: Having the full role in the token didn't seem to be used anywhere so I removed that.

aleph/authz.py
```diff
@@ -156,24 +158,49 @@ def from_role(cls, role):
...
+class _JtwToken(pydantic.BaseModel):
+    u: int  # role ID
+    r: List[int]  # role IDs  # TODO(AD): Is this required or optional?
+    exp: datetime  # Expiration date; automatically checked by the JWT library but ONLY if present here
+    a: bool = False  # is_admin
+    b: bool = False  # is_blocked
+    s: Optional[str] = None  # Scope ie. request path the token is valid for; None means all paths are authorized
+
+    @classmethod
+    def from_str(cls, token_as_str: str) -> "_JtwToken":
```
nabla-c0d3, line 183: The pydantic library automatically validates all the fields and their types so it's pretty cool. The token-validation logic needs to be bulletproof so that's why I ended up adding this (battle-tested) dependency.

pudo:
hey @nabla-c0d3, thanks for submitting this PR. As I mentioned on Slack, I'm generally against adding typing to Aleph at this stage, given the state of the surrounding Python tooling. I simply don't believe the benefits of the type analysis provided outweigh the amount of cruft it leaves in the code base. If we worked together to merge this PR, would you be up for contributing to anything other than typing in the future?

Some more specific comments:

* I think `_JtwToken` is a typo? Why does it start with an underscore? 
* The `role` object embedded in the JWT token is used extensively by the UI. Did you try and use the UI after removing it? 
* Strongly disagree with overriding the default algorithm used by `pyjwt` and hard-coding it to HS256. We should rely on the library authors to change that as they need to. 
* Not loading the role with each request from the DB was a design decision. I'd like to keep it that way, even if that leaves some tokens valid longer than represented in the DB. 
* Doing explicit `int()` casts on `authz.can_*` makes the code base more brittle. These methods should accept `Any`, it's handled in the actual code. 
* Scoped JWT tokens are passed into `?api_key=` in a number of places, including everywhere a user will download stuff in the browser. Are you sure the following still work: `/archive` get, Xref XSLX export, search result export?

nabla-c0d3:
> hey @nabla-c0d3, thanks for submitting this PR. As I mentioned on Slack, I'm generally against adding typing to Aleph at this stage, given the state of the surrounding Python tooling. I simply don't believe the benefits of the type analysis provided outweigh the amount of cruft it leaves in the code base. If we worked together to merge this PR, would you be up for contributing to anything other than typing in the future?

Type annotations don't have to be an "all or nothing" thing - mypy as currently configured already passes in this PR for the whole code base. So this means that it doesn't require you or anyone to write code with type annotations - mypy can be ignored (unless it finds a typing bug/issue, which is a good thing). No commitment required.

However, while I don't think anyone should be forced to use type annotations, I am personally not willing to write Python code without them. It is too difficult to start understanding and contributing to a semi-large code base when it doesn't have type info (especially in Python). I'm willing to do some of the work to gradually add them as I contribute to the code, so that me and everyone else can get into the code base quicker.

But again, it doesn't have to be a "commitment" nor something that's required on every single line of code.

>     * I think `_JtwToken` is a typo? Why does it start with an underscore?

It's just to signal the fact that it's not meant to be used outside of the authz.py file (ie. it's "private" and the methods on Authz() should be used instead for an "external" caller).

>    * The `role` object embedded in the JWT token is used extensively by the UI. Did you try and use the UI after removing it?

I haven't been able to run the UI in my environment yet so your input is crucial here. I will take a look at the frontend code to see which fields exactly are being used,

>     * Strongly disagree with overriding the default algorithm used by `pyjwt` and hard-coding it to HS256. We should rely on the library authors to change that as they need to.

It's not about overriding the default algorithm of the library (their default choice is fine), it's about not letting the (not-yet-validated ie. untrusted) token choose which algorithm the server should use. This is a security problem with JWT tokens:

* "The server should already know what algorithm it uses to sign tokens, and it's not safe to allow attackers to provide this value." from https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/ 
* "Last, during token validation, explicitly request that the expected algorithm was used." https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html

>     * Not loading the role with each request from the DB was a design decision. I'd like to keep it that way, even if that leaves some tokens valid longer than represented in the DB.

I think it's a bit dangerous but I'm fine with leaving this as it is. However, what is the expected behavior if the token and the DB disagree on whether a role is an admin? What should the code do?

>     * Doing explicit `int()` casts on `authz.can_*` makes the code base more brittle. These methods should accept `Any`, it's handled in the actual code.

Ok I can switch it back and the method would take an `Union[int, str]` (what `int()` can take).

>     * Scoped JWT tokens are passed into `?api_key=` in a number of places, including everywhere a user will download stuff in the browser. Are you sure the following still work: `/archive` get, Xref XSLX export, search result export?

Ok I can add this back.

>     * I'd also prefer to keep mypy out of the CI process. Making the aleph code pass mypy is a commitment I refuse to make.

With the current mypy configuration, the aleph code is already passing mypy :). It does not require type annotations - just if they're present they will be checked and the code using them.

nabla-c0d3:
I made the changes/reverts I mentioned in my previous reply.

pudo:
> However, while I don't think anyone should be forced to use type annotations, I am personally not willing to write Python code without them (and, when present, they need to be checked in CI to be useful). It is too difficult to start understanding and contributing to a semi-large code base when it doesn't have type info (especially in Python).

I don't see a potential for collaboration on that basis, I'm sorry.

nabla-c0d3:
That's too bad - there's plenty of code bases with a mix of typed and untyped Python code, and again it doesn't have to be all or nothing. This blog post does a good job at explaining the benefits https://medium.com/@yothinix/python-type-annotation-and-why-we-need-it-91820a708170 :

* An easier code to maintain, because we explicitly state the type of the arguments of function which help other developers include ourselves understand the function easily
* An easier code review, because the reviewer doesn’t need to guess a type of function they looking at
* Easy to debug code, same reason as above because we don’t need to guess a type of function we’re working with
* Validate our assumption, When we writing a complex call to multiple functions until we lost track of what we send/receiving. Using type annotation along with type checker help us proof a flow of functions call we’re working with.
* Reduce cognitive load, Instead of remembering what we need to sent/receiving we just write a type and free our brain to think on the other stuff

These points are especially valid when trying to bring more contributors to a very important open-source project, as a "one-man army" approach does not scale, and is probably not fun for you either.

pudo:
The argument in favour and against typing in Python, given the current state of its ecosystem, is pretty complex on both sides. We may well adopt it at some point. However, running aleph as an open source project for the past five years has taught me at least one thing: first-time contributors that try to dictate development policies are not how we're going to make a good and stable product.You are of course free to maintain a fork.

Issue #1334, opened 2020-08-27
Map entities into lists


pudo:
When running a mapping via the UI (or mapping API), I want to be able to specify an entity set (list) that the generated data should be made part of. This should remain optional.

pudo:
Fixed in develop.

PR #1336, opened 2020-08-28
Support lists in more APIs

pudo:
Refs #1334.


pudo:
@kjacks I think we need to refactor the `EntitySetSelector` to allow custom callbacks to we can add it to the `MappingEditor` to let people pick a "target list"... Can you look into that?

pudo:
Need to check:

* Serialisation of entityset_id into the mapping
* Deleting entities from the entityset when unloading the mapping (how?)

kjacks:
@pudo from my end this one seems ready to go :)

PR #1348, opened 2020-09-04
WIP: Notifications list event filter

### LINKED ISSUE #1286: Add type filter to user notifications
- add notification types to app metadata
- add notification type select facet to user notification screen

kjacks:
Fixes #1286 

@pudo I hard-coded the notification event types, but presumably we would want to include them in the metadata instead. Also, `filter:event` doesn't seem to filter as expected - I'm assuming I'm setting the parameter incorrectly, so just let me know :)


kjacks:
Replaced by #1353

PR #1352, opened 2020-09-08
Upload: Larger Dropzone

felixebert:
upload: dropzone now covers the whole screen, only excluding navbar and breadcrumb

changes `<main>` css display property from `block` to `flex`!

before:
Screenshot 2020-09-08 at 12 09 48

after:
Screenshot 2020-09-08 at 12 09 14


pudo:
Thanks for coding this up, Felix!

PR #1362, opened 2020-09-15
[WIP] Use followthemoney-predict for XRef

mynameisfiber:
This pull request uses `followthemoney-predict` to do xref classification.

Todo:
- [x] Clean up the actual model creation and model selection (ie: who creates a model, when, and how does the correct model get selected by aleph)
- [x] Find final resting place for production model
  - https://public.data.occrp.org/develop/models/xref/
- [x] Add some model information into the XRef index
  - Add model version metadata + inference extra data

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM ubuntu:20.04
 ENV DEBIAN_FRONTEND noninteractive
+ARG FTM_PREDICT_MODEL=https://public.data.occrp.org/develop/models/xref/model.xgboost.58c81d5.pkl
 
 # build-essential 
 RUN apt-get -qq -y update \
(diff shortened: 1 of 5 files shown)


pudo:
Good stuff! I wonder if we shouldn't ship the model with the Docker image rather than making the prod system aware of the artifact bucket? We could achieve this by turning the relevant files to public inside the bucket and then adding a line like this to the Dockerfile:

```docker
ADD http://storage.googleapis.com/occrp-data-dev/....bla.pkl /data
```

WDYT?

mynameisfiber:
@pudo That sounds good to me! The one downside is that changing the model will require a full aleph rebuild. That being said, the downside of the current mechanism is every aleph restart forces a fetch from GCS.

pudo:
Closing as there's a whole new round of work on this happening.

PR #1402, opened 2020-10-16
Mapping editor - Support custom labels

kjacks:
- Allow user to input custom labels for mapped schema types
- By default, use schema name, not schema label as key, but still display label (should solve issue with cyrillic labels being used as keys)


pudo:
@dependabot squash and merge

PR #1410, opened 2020-10-19
Fixed oauth's handle_azure_oauth to use updated jwt.decode

mudsill:


pudo:
Ok, I really like this. Should I merge now or do you want to see if we need to put some flask state into redis to resolve the original issue?

mudsill:
Probably good to wait to bundle 'em, yeah.

mudsill:
Should I leave this up for the next release?

PR #1415, opened 2020-10-23
Transliteration for names and addresses + Phone number validation

### LINKED ISSUE #1371: Show transliterated names in UI
We want to be able to show latin transliterations of names and addresses in other languages in the user interface next to the original alphabet. 

When serving an entity, the API should include the transliterated string fragments.

kjacks:
Resolves #1371 

Based on #1411


pudo:
Wow, this is cool - but I think it could use a little bit of further work. 

* Are the yellow pop-ups something custom-designed? Wouldn't it be nicer to use the Blueprint tooltips and stay in-framework? Is the Copy-C thing cross-platform?
* For some reason, entity links in the EntityReferencesMode aren't highlighted. Is this somewhere we're using non-standard components?

kjacks:
@pudo 

- Ooh I'll take a look at the EntityReferencesMode issue! It's using the same standard Entity/Property components, but maybe there is some other eccentricity happening there :) 
- I had it with the default blueprint tooltips originally and they just seemed way too intrusive for something that appears so frequently on a page. Also, we're using the same default tooltips for so many different purposes/contexts already that do very different things. I went with the yellow because it seemed to be fairly common for in-text annotations when I scanned the web (it's the exact same yellow goggle uses for in-text hover translations...). Happy to talk about other colors and such :) but I really think we need to depart from blueprint's limited default styles sometimes
- The Copy-C is attached to the "copy" event, so should be fairly browser-unspecific. It displays a ⌘ on mac and Ctl on other os

PR #1430, opened 2020-10-28
GET /entities/<entity_id>/entitysets Endpoint

mynameisfiber:
This adds a new endpoint attached to entities in order to find all EntitySets which reference the entity. The entitysets can be filtered based on type, collection_id, label or the judgement on the entity within the set.

### DIFF
--- a/aleph/views/entities_api.py
+++ b/aleph/views/entities_api.py
@@ -13,9 +13,9 @@ from aleph.search.parser import SearchQueryParser, QueryParser
 from aleph.logic.entities import upsert_entity, delete_entity
 from aleph.logic.entities import entity_references, entity_tags, entity_expand
 from aleph.logic.entities import validate_entity, check_write_entity
-from aleph.logic.entitysets import get_entitysets_by_entity
 from aleph.logic.html import sanitize_html
 from aleph.logic.export import create_export
+from aleph.model.entityset import EntitySet, Judgement
 from aleph.index.util import MAX_PAGE
 from aleph.views.util import get_index_entity, get_db_collection
 from aleph.views.util import jsonify, parse_request, get_flag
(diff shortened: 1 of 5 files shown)


aleph/views/entities_api.py
```diff
@@ -576,3 +577,83 @@ def expand(entity_id):
...
+      - Profiles
+    """
+    entity = get_index_entity(entity_id, request.authz.READ)
+
+    parser = QueryParser(request.args, request.authz)
+    collection_ids = [
+        cid
+        for cid in parser.filters.get("collection_id", [])
+        if request.authz.can(cid, request.authz.READ)
+    ]
```
pudo, line 649: If I read this correctly, then having no `filter:collection_id` filters will return an empty array, and in `.by_entity_id` that will disable the authorisation filter.

aleph/views/entities_api.py
```diff
@@ -576,3 +577,83 @@ def expand(entity_id):
...
+    judgements = parser.filters.get("judgement")
+    labels = parser.filters.get("label")
+    types = parser.filters.get("type")
+
+    entitysets = get_entitysets_by_entity(
```
pudo, line 654: Maybe we can inline the query building to here? It's only 2, 3 lines and we can avoid one round of indirection.

pudo (changes requested):
I think there's a security thing here.

aleph/views/entities_api.py
mynameisfiber: Sure thing!

aleph/views/entities_api.py
mynameisfiber: Correct, but then `EntitySetItemsQuery` will filter out based on the `collection_id`. However, at that point the data is already fetched, so I can also add a:

```
if not collection_ids:
    collection_ids = request.authz.collections(request.authz.READ)
```

mynameisfiber: Actually, hold up. I'm confusing this part of code with what I was looking at for #1429. `EntitySetItemsQuery` has no bearing on this.

pudo: Yeah I think that would be cool. I don't think `EntitySetItemsQuery` is involved?

mynameisfiber: For #1429 I was making sure to myself that `GET /entitysets/XXX/entities` wouldn't leak which lead my down a rabbithole.

PR #1442, opened 2020-11-05
UI build fix

kjacks:
- Resolves production ui build failing in #1418 and #1419 
- latest react-ftm (with fix for network diagram crashing when loading a vertex menu on vertex double click)


pudo:
@dependabot squash and merge

PR #1450, opened 2020-11-11
Ensure auth is forced when trying to access session-required screens

### LINKED ISSUE #1448: HotkeysTarget drops props
They hotkey hack in `Screen.jsx` doesn't seem to retain the props for the component. This means all pages on Aleph have no title, and the screen can't show a login screen when the current user is unauthenticated and trying to access an access controlled screen. I don't think this is a great trade-off. Perhaps we could fail towards removing the hotkeys for now, if this is not an easy fix?

kjacks:
Resolves #1448  

Issue wasn't with hotkeys in the end, but with the requireSession param not being passed through the loading screen


pudo:
Ups my bad!

PR #1452, opened 2020-11-11
Remove `entities` from entity sets API

pudo:


kjacks:
Known Issues: 
- Lists: clicking the "Add link" button while viewing a list allows a user to select target entities from not only those within the list, but also entities from the rest of the collection (this could be desirable but if so then we should ideally automatically add those target entities to the list as well)
- Diagrams: because expanded entity results are now added to the diagram in separate parallel PUT requests, expansions with large quantities of entities can be interrupted by page refresh etc. and result in only a partial expansion

pudo:
these seem acceptable to me, but very good to note down as especially the list one may cause a bit of confusion down the way.

kjacks:
Exactly - like not breaking, but would be nice to fix :) And the lists one is definitely fix-able with a little bit of finnagling

PR #1472, opened 2020-11-25
Log more contextual info with the es query

sunu:


pudo:
Should we discard this in favor of #1498?

sunu:
Yes, we should.

PR #1487, opened 2020-11-29
Profiles UI

pudo:
This is going to be incredibly boring at first, but trying to make profiles show like fat-ish entities.


kjacks:
It's looking really amazing - especially as a first-go at a super complicated and potentially confusing feature. Here are my jumbled thoughts (I'm happy to push some changes to implement any or none of them if it's helpful):

**STRUCTURAL THOUGHTS**

- Instead of upgrading a person entity to a profile, would it be possible to show the profile as a a third item. I.e. a person profile that is a bag that contains 2 constituent person entities?

- For me, when viewing a profile there are two main thematic sections:
    - The entity judgements + any suggestions
    - The aggregated properties and relationships of constituent entities
Structurally, the “Profile” and “Suggested” tabs when viewing the profile are very different than “Memberships”, “Directorships”, other references, etc., but currently they are comingled 

**CRITICALISH-FEELING-LEVEL ISSUES**

- When viewing a profile, deleting a judgement on the “original” golden entity deletes the profile -> leads to 404

**KINDOFCONFUSING-LEVEL ISSUES**

- When I make a judgement, it’s not bi-directional (i.e. I judge Kirk1 to be the same as Kirk2, I also need to also judge Kirk2 to be the same as Kirk1)
- The search preview and the expanded view for a profile are different -> should we get rid of the similar screen in the preview for profiles?
- The way we show profiles could be a bit more consistent throughout the site
    - In search results, search preview, breadcrumbs, and the entity/profile view it’s currently a different combo of icon/golden-ness/profile tag display

**MINOR ISSUES**
- Make a judgement -> then click the preview for the entity I just judged -> suddenly it turns gold, not after I make the judgement

**DESIGNY THOUGHTS**
- More gold! :)

PR #1498, opened 2020-12-02
Structured logging using structlog

### LINKED ISSUE #1479: Structured logging for Aleph on Production
Downsides of Aleph's unstructured text logs: 
- log entries in Stackdriver show up as text blobs with incorrect level indicators in production
- we can't analyze these logs easily - either with filters provided by Stackdriver or with something like Google's Datastudio tools.

Downside of JSON logs:
- Hard to read in local dev environment

Ideally, we want to
- format our logs as JSON that Stackdriver can parse correctly in the production environment and include context info like request_id, service name, hostname etc
- format them as human readable text on local dev environment

Logging tools to consider:
- [python-json-logger](https://pypi.org/project/python-json-logger/)
- [structlog](https://www.structlog.org/en/stable/)

sunu:
Fixes #1479 

Some notes:
- The standard library logging module lets us pass data into logging statements via `extra` argument. While `structlog` is pretty nice overall, it doesn't support taking that `extra` argument and piping it out into its own nicely formatted output. There's an open issue about it [hynek/structlog/issues/209].
This means we can use the standard logging interface for text only logs but when we want to pass structured data, we need to use the `structlog` interface.
- I've removed the Google logging client we were using to log requests separately for analytics and replaced it with `structlog` logging. We'll have to change the source in Data Studio so that analytics reporting doesn't break.


aleph/views/context.py
```diff
@@ -168,41 +172,42 @@ def finalize_response(resp):
     return resp
 
 
+def setup_logging_context(request):
+    role_id = None
+    if hasattr(request, "authz"):
+        role_id = request.authz.id
+    # Set up context varibales for structured logging
+    clear_contextvars()
+    bind_contextvars(
```
pudo, line 181: Does this mean that if a request doesn't result in a log message, it will not end up in BigQuery? I also don't think we need all that data in each log message necessarily (although it probably doesn't hurt).

aleph/views/context.py
```diff
@@ -168,41 +172,42 @@ def finalize_response(resp):
...
-    # log.info("Log: %s", pformat(payload))
-    if not hasattr(local, "_gcp_logger"):
-        from google.cloud.logging import Client
-
-        client = Client()
-        logger_name = "%s-api" % settings.APP_NAME
-        local._gcp_logger = client.logger(logger_name)
-    local._gcp_logger.log_struct(payload)
+    bind_contextvars(**payload)
+    log.info("Request handled", **payload)
```
pudo, line 213: Ah here's my answer.

pudo (commented):
LGTM

PR #1506, opened 2020-12-07
Scoped Search Removal / Faceted Collection Search Mode

kjacks:
- Removes scoped search from throughout the site

- New searchbars:
1. Collection Screen breadcrumbs
2. Entity Screen breadcumbs (where searchable)

- Faceted entity search mode now in collections


kjacks:
@pudo This now has the following changes as discussed: 

- [x] Search in “This investigation” instead of collection title
- [x] Move folder search back to doc manager
- [x] Move search bar from entity breadcrumbs to it’s own action bar
- [x] Entity action bar should always display search on the right even when no button content
- [x] New search heading component with: result text, “Date distribution”, and export button
- [x] Remove crumbs from search screen

Would love a second set of eyes to test, but otherwise should be set to merge

pudo:
Checking it out now :)

pudo:
This is great! If I could make one wish, it would be to generally switch the search placeholders from the title of the "thing" to the subject of the search, e.g. "Search in investigation...", "Search in folder...", "Search inside document...". What do you think?

kjacks:
@pudo I think I would actually tend towards preferring the other direction, and being as specific as possible about where you are searching. I'm just thinking that using the "in investigation", "in dataset", "in document" relies on people knowing and internalizing the terminology we've adopted - which we've struggled to standardize ourselves. As opposed to the logic of: Ok I'm in the UK companies registry, and now I want to search the registry - where I don't need to know if it's technically a dataset or whatever we happen to call it.

 Especially in the case of something like "Search in folder", this could be super ambiguous when navigating layers of heirarchy.

pudo:
Conceded.

PR #1557, opened 2021-01-14
Move helm charts into the repo and publish helm charts in a helm repository

sunu:
Fixes https://github.com/alephdata/aleph-helm-charts/issues/4, https://github.com/alephdata/aleph-helm-charts/issues/5 and https://github.com/occrp/data.occrp.org/issues/85

ToDo:

- [x] Adjust the CI and release flow to work alongside Aleph's CI
- [x] Set up a helm repository to publish charts

Charts are built from `develop` branch and tagged releases, pushed into a GCS bucket and published to the helm repo at https://aleph-helm-charts.storage.googleapis.com/


sunu:
Releasing charts with github releases doesn't work because each release needs to be based on a tag. And we don't want to create a tag for every commit.

Right now, a helm repo is set up at https://aleph-helm-charts.storage.googleapis.com and charts are uploaded to it for each commit (aleph-3.9.9+commithash.tgz) and for each tag (aleph-3.9.9.tgz). But `helm repo search` returns a single result only. Need to figure out what's going wrong there.

sunu:
Leaving out the `+commithash` seems to make helm less confused.

pudo:
Minor aesthetic question: why `contrib/helm/charts/aleph` instead of `helm`? Is some of the depth required by the chart layout?

PR #1576, opened 2021-01-20
UI: PDF Viewer fixes

### LINKED ISSUE #1570: PDF infinite loading when re-fetching
Steps to reproduce:
- click a document-type search result -> the pdf is correctly loaded in entity preview pane
- click "Expand" -> when loading the entity screen, pdfViewer hangs and never re-fetches the pdf

I'm a bit miffed by this one. It only happens when attempting to load a pdf that has already been previously fetched before, and seems to have something to do with pdf-js's internal diffing mechanism for the file parameter to its Document component, which prevents the request for the necessary pdf from ever being sent.

From react-pdf [docs](https://github.com/wojtekmaj/react-pdf): "Warning: Since equality check (===) is used to determine if file object has changed, it must be memoized by setting it in component's state, useMemo or other similar technique."

I was able to fix this issue by adding a short sleep() in between steps 3 and 4 above, but this is less than ideal. It also seemed to be resolved by importing react-pdf components using standard syntax instead of dynamically on mount, but need to be sure this wouldn't jeopardize the worker in prod.

kjacks:
Fixes #1570 
I spent probably way too long trying to figure out a way to do this without the CDN, but it really seems that's the only way to ensure the pdf worker is loaded correctly - and is the officially recommended approach when using an un-ejected create-react-app. But PDFs are now loading so much more smoothly and it seems quickly as well.

This PR also fixes an issue where the entire app would flash into a brief loading state when the PDFViewer was lazy loaded for the first time.

### DIFF
--- a/ui/src/components/Document/DocumentViewMode.jsx
+++ b/ui/src/components/Document/DocumentViewMode.jsx
@@ -1,4 +1,4 @@
-import React, { lazy } from 'react';
+import React, { lazy, Suspense } from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


pudo:
This is going to break Aleph for installations where the machines don't have access to the public internet (eg. Süddeutsche). I kinda wonder if we could find a work-around for this where the specified worker URL is still pointing at the aleph ui container, maybe by modifying it's build process?

kjacks:
I knew there was at least one reason the cdn wouldn't work... 

Ok how about this? It's now directly importing the worker script from the pdfjs build in node_modules

PR #1626, opened 2021-02-11
Get rid of sample-entities command. Provide option to filter by secret and casefile flags when listing collections.

sunu:
Doing the sampling inside ES led to hanging queries. So we do the sampling outside ES.

The other alternative is to do the sampling inside ftm-store database but we have to additionally implement property based filters in SQL in that case.

### DIFF
--- a/aleph/index/entities.py
+++ b/aleph/index/entities.py
@@ -1,6 +1,5 @@
 import logging
 import fingerprints
-import warnings
 from pprint import pprint, pformat  # noqa
 from banal import ensure_list, first
 from followthemoney import model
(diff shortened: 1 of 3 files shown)


pudo:
Hey if we're getting rid of the `randomize=True` flag in `iter_`, can we also delete it upstream? Otherwise it just sticks around as cruft.

sunu:
Oh I should mark this PR as WIP. Let's not merge this yet. I'm still trying to figure out if it holds good for the bigger stream or not. Will clean it up after that.

sunu:
@mynameisfiber Should we get rid of `sample-entities` command now that we sample from ftm-store or is it still useful to keep?

mynameisfiber:
@sunu delete it!

pudo:
@mynameisfiber please can you handle?

mynameisfiber:
@pudo yup

mynameisfiber:
@sunu @pudo updated

aleph/index/entities.py
```diff
@@ -66,19 +64,31 @@ def iter_entities(
     includes=PROXY_INCLUDES,
     excludes=None,
     filters=None,
-    should=None,
     sort=None,
-    raise_on_error=True,
+    randomize=False,
+    random_seed=None,
```
sunu: @mynameisfiber May be we should get rid of these arguments too since they are not used anywhere anymore?

mynameisfiber:
@sunu good catch!

mynameisfiber:
anyone mind if i squash all these commits and resolve the merge conflict?

sunu:
@mynameisfiber no, please go ahead!

mynameisfiber:
@sunu all cleaned up

pudo:
Lovely! Thanks, guys!

Issue #1516, opened 2020-05-13
Implement 'last seen by you' in localStorage


pudo:
"As a reporter going through I leak, I want to know which documents I've already visited, and which ones are yet to be explored."

Implement a function in the UI that stores the last time a visitor has viewed an entity, either in full screen or preview mode. Visited entity have a little eye behind their links, with a tooltip that shows the last visit time ('3 mins ago', 'May 3, 2020').

This should not require the user to be logged in (it's just a redux state). Maybe visits older than two months should be deleted.

mattcg:
So annoying that we have to reinvent the browser history engine. Wish browsers would expose this data for the current origin.

kjacks:
Solved in #1628

PR #1628, opened 2021-02-14
WIP #1516 Implement 'last seen by you' in localStorage

Rosencrantz:
This is a first pass at trying to add in 'last seen' functionality into Aleph. A couple of notes on the implementation that may help:

- Originally I had simple name value pairs in localStorage which would not require parsing. However, in order to be able to filter out anything older than 2 months I had to put all recently viewed items into a single name/value pair.
- Rather than implementing an icon (an eye) that would identify a entity as having been viewed I went with a simple change in the opacity of the row. This felt like a more natural UI/UX and something akin to read vs unread in email clients like protonmail or gmail.
-  To provide the nomenclature for the dates I've used date-fns. It's a popular library, similar to the now defunct moment.js, it's updated regularly and doesn't seem to have any security issues associated with it right now.
- I appreciate that the organisation of functions may not be perfect, would appreciate guidance on the best place for everything.

In addition to this change I believe that in order to have dates put into entity titles there may need to be an update to react-ftm. I still need to wrap my head around the setCaption/getCaption functionality as it doesn't overwrite the caption info as I'd expect it to. It might be easier to simply to update Entity.tsx to pull the recentlyViewed data directly from localStorage.


kjacks:
Hi @Rosencrantz - thank you for digging into this! :) 

Some thoughts:
- I agree that changing opacity is nice non-intrusive ui convention to indicate what’s been seen, and I really like the way you've got the entity selector handling the merging of an entity with its last seen date - looks great

- There's an existing localized date component that could potentially be useful instead of the custom getFormattedDate util - it's `FormattedRelativeTime` from react-intl. which you can see in action in the `Notification.jsx` component. This way we wouldn't have to add the date-fns dependency and wouldn't have to set any locale data on the `window` global

- I'm going back and forth on the benefits of calling the `expireRecentlyViewed` function on initialization versus checking the expiration only when a specific entity is requested (a pattern like this? https://www.sohamkamani.com/blog/javascript-localstorage-with-ttl-expiry/) But would be curious to hear your thoughts if you think the bulk expiration approach is better

Rosencrantz:
@kjacks Thanks for the feedback, it's always good to get insights from people experienced in a particular codebase. 

- I will take a look at the FormattedRelativeTime component, if that can be used instead of date-fns then I'll swap it out. getFormattedDate was a wrapper for the i18n of date-fns so I didn't need to import all the i18n files every time I used the library.

- expireRecentlyViewed wasn't my first choice either. I even read the article you shared as I was considering the implementation (great minds and all that). The problem I found with that particular approach was that you'd never remove a record because localStorage is checked and then updated only at the point at which an entity is viewed. What you really need is something that will get rid of all the entities that you're never going to look at again, and the only way I could come up with for that was to clean out all the old records on initialization.

ui/src/components/EntitySearch/EntitySearchResultsRow.jsx
```diff
@@ -75,9 +75,11 @@ class EntitySearchResultsRow extends Component {
     const isPrefix = !!highlights.length;
     const resultClass = c('EntitySearchResultsRow', 'nowrap', { active: isActive }, { prefix: isPrefix });
     const highlightsClass = c('EntitySearchResultsRow', { active: isActive });
+    const viewed = entity.lastViewed ? 'viewed' : '';
+
     return (
       <>
-        <tr key={entity.id} className={resultClass}>
+        <tr key={entity.id} className={`${resultClass} ${viewed}`}>
```
kjacks, line 82: Here you could maybe use the classnames library for class name composition, something like this?:
`className={c(resultClass, { viewed: !!entity.lastViewed })} `

ui/src/screens/EntityScreen/EntityScreen.jsx
```diff
@@ -53,6 +54,8 @@ class EntityScreen extends Component {
       );
     }
 
+    setRecentlyViewedItem(entityId);
```
kjacks, line 57: This maybe wants to live in componentDidMount? so that it's not being called on every render?

ui/src/selectors.js
```diff
@@ -51,13 +52,18 @@ export function selectLocale(state) {
...
+  window.__localId__ = 'en';
+  
   if (role.locale) {
+    window.__localId__ = role.locale;
     return role.locale;
   }
   const { metadata } = state;
   if (metadata && metadata.app) {
+    window.__localId__ = metadata.app.locale;
     return metadata.app.locale;
```
kjacks, line 64: I think now with the FormattedRelativeTime component, we no longer need to set anything on the `window` ?

ui/src/components/Entity/EntityHeading.jsx
```diff
@@ -9,6 +10,7 @@ import 'components/common/ItemOverview.scss';
 class EntityHeading extends React.PureComponent {
   render() {
     const { entity, isProfile = false } = this.props;
+    const { value, unit } = selectUnit(new Date(parseInt(entity.lastViewed, 10)), Date.now());
```
kjacks, line 13: What happens if entity.lastViewed is undefined? maybe add a conditional to check before parsing

kjacks (commented):
@Rosencrantz looking great - the change to `FormattedRelativeTime` is perfect.

- I see what you mean about the value of having some kind of passive expiration like `expireRecentlyViewed`. I'm truthfully not so familiar with the detriments of not doing this kind of cleanup on a user's localStorage - I guess it's a good practice, so why not :)

- I'd be curious to hear your thoughts also on storing the last viewed data as an array versus as an object/map of key value pairs - I guess my only thought is that a key-value store would eliminate having to iterate through a potentially large array to check each individual entity id on render?

- re: getCaption / setCaption - I really like how you have it currently set up, where the last viewed data is populated by the selector and not stored in any way in the Entity properties itself

Thank you so much once again for taking the time to work through this! :)

ui/src/components/Entity/EntityHeading.jsx
Rosencrantz: Shoot, I had that previously, but removed it when refactoring, nice catch. I'll update this accordingly

ui/src/selectors.js
Rosencrantz: Very true. I'll go through and remove this

Rosencrantz:
@kjacks Hah. I did in fact have the initial implementation use a simple name/value pair directly in localStorage for exactly the reasons you specify. IIRC, the name was the entityId and the value was the date. The problem was, once I accepted the need to iterate over all dates in order to remove the expired ones, I needed to be able to target all items that had been recentlyViewed which lead me onto putting everything into a JSON object. Having each name/value as an independent record in localStorage is a lot more simple but doesn't allow you to target all recentlyViewed items.

ui/src/components/Entity/EntityHeading.jsx
```diff
@@ -9,6 +10,9 @@ import 'components/common/ItemOverview.scss';
 class EntityHeading extends React.PureComponent {
   render() {
     const { entity, isProfile = false } = this.props;
+    const lastViewedDate = entity.lastViewed ? new Date(parseInt(entity.lastViewed, 10)) : Date.now();
```
Rosencrantz, line 13: I'm a bit meh on this, mainly because although it's logical I know that the Date.now() part of this will never actually be used. Stops problems with entity.lastViewed not existing though, which is nice

kjacks:
Thank you so much for your work on this @Rosencrantz - it's all merged into the Aleph develop branch now :)

Rosencrantz:
Thanks @kjacks let me know if there is anything else you’d like looked at!

PR #1633, opened 2021-02-17
Xref Party Mode

mynameisfiber:
This PR adds xref party mode which jumbles up the xref to include a healthy mix of good/bad results. This is going to be useful in creating training data for our xref models.

### DIFF
--- a/aleph/search/__init__.py
+++ b/aleph/search/__init__.py
@@ -103,6 +103,11 @@ class XrefQuery(Query):
         return xref_index()
 
 
+class XrefPartyQuery(XrefQuery):
+    def get_sort(self):
+        return [{"_id": "desc"}]
+
+
 class NotificationsQuery(Query):
     AUTHZ_FIELD = None
     TEXT_FIELDS = ["text"]
--- a/aleph/search/__init__.py
+++ b/aleph/search/__init__.py
@@ -103,7 +103,7 @@ class XrefQuery(Query):
         return xref_index()
 
 
-class XrefPartyQuery(XrefQuery):
+class XrefEvaluationQuery(XrefQuery):
     def get_sort(self):
         return [{"_id": "desc"}]
 
--- a/aleph/views/xref_api.py
+++ b/aleph/views/xref_api.py
@@ -1,8 +1,9 @@
 import logging
+import random
 from flask import Blueprint, request
 from pantomime.types import XLSX
 
-from aleph.search import XrefQuery
+from aleph.search import XrefQuery, XrefPartyQuery
 from aleph.logic.profiles import pairwise_judgements
 from aleph.logic.export import create_export
 from aleph.views.serializers import XrefSerializer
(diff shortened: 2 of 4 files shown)


kjacks:
@mynameisfiber woohoo party 🎈 🎈 🎈

I'm getting this error when I send an xref request with party=true

Screen Shot 2021-02-17 at 18 55 32

aleph/views/xref_api.py
```diff
@@ -53,6 +61,13 @@ def index(collection_id):
     """
     get_index_collection(collection_id, request.authz.READ)
     result = XrefQuery.handle(request, collection_id=collection_id)
+    if get_flag("party", default=False):
```
pudo, line 64: I'm not convinced we really want to call this ``party`` - that's a bit cryptic: party to what? The name makes sense in our internal discourse but it'll age really badly in 18 months.

aleph/search/__init__.py
```diff
@@ -103,6 +103,11 @@ def get_index(self):
         return xref_index()
 
 
+class XrefPartyQuery(XrefQuery):
```
pudo, line 106: The alternative to subclassing here would be to define an option in ``XrefQuery.SORT_FIELDS`` which makes the sort available via the normal API.

aleph/views/xref_api.py
```diff
@@ -53,6 +61,13 @@ def index(collection_id):
...
+        # In party mode we merge in some random xref pairs and shuffle the
+        # collection. This gives us a good mix of positive and negative matches
+        results_party = XrefPartyQuery.handle(request, collection_id=collection_id)
+        results_merged = result.results + results_party.result
```
pudo, line 68: This means we're returning twice as many results as requested in the query. I'm a bit concerned this is going to confuse the frontend, but we can try.

pudo (commented):
One thing I wonder is if we should try to find a way to exclude pairings from this mode that have already been decided. Since ES has limits on how many clauses you can feed it, this would probably be a post-filter, which in turn would leave the limit/offset parameters a mess and I think that'll cause a bit of chaos on the frontend. Can you think of a method?

aleph/views/xref_api.py
mynameisfiber: I'll rename it to the more sensibly boring `evaluation_mode`

aleph/search/__init__.py
mynameisfiber: magic. although then i'd have to modify the request options parser so i can create one XrefQuery with the evaluation mode on and another with it off... right?

mynameisfiber:
> I'm getting this error when I send an xref request with party=true

Fixed! Of course I put the typo in while cleaning up variable names just before pushing :disappointed:

aleph/views/xref_api.py
mynameisfiber: @kjacks thoughts?

kjacks:
@mynameisfiber @pudo - ok the front end toggle for eval mode should be good to go. In the UI it felt a bit more consistent and easy to understand to call it Sort by: Default / Sort by: Random. And it's behind the tester flag.

mynameisfiber:
@pudo any other critical changes you'd like us to make on this PR?

PR #1656, opened 2021-03-01
Support Diagram Embeds in the UI

kjacks:
Moves svg, ftm, and embed export options into a single menu

@pudo I wasn't sure on the wording for the ui explanations on some of this, so feel free to suggest any edits :)

Screen Shot 2021-03-01 at 09 20 44

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@alephdata/followthemoney": "2.3.2",
-    "@alephdata/react-ftm": "2.3.1",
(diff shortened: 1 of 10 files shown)


pudo:
This looks fabulous. Maybe regarding the copy: 

* "Vector graphic" instead of "still image"? 
* "Generate an embed code for the diagram that can be used in an article. The embedded version will not reflect future changes in the diagram - to update, you will need to make a new embed code."
* "download the diagram as a data file that can be used in Aleph Data Desktop or another Aleph installation." ?

PR #1669, opened 2021-03-09
Landing Quick Links

### LINKED ISSUE #1660: Quicklinks on dashboard home
The quick links on the investigation home page have come out really cool. I wonder if we should repeat that element on the dashboard that is shown to the user when they login (the notifications index screen). It could be stuff like:

* Search NN entities
* Browse datasets 
* Create an investigation (to make networks and upload documents)
* Make some alerts (to follow important searches)

kjacks:
Fixes #1660


pudo:
@dependabot squash and merge

PR #1719, opened 2021-04-06
Group entities search

kjacks:
- Link from group screen to `/search` with group's collections pre-selected as filters
- Togglable display cutoff for query tags component to handle cases where many filters are active


pudo:
@dependabot squash and merge

PR #1747, opened 2021-04-23
Add OAuth2 support for Auth0 (audience, permissions)

davidknezic:
These two changes make it easier to use the Auth0 OIDC provider with Aleph.

**Optional `ALEPH_OAUTH_AUDIENCE` env var**

Auth0 requires the `?audience` parameter to be submitted with the `authorize` endpoint, so it can create a valid JWT for that particular API. Without it, it generates an opaque token that crashes the token parser. The optional `ALEPH_OAUTH_AUDIENCE` makes it possible to set that `audience` parameter.

**Read groups from `permissions` claim**

Auth0 uses the `permissions` claim to communicate what the token subject can do (e.g. `superuser`). This PR appends the `permissions` array to `groups` during token processing.


pudo:
Hey @davidknezic thanks for submitting that PR, it's very much appreciated since Auth0 seems like the way of the future. Just to verify: is the audience value the same thing as OAUTH_KEY? That's how I've seen it before and then we could default it to being set to the same value as OAUTH_KEY. Thanks!

davidknezic:
Hi @pudo!

Conceptually, the client id (OAUTH_KEY) and the audience are different and not meant to be the same. This is so there can be multiple clients for one single resource server (API).

```

 ┌───────────┐    ┌─────────────────────┐
 │ client #1 ├─┐  │   resource server   │
 └───────────┘ └─►│                     │
                  │ https://aleph.local │
 ┌───────────┐ ┌─►│     (audience)      │
 │ client #2 ├─┘  │                     │
 └───────────┘    └─────────────────────┘

```

However, I've also often seen the client id in the audience field. Probably because it's very common for the resource server and client to be bundled together.

I still wouldn't recommend to set `ALEPH_OAUTH_AUDIENCE = OAUTH_KEY` by default, because I can imagine that the additional OAuth2 `?audience` param could break some of the existing deployments out there.

pudo:
Ok, I'm sold.

pudo:
Thanks for your PR!

PR #1769, opened 2021-05-03
Hotkeys refactor

kjacks:
Refactor to use new blueprint [HotkeysProvider](https://blueprintjs.com/docs/#core/context/hotkeys-provider) wrapper and [useHotkeys](https://blueprintjs.com/docs/#core/hooks/use-hotkeys) hook.


kjacks:
@pudo I reactivated the `/` to focus the search bar on any screen.

If there are others that come to mind that would be useful, I'd be happy to add them

PR #1770, opened 2021-05-03
Replace Popover and Tooltip with Popover2 and Tooltip2

kjacks:
Replaces Popover and Tooltip components with imports from @blueprints/popover2

https://github.com/palantir/blueprint/wiki/Popover2-migration

A benefit, among other things, is now having access to more RTL-compatible tooltip placement props like "bottom-start" instead of "bottom-left"


pudo:
@dependabot squash and merge

kjacks:
Merged: https://github.com/alephdata/aleph/commit/9a6a3f5ba071848d5e5bf3fb3e8626a604f04c89

Issue #1796, opened 2021-05-14
Aleph should run on Apple Silicon


Rosencrantz:
Building and running Aleph on Apple Silicon currently fails.  The intial error raised during the build process is:

`#10 22.09     Error: pg_config executable not found.
#10 22.09     
#10 22.09     pg_config is required to build psycopg2 from source.  Please add the directory
#10 22.09     containing pg_config to the $PATH or specify the full executable path with the
#10 22.09     option:`

It is likely that this is not the only issue that will need to be resolved in order to have Aleph build and run on Apple Silicon

Rosencrantz:
Marking this one as done for the time being. Aleph does now run on Apple Silicon. There may be more work to do in the future, but for now, things are working

PR #1800, opened 2021-05-17
#1796 Upgrade to Apple Silicon compatible node.js and bump node-sass too

Rosencrantz:
This contains changes from @pudo, but also an update to Node and node-sass. Node 16 introduces native support for Apple Silicon.


pudo:
Very cool! I think it'd be good if @kjacks could give his thoughts on going to node16, and I guess we have to figure out how to do `arm64` builds of the ingestors before we can call this done.

Rosencrantz:
Spoke with @kjacks yesterday, he's going to double check that there is nothing devastating in either the node or node-sass upgrades. But, right now, it looks good to merge.

kjacks:
@Rosencrantz @pudo - Both the node-16 and node-sass updates seem all good to go for me. :) 
I played around with a workaround yesterday to try to get us onto dart-sass instead of node-sass bc it's deprecated, but it does seem like we're going to have to wait for the blueprint folks to: [catch up](https://github.com/palantir/blueprint/issues/4032) - which mayyybe might happen in blueprint v4.

For the bump to node-sass v.6.00, I re-ran the ui-prod build that was failing and it passed both locally and remotely, so I'm actually not sure why it failed the first time. From the release notes it seems like the changes are mostly related to tracking node versions anyways

pudo:
Ok thanks for verifying this all, @kjacks! Let's take this into develop and see how it goes!

Issue #1801, opened 2021-05-18
Create a simple article viewer for aleph


Rosencrantz:
Currently Aleph does not have the capability to render Article entities. We should create a simple article viewer that provides the ability to view these entity types when placed into Aleph.

Rosencrantz:
There is now a simple Article viewer in Aleph.

PR #1802, opened 2021-05-18
#1801 Introduce a basic ArticleViewer template to show the body of Ar…

Rosencrantz:
…ticles


pudo:
Does this want to have a little header section like on emails to show author, publication date etc. in the main preview? Looks great, otherwise.

Rosencrantz:
@pudo Honestly hadn't given it much consideration beyond 'make the content render' but I can take a look and add in those details. Shouldn't be a problem

PR #1803, opened 2021-05-20
Configurable search screen - facets + columns

### LINKED ISSUE #647: Make facets configurable
As a user, I want to be able to add or remove facets on the main search. There are two types of facets: `group` facets (such as `names`, `phones`, `ibans` and `property` facets (such as `BankAccount.bankName`). Both should be selectable from a menu or drop-down.

kjacks:
Fixes #647 

- Allows users to configure custom facets for entity search screens - supports both type group facets (names, file types, addresses, etc.) as well as property facets.
- Selected facets are persisted to localStorage

What's left @pudo (or @Rosencrantz ):

- [ ]  Date type property facets are currently returning non-readable dates from the backend as values
Screen Shot 2021-05-20 at 12 53 38

- [x] As per our discussion, right now all ftm properties with the `matchable` flag are allowed as facets - should this be further restricted?

- [x] When filtering by a property facet, like location or birth date, it feels a bit strange that the search results don't include that property column - so you end up seeing a results list filtered by an invisible property value. We could add any property columns to the results list when that property facet is active?


kjacks:
@Rosencrantz This should be ready to go from my end! 

It should probably wait for #1819 before being merged.

This PR does not include results list formatting changes described by #1844

Rosencrantz:
I've done a quick a quick play on an instance running on my local machine. To be honest it all looks pretty good. A little bit of feedback that I did have:

- When I switch from ASC to DESC there seems to be three seperate states
- Still not sure about the configure filters button being at the bottom instead of the top. Doesn't feel right to me
- Would it make sense to move the asc/desc arrow next to the text (on the left), currently it can be tricky to see which value the sort is associated with

ui/src/components/EntitySearch/EntitySearchResultsRow.jsx
```diff
@@ -84,39 +105,15 @@ class EntitySearchResultsRow extends Component {
...
+          {columns.map(column => {
+            const content = this.renderCellContent(column);
+            return <td key={column.name} className={column.type?.name || column.name}>{content}</td>
+          })}
         </tr>
         {!!highlights.length
           && (
             <tr key={`${entity.id}-hl`} className={highlightsClass}>
-              <td colSpan="5" className="highlights">
+              <td colSpan="100%" className="highlights">
```
Rosencrantz, line 116: I think colspan should be all lowercase. Also, colspan? Is there a reason why we wouldn't just have width: 100%; in the css?

ui/src/components/EntitySearch/FacetedEntitySearch.jsx
```diff
@@ -203,15 +262,30 @@ export class FacetedEntitySearch extends React.Component {
   }
 }
 const mapStateToProps = (state, ownProps) => {
-  const { query, result } = ownProps;
+  const { query } = ownProps;
+  const searchConfig = getSearchConfig();
+  const facets = searchConfig?.facets || defaultFacets.map(getGroupField);
+  const columns = searchConfig?.columns || defaultColumns.map(getGroupField);
```
Rosencrantz, line 268: I've not seen this before, what do these two line denote? is searchConfig?.facets a variable, or a shorthand if statement, or something specific to react that I'm just not familiar with?

Rosencrantz:
One general comment: I'd love to find a way to have smaller changes in reviews. It's something I do a bunch too, but the bigger the change the harder it is to review the impact... Not sure what the solution is here though.

ui/src/components/EntitySearch/EntitySearchResultsRow.jsx
kjacks: In jsx I think it does have to be "colSpan" not "colspan" or else the compiler will throw a warning - I could be wrong about this though.

And in this case I feel colspan is favorable as the cleanest way to make a table cell span the width of the table

ui/src/components/EntitySearch/FacetedEntitySearch.jsx
kjacks: Ahh yes :) fun new ES6 optional chaining syntax - which allows for a more compact way to catch attempts to access props on undefined objects

kjacks:
@Rosencrantz 

- I agree it is a bit confusing that the column sorting has three states instead of two. I guess we could switch it to just ASC and DESC, the only thing is that then there will be no way to get back to an unsorted state
- Re: configure filters button placement - I guess my thoughts here are that it's not a feature I would expect or really want a user to be interacting with so often, so I'm fine with it intentionally being a bit less prominent. But if it feels too hidden away I'd be happy to move it! Just let me know where you feel it would make more sense
- totally agree - I'll play around with the sorting arrows placement today

ui/src/components/EntitySearch/FacetedEntitySearch.jsx
Rosencrantz: Nice. I'm obviously losing touch with the state of ES. I'll have a look into this. Thanks for the heads up.

kjacks: I think it's pretty much impossible to keep track of this crazy ever-changing javascript monster XD

sunu:
@kjacks It's live on staging now

PR #1818, opened 2021-05-25
Add followthemoney-compare model to XREF

mynameisfiber:
This PR puts the glm-bernoulli-2e model from `followthemoney-compare` into use for aleph's XREF system. In order to do so, the Dockerfile now downloads word frequency information and the model data from public.data.occrp.org and it's all loaded in `aleph.logic.xref`. The model is lazy-loaded on xref evaluation and we fall back to `followthemoney.compare` if the model can't be loaded.

Since the new model also has a measure of confidence associated to it, we added a `doubt` field to the xref index. The smaller this doubt is, the more confident we are in the result. This doubt parameter can be sorted on when requesting XREF from the API. (@kjacks afaik all we need to change the evaluation mode from random to this doubt is to change the requsted sort parameter to `&sort=doubt:desc`).

In order to maintain some sort of insight as to what is calculated by who, we also add a `method` field into the xref index which contains version information about the model which calculated that particular xref.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ENV DEBIAN_FRONTEND noninteractive
 # build-essential 
 RUN apt-get -qq -y update \
     && apt-get -qq -y install locales \
-    ca-certificates postgresql-client libpq-dev curl jq\
+    ca-certificates postgresql-client libpq-dev curl jq \
     python3-pip python3-icu python3-psycopg2 \
     python3-lxml python3-crypto \
     && apt-get -qq -y autoremove \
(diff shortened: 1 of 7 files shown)


mynameisfiber:
@sunu This PR is ready for another round of review! I've fixed up the merge conflicts and testing in dev.

The only issue I can see is that the _new_ xref'ed collections have the possibility to be sorted on the new field `doubt`. Currently that is kindof meaningless when done on old xref collections that don't have that field populated. However, this field isn't exposed to the UI so it really is an edge case for people playing with the API at the moment.

Dockerfile
```diff
@@ -29,13 +29,20 @@ WORKDIR /aleph
 ENV PYTHONPATH /aleph
 RUN pip install -q -e /aleph
 
+RUN mkdir -p /opt/ftm-compare/word-frequencies/ && \
+    curl -L -o "/opt/ftm-compare/word-frequencies/word-frequencies.zip" "https://public.data.occrp.org/develop/models/word-frequencies/word_frequencies.zip" && \
+    python3 -m zipfile --extract /opt/ftm-compare/word-frequencies/word-frequencies.zip /opt/ftm-compare/word-frequencies/ && \
+    curl -L -o "/opt/ftm-compare/model.pkl" "https://public.data.occrp.org/develop/models/xref/glm_bernoulli_2e_wf-v0.4.1.pkl"
```
sunu, line 35: I'm assuming the model is versioned to match FTM-compare? Are we going to generate a new model for each new version? Should we expect any weird behaviour if the library version doesn't match the model version?

How often are we going to generate the word frequencies? Should we version those as well, may be with a timestamp?

aleph/logic/xref.py
```diff
@@ -32,6 +37,47 @@
...
+def _bulk_compare_ftmc_model(proxies):
+    for score, confidence in zip(*MODEL.predict_proba_std(proxies)):
+        yield score, confidence, MODEL.version
+
+
+def _bulk_compare(proxies):
+    if not proxies:
+        return
+    if settings.XREF_MODEL is None:
+        yield from _bulk_compare_ftm(proxies)
```
sunu, line 70: Is it ok to add an explicit `return` after the `yield from` statements? I know the execution stops because the number of items we yield is controlled by the number of candidates we process in `_query_item`. But it was not very obvious to me when looking just at this function.

sunu (approved):
@mynameisfiber Thanks Micha! Everything looks good to me. I agree that the `doubt` field shouldn't be a problem for now. And I have added a couple of comments with some questions I had.

aleph/logic/xref.py
mynameisfiber: Good catch! It's definitely better to add the return after these `yield from`'s to make the flow more explicit in case things change in the future.

Dockerfile
mynameisfiber: Good questions. New models don't _need_ to be generated for each ftm-c version, but I expect many of the ftm-c commits being related to model updates which would require a model rebuild.

The way I was thinking about model version matching is:
- Major version must match
- Minor version is optional to match, but you may be missing out on a feature
- Patch versions are irrelevant

Word frequencies should probably also be labeled with the ftm-c version they were built with for the same reason.

For all of these static files, I imagine they only need to be rebuilt once per quarter at most. The word frequency file only needs to be updated if the entities in aleph significantly change and the model file only needs to be updated if the xref decisions on aleph significantly change.

PR #1837, opened 2021-06-04
Fix Elasticsearch OOM issue when fetching tags

sunu:
When querying Elasticsearch for tag counts or neighbour counts through
msearch batch the filters and queries into smaller groups so that ES
doesn't run out of memory when handling the request.

Fixes https://github.com/occrp/data.occrp.org/issues/177


sunu:
@kjacks We may need a copy change to go with this as well. In the live Aleph version, the UI says "No selectors were extracted from this entity" when the tags are yet to load. We should probably say something to indicate the tags are still being loaded or computed instead.
image

Rosencrantz (approved):

PR #1845, opened 2021-06-09
Abstract Entity Set Routing

### LINKED ISSUE #1804: Emails contain invalid URLs
https://github.com/alephdata/aleph/blob/main/aleph/logic/util.py#L39 - route doesn't exist in the UI. Probably should.

kjacks:
Fixes #1804


Rosencrantz (approved):
Would love some further context on why we need this, which I'm sure we do, I'd just like to understand the background

kjacks:
@Rosencrantz #1804

Rosencrantz:
@kjacks This is ready to merge now, right?

PR #1858, opened 2021-06-23
Handle long role names

### LINKED ISSUE #1853: Share box layout breaks for long usernames
Hi, when sharing an investigation with a user who uses a long email or username, the box looks weird.

image

kjacks:
Fixes #1853

### DIFF
--- a/ui/src/components/AuthButtons/AuthButtons.jsx
+++ b/ui/src/components/AuthButtons/AuthButtons.jsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
 import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
+import Truncate from 'react-truncate';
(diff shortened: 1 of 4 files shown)


Rosencrantz (approved):
Truncation could come back and bite us in the future, but I agree it's the right solution for now.

kjacks:
@Rosencrantz how so?

Rosencrantz:
> @Rosencrantz how so?

In my experience, especially when work with internationalised products there comes a point where truncating text becomes detrimental, either because it truncates too much, or truncates just enough to make the text confusing, non unique, or unitelligable.

kjacks:
Got it - yes, totally. In this case it feels manageable because we're just truncating the user's own name, so hopefully it shouldn't mess them up too much

Rosencrantz:
@kjacks Did you want to merge this?

kjacks:
@Rosencrantz it's ready to go

Issue #1831, opened 2021-05-31
Add `publisher` as a metadata field to the documents table


sunu:
related issues https://github.com/alephdata/memorious/issues/182

sunu:
Fixed in https://github.com/alephdata/aleph/pull/1862

PR #1862, opened 2021-06-25
Add a new document metadata field to store the publisher of a document

sunu:
Some of the existing fields were not being used; fix those as well. Fixes https://github.com/alephdata/aleph/issues/1831

### DIFF
--- a/aleph/model/document.py
+++ b/aleph/model/document.py
@@ -69,6 +69,7 @@ class Document(db.Model, DatedModel):
             "title",
             "summary",
             "author",
+            "publisher",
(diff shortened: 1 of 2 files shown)


sunu:
`publisher` is the only new field I'm adding.

`author`, `summary` and `date` are existing fields that were not being passed into the proxy object. These fields already exist in ingest.yml and in the meta field in our database.

`sourceUrl` was being set on the proxy object twice. I have only removed the duplicate statement.

Rosencrantz (approved):

Issue #1863, opened 2021-06-25
Provide the ability to attach debugpy to the Aleph API container


Rosencrantz:
Aleph does not currently make it super easy for developers to attach a debugger and step through the execution of the application. 

Let's start by having debugpy available for vscode users.

PR #1864, opened 2021-06-25
Rosencrantz/#1863 vscode debugger

Rosencrantz:
Addresses #1863


sunu (commented):
I did not know about `debugpy`. I just tried it out and the vscode integration is indeed very neat.

However, I don't think it's the go to debugger for Python and vscode may not be the editor everyone uses. So I think we should at least document how to debug with plain old `pdb` too. Generally adding this block:
```
stdin_open: true
tty: true
```
to the api section in the docker-compose file, setting a `breakpoint()` in the code and running `docker attach aleph_api_1` should give us access to a debugger shell. We don't need any extra config for the worker. Running `aleph worker` inside a `make shell` session will run the worker in an interactive shell and it will give us a debugger shell when the worker hits a breakpoint.

The vscode integration is so much nicer though. So we should have that as well. I would prefer to set it up in the dev environment without adding any extra env var or code. This is the set up I have, for example:
```diff
diff --git a/Makefile b/Makefile
index 583acaa73..46e1b6eb0 100644
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ web: services
        $(COMPOSE) up api ui
 
 worker: services
-       $(COMPOSE) run --rm app aleph worker
+       $(COMPOSE) run -p 127.0.0.1:5679:5679 --rm app python3 -m debugpy --listen 0.0.0.0:5679 /usr/local/bin/aleph worker
 
 tail:
        $(COMPOSE) logs -f
diff --git a/docker-compose.dev.yml b/docker-compose.dev.yml
index 0264b5d5d..f50f86ead 100644
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -99,9 +99,10 @@ services:
     build:
       context: .
     image: alephdata/aleph:${ALEPH_TAG:-latest}
-    command: aleph run -h 0.0.0.0 -p 5000 --with-threads --reload --debugger
+    command: python3 -m debugpy --listen 0.0.0.0:5678 -m flask run -h 0.0.0.0 -p 5000 --with-threads --reload --debugger
     ports:
       - "127.0.0.1:5000:5000"
+      - "127.0.0.1:5678:5678"
     restart: on-failure
     depends_on:
       - postgres
@@ -117,6 +118,7 @@ services:
       ALEPH_DEBUG: "true"
       ALEPH_PROFILE: "false"
       ALEPH_SECRET_KEY: "development"
+      FLASK_APP: "aleph.wsgi"
     env_file:
       - aleph.env
 
diff --git a/requirements.txt b/requirements.txt
index b8e5efe1d..053e5c9dd 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -42,3 +42,6 @@ Faker >= 5.6.0, < 8.0.0
 nose
 coverage
 coveralls
+
+# Development & debugging
+debugpy >= 1.3.0
```
It runs 2 debugpy debuggers - one for the api and one the worker. We can even add more for the ingest-file service if we need to.

More of a general comment on the PR: may be we should clean it up a bit so that the history from other branches/commits doesn't get included. For example, we definitely don't want the 3.14.4 to 3.14.5 version bump commit since that didn't really happen. A littered history will make it hard to follow the changes later.

sunu:
And btw here's what my `launch.json` vscode debugger configuration looks like for the debugpy debuggers
```js
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Python: Aleph API",
            "type": "python",
            "request": "attach",
            "connect": {
                "host": "localhost",
                "port": 5678
            },
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}/aleph",
                    "remoteRoot": "/aleph/aleph"
                }
            ]
        },
        {
            "name": "Python: Aleph Worker",
            "type": "python",
            "request": "attach",
            "connect": {
                "host": "localhost",
                "port": 5679
            },
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}/aleph",
                    "remoteRoot": "/aleph/aleph"
                }
            ]
        },
    ]
}
```

Rosencrantz:
@sunu Love it. I struggled to get Aleph running using only changes to the Makefile, so thank you for showing me the way. With this setup, am I right in thinking that the debugger will run by default. If that is the case we'll need to do something when code is pushed to the remote. I had problems with the builds hanging because they were waiting for a debugger to attach.

Rosencrantz:
We should probably add some instructions to the documentation that make it clear how a developer can debug.

sunu:
> @sunu Love it. I struggled to get Aleph running using only changes to the Makefile, so thank you for showing me the way. With this setup, am I right in thinking that the debugger will run by default. If that is the case we'll need to do something when code is pushed to the remote. I had problems with the builds hanging because they were waiting for a debugger to attach.

Yes, the debugger will run by default. But we can always tweak the command used for testing, building etc. Do you remember which command was hanging?

> We should probably add some instructions to the documentation that make it clear how a developer can debug.

Yes, absolutely! We should document both the `pdb` + `shell` approach and the `debugpy` + `vscode` approach.

Rosencrantz:
"Do you remember which command was hanging?" For me, on this branch if I leave ALEPH_DEBUGGER set to true then it will hang in the worker.py file on the debugpy.wait_for_client() line. This pauses execution until the vscode debugger attaches.

sunu:
I propose we document how to do it by adding the flag but not make it the default behaviour. Personally, as a developer new to the code base, I would expect the debugger to interfere only when I set a breakpoint; otherwise it becomes confusing if the program hangs.

We could also document some useful places to set a breakpoint so that it's possible to debug from the beginning without the flag. That would give some helpful context to the developer as well. For example, setting a breakpoint in `Worker.handle` lets you debug all tasks executed on the worker from the beginning. Setting a breakpoint in `aleph.view.context.setup_request` will let you debug all requests from the beginning.

sunu (approved):

Issue #1819, opened 2021-05-25
Fix up facets API for non-standard property types


pudo:
Via #1803, users can choose random index fields to facet on, especially all the `properties.bla` fields. However, facets need some post-processing to be useful to the user - e.g. country codes are supposed to be expanded into codes and labels. This happens in a manner that's currently hard-coded to certain index fields here: 

https://github.com/alephdata/aleph/blob/main/aleph/search/result.py#L76

We need to make this more flexible to that it supports all `followthemoney.types:registry.groups` and all properties defined in ftm.

kjacks:
<img width="516" alt="Screen Shot 2021-06-02 at 12 51 01" src="https://user-images.githubusercontent.com/6720200/120468358-69604580-c3a1-11eb-8fd5-4ceb93029565.png">

versus 

<img width="565" alt="Screen Shot 2021-06-02 at 12 51 49" src="https://user-images.githubusercontent.com/6720200/120468381-711fea00-c3a1-11eb-901d-47712d89bc2f.png">

PR #1865, opened 2021-06-25
Rosencrantz/#1819 update facet api

Rosencrantz:
This is where the work for #1819 lives. This is connected to #1803.

@kjacks So. The way that this currently works is to send over the property type along with other information. This is then used to make a basic call on how/what type of formatting we want to add to the label that is shown. Currently it handles country, language, date, most other property types are still treated as a string with no extra formatting.


sunu (commented):
I feel weirded out by the amount of extra unrelated diffs that has ended up in this PR. I propose we make a new one by `git pick`ing the relevant commits to a new branch. I can quickly do that if it's ok.

Rosencrantz:
@sunu If you feel weirded out by this and feel strongly that we shouldn't merge then I'm happy for you to make a seperate branch with these changes :)

sunu:
@Rosencrantz I have opened https://github.com/alephdata/aleph/pull/1881 with the changes. Let's close this and move the discussion there?

PR #1881, opened 2021-07-02
Rosencrantz/#1819 update facet api

sunu:
Continues the work from https://github.com/alephdata/aleph/pull/1865

### DIFF
--- a/aleph/search/facet.py
+++ b/aleph/search/facet.py
@@ -29,6 +29,9 @@ class Facet(object):
     def update(self, result, key):
         pass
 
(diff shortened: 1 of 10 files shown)


Rosencrantz:
My lesson from this is to branch from develop when starting new work, and in doing so, reduce the number of commits and keep Sunu happy.

sunu:
Haha, we have to branch from the target branch to keep the commits relevant. It may not always be `develop`. I branched this PR from `kirk/configurable-facets` branch for example.

Rosencrantz (approved):

PR #1890, opened 2021-07-07
Reduce batch size in Elasticsearch bulk delete query

### LINKED ISSUE #1887: High CPU usage on Elasticsearch when flushing large collections
When flushing large collections, Elasticsearch CPU usage regularly reaches 100% which impacts overall performance of Aleph until the task is complete.

sunu:
The default batch size for delete_by_query is 1000. But it makes our
Elasticsearch cluster reach 100% CPU usage during large delete
operations. A smaller batch size keeps the CPU load under control.

Fixes #1887


Rosencrantz (approved):

Rosencrantz:
On hold... We want to test it out separately after the search facets stuff is out. So that it is easy to roll back, goes out in a different release.

Issue #1876, opened 2021-06-28
Change date filter default cutoff date from 1900 to 1970


kjacks:
1900 feels a bit arbitrary and still too wide of a range for the default cutoff.

1970 was proposed as a somewhat standard cutoff date.

<img width="364" alt="Screen Shot 2021-06-28 at 12 49 53" src="https://user-images.githubusercontent.com/6720200/123624830-5d846980-d80f-11eb-83d1-f164cfa06308.png">

PR #1901, opened 2021-07-19
Search Results: use date histogram for all date-type property filters

### LINKED ISSUE #1893: Search results: handle date filter values with varying precision
With the ability to add date-type property filters introduced in #1803 , users can now select individual dates to filter search results by. 

Though the rest of Aleph supports dates of varying precision (i.e. YYYY, YYYY-MM, or YYYY-MM-DD), the filter value processing in the backend incorrectly converts all dates to January first of the given year.

Solution 1: Modify backend date processing to support varying precision of dates

Solution 2: Utilize the date histogram for all date-type properties

Screen Shot 2021-07-14 at 14 09 55

kjacks:
Fixes #1893 

@sunu @Rosencrantz - I realized we could use the existing supported elastic search operators on these date properties for the histogram API requests, so it was only some front-end component adaptation to show all of the date props as histograms :)

### DIFF
--- a/ui/src/app/Query.js
+++ b/ui/src/app/Query.js
@@ -191,14 +191,15 @@ class Query {
 
   addFacet(facet) {
     const field = facet.isProperty ? `properties.${facet.name}` : facet.name;
-    const newQuery = this.add('facet', field)
+    let newQuery = this.add('facet', field)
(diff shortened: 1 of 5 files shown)


kjacks:
Also, fixes #1876

Rosencrantz (approved):

PR #1905, opened 2021-07-22
Search results list formatting

### LINKED ISSUE #1844: Search results list formatting
With the ability to add additional custom columns to the search results screen introduced in #1803 , the results table can grow quite large and unwieldy, leading to cutoff and unreadable property values.

Proposal
- Allow users to collapse/uncollapse the facets left sidebar, allowing for more horizontal space for the results

- In the results table, fix the caption column and allow the rest of the table to scroll horizontally, with columns having a minimum width
-- should column width should be determined by the type - i.e. date and numeric type columns should have a smaller default width?

- Value truncation:
-- For values with > 2 listed items, display "Deutsche Bank and X more"
-- Entity cells are allowed to take as many lines as they need. Cells of other types should be truncated to a maximum of 2 lines with ellipses

Other
- [ ] Standardize no-value placeholder for all property value and type display components (currently they use a mix of "-", "--", or "")

kjacks:
Fixes #1844 

- Allows users to collapse/expand left-side facets to increase horizontal space for results properties
- Sets a minimum width for all columns in the results table and allows for horizontal scrolling with fixed caption column

@sunu @Rosencrantz curious your thoughts on the user experience for both of these :)

https://user-images.githubusercontent.com/6720200/126672623-ca3ac382-5884-4a1d-85ef-e07559551706.mov

https://user-images.githubusercontent.com/6720200/126672660-e484f793-cbfb-41f1-8d29-cc1e51817a28.mov


Rosencrantz:
Hey Kirk

I love the second idea, being able to close that sidebar once you’re finished with it is a great idea and allows the journalist to focus on the search results. One thought here is whether that might mean we need to do something with the button that allows adding extra facets…

I don’t mind the first idea, I think it’s better than having to scroll the entire page, which never feels quite right to me.

Perhaps these are something that we could trial with our journalists? Maybe put them both under a feature flag and then send out some comms to let people know what they can do…

One awesome bonus would be if we could track who enables and uses these features. It would be a great help in determining whether they’re useful or just noise.

Thanks Kirk, this is great stuff

Steve Haffenden
Aleph Lead
OCCRP
Tyrol, Austria

+436603886980
@smhaffenden

kjacks:
@Rosencrantz great! 
- I'm not too worried about people not finding the configure facets button, because the facets panel is still open by default, and I don't think it's a button that someone needs to see at all times, but only when they want to do deeper configuration.
- Yehh I certainly don't love anything with horizontal scrolling - but I can't really think of anything else we can do to accommodate the extra width?

sunu:
I like both the changes. Horizontal scroll is not the best thing in the world, but I think it's useful in this case.

kjacks:
Along with a bunch of other little tweaks, I also added togglable show more/show less value truncation to allow users to view fields with many values without breaking the layout

ui/src/components/EntitySearch/EntitySearchResults.scss
```diff
@@ -1,14 +1,19 @@
...
 
+.EntitySearchResults-outer-container {
+  position: relative;
+}
+
+.EntitySearchResults-inner-container {
+  overflow-x: scroll;
+  overflow-y: visible;
+  padding: 0;
+  position: relative;
```
Rosencrantz, line 12: @kjacks Curious whether you deliberately organise these to be alphabetically ordered, wether it's handled by a tool, or is it just luck?

ui/src/components/EntitySearch/FacetedEntitySearch.jsx
```diff
@@ -78,6 +88,36 @@ class FacetedEntitySearch extends React.Component {
...
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.checkMobileWidth);
+  }
+
+  checkMobileWidth() {
+    const width = this.ref.current?.clientWidth;
+    if (!width) { return; }
+
+    if (width < 620) {
```
Rosencrantz, line 108: Nitpick, what is 620 here? Might it be better wrapped in a variable name

ui/src/components/EntitySearch/EntitySearchResults.scss
kjacks: heheh nope just luck XD

In an ideal world I guess I would try to follow one of the CSS style guides like SMACSS or BEM to determine the order buuuut have never been quite sure if it's worth it...

ui/src/components/EntitySearch/FacetedEntitySearch.scss
```diff
@@ -46,18 +52,81 @@
...
+    transition: width 120ms linear;
+    min-height: 100vh;
+    width: $aleph-sidebar-width !important;
+
+    .collapsed & {
+      width: 0 !important;
+      padding: 0;
+    }
+  }
+
```
Rosencrantz, line 85: is the !important required for some reason? I've never been a fan.

ui/src/components/EntitySearch/FacetedEntitySearch.jsx
kjacks: Very valid - I'll make the change :)

Rosencrantz (approved):

ui/src/components/EntitySearch/FacetedEntitySearch.scss
kjacks: Mostly laziness tbh

PR #1907, opened 2021-07-26
Unauthenticated request handling

kjacks:
Require auth for all non-index routes for apps with `metadata.auth.auth_required == true`.

TODO:

- [ ] add metadata flag to OCCRP Aleph instance metadata settings

### DIFF
--- a/aleph/settings.py
+++ b/aleph/settings.py
@@ -85,6 +85,9 @@ OAUTH_ADMIN_GROUP = env.get("ALEPH_OAUTH_ADMIN_GROUP", "superuser")
 # No authentication. Everyone is admin.
 SINGLE_USER = env.to_bool("ALEPH_SINGLE_USER")
 
+# Require authentication. No anonymous access
+REQUIRE_AUTH = env.to_bool("ALEPH_REQUIRE_AUTH", False)
+
 # Default session duration.
 SESSION_EXPIRE = 800_000 if SINGLE_USER else 60_000
 SESSION_EXPIRE = env.to_int("ALEPH_SESSION_EXPIRE", SESSION_EXPIRE)
--- a/aleph/settings.py
+++ b/aleph/settings.py
@@ -86,7 +86,7 @@ OAUTH_ADMIN_GROUP = env.get("ALEPH_OAUTH_ADMIN_GROUP", "superuser")
 SINGLE_USER = env.to_bool("ALEPH_SINGLE_USER")
 
 # Require authentication. No anonymous access
-REQUIRE_AUTH = env.to_bool("ALEPH_REQUIRE_AUTH", False)
+REQUIRE_LOGGED_IN = env.to_bool("ALEPH_REQUIRE_LOGGED_IN", False)
 
 # Default session duration.
 SESSION_EXPIRE = 800_000 if SINGLE_USER else 60_000
--- a/aleph/views/__init__.py
+++ b/aleph/views/__init__.py
@@ -1,6 +1,3 @@
-from werkzeug.exceptions import Unauthorized
-from flask import request
-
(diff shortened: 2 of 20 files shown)


Rosencrantz (approved):

kjacks:
@sunu I'm having an issue when I try to run this locally with the `ALEPH_REQUIRE_AUTH` set to true. I'm unable to log in because it seems like it's blocking the login request with a 401

sunu:
@kjacks Arrh, should be fixed now!

Do you think we can move the message to markdown so that it can be easily customised to be something not OCCRP specific without messing with the code? May be we can also decouple the homepage message from the auth handling part so that it can be used for other messages as well. What do you think?

aleph/views/__init__.py
```diff
@@ -19,25 +22,44 @@
...
+
+
+def require_auth():
+    authz = get_authz(request)
+    if authz is None:
+        raise Unauthorized()
+
+
+def register_blueprint(app, blueprint, **kw):
+    if settings.REQUIRE_AUTH:
```
pudo, line 36: My intuition would be that this wants to live on `authz` somewhere to make it more coherent with the rest of the system. Maybe like a `authz.can_browse_anonymous` that gets `require`'d when no other check is conducted (and is also implicitly checked by `can()` etc. - analogous to `session_write`? Doing it at the blueprint layer introduces a whole new dimension to authz that's sort of invisible and unexpected.

aleph/settings.py
```diff
@@ -85,6 +85,9 @@
 # No authentication. Everyone is admin.
 SINGLE_USER = env.to_bool("ALEPH_SINGLE_USER")
 
+# Require authentication. No anonymous access
+REQUIRE_AUTH = env.to_bool("ALEPH_REQUIRE_AUTH", False)
```
pudo, line 89: `REQUIRE_LOGGED_IN` maybe?

pudo (commented):
Have you considered whether just improving the rate limiting mechanism and bringing down its limits could be a valid alternative to doing this?

Rosencrantz:
> Have you considered whether just improving the rate limiting mechanism and bringing down its limits could be a valid alternative to doing this?

Yes, it's something that we need to look at. As a first step though, this allows us to shut down potential intrusions quickly. It's a bit of a big hammer and there are certainly finer tools that we should be using but for now this will work for us.

aleph/views/__init__.py
sunu: Ok, in hindsight that's definitely a much cleaner implementation. Thanks @pudo!

sunu:
> Have you considered whether just improving the rate limiting mechanism and bringing down its limits could be a valid alternative to doing this?

The best I can think of is IP based rate limiting for anonymous requests. The traffic we saw yesterday kept switching between tor exit nodes. So I'm not sure if IP based rate limiting will be effective in that case. Is there any other criteria we can use?

aleph/settings.py
Rosencrantz: Either makes sense

Rosencrantz (approved):

ui/src/components/Screen/Screen.jsx
```diff
@@ -75,10 +82,15 @@ export class Screen extends React.Component {
...
 
-const mapStateToProps = state => ({
-  metadata: selectMetadata(state),
-  session: selectSession(state),
-});
+const mapStateToProps = (state, ownProps) => {
+  const { location } = ownProps;
+
+  return ({
+    isHomePage: location.pathname === '/',
```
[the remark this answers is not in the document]
kjacks: Ooh yes! Thank ya for pointing that out :)

Rosencrantz (approved):

PR #1922, opened 2021-08-17
Use the caption of entities as label for facet values

### LINKED ISSUE #1880: Search results: handle entity filter values
Currently, entity filter values are not serialized correctly in the search results facets response

As a temporary fix, we have removed all entity type properties from the facet configuration options introduced in #1803 

Screen Shot 2021-06-30 at 11 42 05

sunu:
Fixes #1880

### DIFF
--- a/aleph/search/facet.py
+++ b/aleph/search/facet.py
@@ -93,14 +93,6 @@ class EventFacet(Facet):
         result["label"] = key if event is None else event.title
 
 
-class DateFacet(Facet):
-    def update(self, result, key):
-        result["label"] = datetime.strptime(key, "%Y-%m-%dT%H:%M:%S").date()
-
-    def get_key(self, bucket):
-        return bucket.get("key_as_string")
-
-
 class EntityFacet(Facet):
     def expand(self, keys):
         for key in keys:
--- a/aleph/search/facet.py
+++ b/aleph/search/facet.py
@@ -1,7 +1,7 @@
 from followthemoney import model
 from followthemoney.types import registry
 
-from aleph.model import Collection, Events
+from aleph.model import Collection, Events, Entity
 from aleph.logic import resolver
 
 from datetime import datetime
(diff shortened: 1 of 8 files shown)


aleph/search/facet.py
```diff
@@ -93,12 +93,17 @@ def update(self, result, key):
         result["label"] = key if event is None else event.title
 
 
-class DateFacet(Facet):
```
Rosencrantz, line 96: Are there any compatibility issues with removing this DateFacet? Is there a chance that someone might be relying on this and that it might break their code?

aleph/search/facet.py
```diff
@@ -93,12 +93,17 @@ def update(self, result, key):
...
-    def update(self, result, key):
-        result["label"] = datetime.strptime(key, "%Y-%m-%dT%H:%M:%S").date()
+class EntityFacet(Facet):
+    def expand(self, keys):
+        for key in keys:
+            resolver.queue(self.parser, Entity, key)
+        resolver.resolve(self.parser)
 
-    def get_key(self, bucket):
```
Rosencrantz, line 100: Are there any issues with removing this function?

Rosencrantz (commented):
I would have expected that there would be a need to remove tests of the EntityFacet class and get_key function, but I assume that these did not come with tests. We should take the time now to great tests for this stuff.

sunu:
Yeah, I have added a test to check the functionality at the api level.

aleph/search/facet.py
sunu: It hasn't been released yet and I don't think the frontend needs it any more since we switched to histograms for all date type fields. Correct me if I'm wrong here @kjacks

aleph/search/facet.py
sunu: No, it's part of the DateFacet class we're no longer using.

aleph/search/facet.py
kjacks: Correct :) It's no longer needed by the ui

Rosencrantz (approved):

kjacks (approved):

PR #1975, opened 2021-09-24
Bump pantomime

Rosencrantz:


requirements.txt
```diff
@@ -5,7 +5,7 @@ followthemoney-store[postgresql]==3.0.3
 fingerprints==1.0.3
 servicelayer[google,amazon]==1.18.2
 normality==2.2.3
-pantomime==0.4.1
+pantomime==0.5.1
```
sunu, line 8: There is a PR from dependabot for pantomime already: https://github.com/alephdata/aleph/pull/1951. Why do we need to do this manually?

sunu (commented):
Also, we want `develop` as the target vs `main` I think?

requirements.txt
Rosencrantz: Didn't see the depdabot bump, hence why this exists. Will close this give that it's pointing at the wrong branch.

PR #2048, opened 2021-11-07
Add support for Oauth2 Bearer Tokens

goctav:
**Context**
In order to call Alephs REST endpoints using a JWT token,
support for "Bearer" tokens needs to be added. With it,
Aleph will be able to serve requests from other clients
that use OIDC with the same Identity Server.

**Solution**
The changes are somewhat similar to the logic present in
the OAuth callback endpoint. Aleph tries to determine the
Role based on the foreign ID derived from the JWT token fields
but if it doesn't find one it will create it.
There is no support for groups yet since I haven't found the
need just yet, later tests with an actual Aleph setup and a
configured Indentity Server may uncover that they're needed.

### DIFF
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -166,6 +166,19 @@ class Authz(object):
         )
 
     @classmethod
+    def from_access_token(cls, token):
+        email = token.get("email", token.get("upn"))
+        sub = token.get("sub", email)
+        if not sub:
+            raise Unauthorized()
+        role_id = "%s:%s" % (settings.OAUTH_HANDLER, sub)
+        role = Role.by_foreign_id(role_id)
+        if role is None:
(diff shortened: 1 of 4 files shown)


goctav:
At the moment the testing part is superficial, any advice around how to better test the changes are welcome. I was not able to locate the unit tests folder just the integration ones.

Rosencrantz:
@sunu Do you think that this is something that we can add in to Aleph?

sunu:
@Rosencrantz This was mostly meant to facilitate the Aleph-Confluence plugin prototype. Since we haven't made any progress there, I would like to hold off adding a new form of authentication until we actually need to.

Rosencrantz:
This isn't going to get merged any time soon unfortunately, so closing for now.

PR #2053, opened 2021-11-08
Allow for month / day selection in date histogram filter

### LINKED ISSUE #2021: Date filter / histograms should allow for finer selection of date range
The histogram date filter in the facet selection would be an even greater tool if the user was able to select more granular time spans. After selecting a year or a span of years, the user should be able to select month, days, in some instances maybe even hour of the day.

kjacks:
Fixes #2021 

Allow for more granular selection of month / day date ranges in the date histogram filter

To Do: currently dates with no month or no day specified are returned from the backend in the first interval of the given date facet results. So an entity with the property value "2010" is returned in the January 2010 interval, and an entity with the property value "2010-03" is returned in the March 1, 2010 interval. @sunu how possible would it be to separate these results into their own interval bucket?

histogram


sunu:
Can you describe the ideal outcome you would expect in those cases? I'm afraid it may not be possible through ES directly and we might need to do some post processing on the result before returning it.

kjacks:
The ideal functionality would be: for the given `facet_interval` in the request to create an additional interval bucket to contain null values for that facet_interval. So if the `facet_interval` on the request was set to month, then there would be an additional interval bucket to contain any values that fall within the range but do not have a month specified.

Screen Shot 2021-11-08 at 12 50 46

sunu:
So Elasticsearch stores dates as a long number representing milliseconds-since-the-epoch. And the aggregations are range queries against these long numbers. That means the aggregations can't really see the date format because it's already been converted to a single number.

Not sure if we can solve this without adding extra fields to store parts of dates separately. @pudo do you have any advice on how this could be solved?

Rosencrantz (approved):
There's a lot going on here, and I don't have the depth of knowledge yet to be able to give this an in-depth review. The logic/code structure seems good.

Rosencrantz:
@sunu I think that the outstanding work here (dealing with more vague dates) is something that we can address down the line. Having played with this a little I think that we can merge this work in and move forward. Unless you have a strong desire not to do that?

sunu:
As long as we're not hiding the uncertainty of the dates, I don't mind merging this. iirc Kirk added a warning to clarify that the dates might not be accurate in some cases. So it should be ok.

Rosencrantz (approved):

Issue #2036, opened 2021-10-26
Google OAuth with aleph 3.11.0


thimios:
I upgraded Aleph from 3.8.9 to 3.11.0

This is the configurations I **used to have** with version **3.8.9** and was working:

```
ALEPH_OAUTH=true
ALEPH_OAUTH_HANDLER=google
ALEPH_OAUTH_KEY=xxxxxxxxxxxxxxxx.apps.googleusercontent.com
ALEPH_OAUTH_SECRET=xxxxxxxxxxxxxxxxxxxxxx
ALEPH_OAUTH_SCOPE=openid email profile
ALEPH_OAUTH_BASE_URL=https://www.googleapis.com/oauth2/v2/
ALEPH_OAUTH_TOKEN_URL=https://www.googleapis.com/oauth2/v4/token
ALEPH_OAUTH_AUTHORIZE_URL=https://accounts.google.com/o/oauth2/v2/auth?access_type=offline&prompt=consent
```

In my .env file **I now have with version 3.11.0** the following configuration:

```
ALEPH_OAUTH=true
ALEPH_OAUTH_METADATA_URL=https://accounts.google.com/.well-known/openid-configuration
ALEPH_OAUTH_KEY=123456789999-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.apps.googleusercontent.com
ALEPH_OAUTH_SECRET=xxxxxxxxxxxxxxxxxxxxxxxx
```
I have removed all other OAUTH related configuration variables. 

In google Cloud Console I have configured an "OAuth 2.0 Client ID" and in "Authorised redirect URIs" I have: "https://sanctions.givve.com/api/2/sessions/callback" which used to work with my previous version of aleph (3.8.9), unlike mentioned in: https://docs.alephdata.org/developers/installation#using-google-oauth ("http://localhost:8080/api/2/sessions/callback/google").

When clicking on "Sign in via OAuth" I get the google dialog to select which google account to login with, but after selecting one, I get redirected to the url:

[...]

Issue #2003, opened 2021-10-07
oauth.py _parse_access_token function uses access_token instead of id_token


keanpaderes:
OS: MacOS Catalina 10.15.7
Docker Version: 20.10.7
Aleph Version: 3.11.0

Good day! We stumbled upon an error while integrating Google OAuth to our Aleph Instance. We have traced the problem to `aleph/aleph/oauth.py`, specifically the `_parse_access_token` function (line 28). In its current form, it uses the `access_token` property which leads to an error when authlib's JsonWebToken.decode function parses it as `access_token` is not JWT. It should use `id_token` property which is also available in the `oauth_token` parameter and we confirmed that it works when we applied the change on our local image.

![Screen Shot 2021-10-07 at 4 29 30 PM](https://user-images.githubusercontent.com/11690058/136348518-d2427bb4-ce7a-4bb7-8bc8-0163a8718416.png)

thimios:
https://github.com/alephdata/aleph/pull/2054

sunu:
Hey everyone, I know multiple people are facing this same issue and we want to help fix it soon. But unfortunately we are blocked on some urgent internal work at the moment. We'll get to fixing the issue as soon as we can and ship it in the next Aleph release.

PR #2054, opened 2021-11-09
fix oauth with google provider

thimios:
https://github.com/alephdata/aleph/issues/2036

https://github.com/alephdata/aleph/issues/2003

I guess this could break getting the groups with other oauth providers, but it seems that all the specs are green and it works with google cloud oauth as I have tested it.

### DIFF
--- a/aleph/oauth.py
+++ b/aleph/oauth.py
@@ -26,7 +26,7 @@ def configure_oauth(app, cache):
 
 
 def _parse_access_token(provider, oauth_token):
-    token = oauth_token.get("access_token")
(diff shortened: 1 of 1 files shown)


sunu:
Hey @thimios, thanks for the PR. Unfortunately, you're right and this might break Aleph's integration with other OAuth providers such as keycloak. We don't have tests for OAuth integrations yet. So the passing tests are not a very helpful indicator here, I'm afraid.

sunu:
Closing in favour of #2062

PR #2062, opened 2021-11-19
Fix issue with Google OAuth integration

### LINKED ISSUE #2003: oauth.py _parse_access_token function uses access_token instead of id_token

sunu:
OIDC doesn't specify the format for access_token. Some providers use it
as a JWT to store information about groups (eg: Keycloak), some don't
use it as a JWT (eg: Google).

So we try to parse it as a JWT but if that fails we ignore the error
and move on. In case the access_token is not an JWT, the groups info is
stored in the id_token and that's where we look.

Fixes #2003

### DIFF
--- a/aleph/oauth.py
+++ b/aleph/oauth.py
@@ -1,6 +1,8 @@
 import logging
+from pprint import pformat  # noqa
 from authlib.jose import JsonWebToken, JsonWebKey
 from authlib.integrations.flask_client import OAuth
+from authlib.jose.errors import DecodeError
 
 from aleph import settings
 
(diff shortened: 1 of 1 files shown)


sunu:
Hey @thimios and @keanpaderes! Can you please check if this fixes the issue with Google OAuth for you?

Also, I'm curious if any of you use groups to manage users and whether you know where Google stores that info in the OAuth payload? It seems groups are only available in a Google Workspace environment. And I don't have access to one to test that functionality.

thimios:
@sunu the fix works for our setup. 

We do not use google groups to manage the users, so I do not know much about them. 

Thanks for pushing this forward. It would be great to have some specs covering the OAuth functionality so that it does not break again in the future.

PR #2096, opened 2022-01-04
Update helm examples

monneyboi:
When setting up a local staging environment, i noticed that in the helm examples, elastic was using an old image and the ingresses were using a deprecated api.

besides that i removed configuration that matched the defaults, and removed the `PREFIX` from the makefile as "aleph" was also used in the default secrets.


sunu:
Thanks @monneyboi!

PR #2112, opened 2022-02-02
Hotfix: Use the correct collection_id while generating xref matches

sunu:
Match.collection_id should be set to the collection_id of the xref match


mynameisfiber (approved):
Looks great to me! :hammer:

PR #2115, opened 2022-02-04
Convert null value to string

Rosencrantz:


ui/src/util/getEntitySetLink.js
```diff
@@ -1,3 +1,4 @@
 export default function getEntitySetLink(entitySet) {
-  return (entitySet && entitySet.id) ? `/${entitySet.type}s/${entitySet.id}` : null;
+
```
sunu, line 2: Do we use a code formatter for our JS code? Personally I would prefer to not have a blank line here.

Other than that, it looks good. I think it should avoid the error. I'll give it a spin on Monday

Rosencrantz: We do, eslint, but for some reason it did not pick up on this . I've updated the code manually.

sunu (approved):

PR #2145, opened 2022-02-25
Set the node-sass version to 6.0.1 until we can start using dart instead

Rosencrantz:


sunu (commented):
I don't have an m1 machine to test this unfortunately.

Could we create an issue for adopting a replacement? Also, downgrading to node-sass < 7.0.0 means we're reintroducing this security issue? https://github.com/alephdata/aleph/security/dependabot/1

Rosencrantz:
The CVSS score for the vuln is pretty middling so I'm not overly concerned about it. Good suggestion on the ticket creation. Will create one.

PR #2154, opened 2022-03-11
Fix search handler on home screen

tillprochaska:
This PR fixes the submit handler for the search input on the home screen.

To reproduce, open the [OCCRP Aleph instance](https://data.occrp.org), make sure you’re logged-out, enter any string into the search input, and press enter. The app should navigate to the `/search` route, but instead nothing happens.

Navigation fails, because the navigate prop on the `HomeScreen` component is undefined. The `HomeScreen` component needs to be wrapped in the `withRouter` HOC in order to have access to the `navigate` prop.

### DIFF
--- a/ui/src/screens/HomeScreen/HomeScreen.jsx
+++ b/ui/src/screens/HomeScreen/HomeScreen.jsx
@@ -9,6 +9,7 @@ import { connect } from 'react-redux';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import { Callout, Intent } from '@blueprintjs/core';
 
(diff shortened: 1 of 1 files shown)


sunu:
@Rosencrantz Could you take a look at this?

Rosencrantz:
@tillprochaska Thanks for taking a look at this, all looks good, but I need to verify the change locally. In the meantime could I ask you to change the pull request to merge into our develop branch?

tillprochaska:
@Rosencrantz Sorry, I missed the develop branch when opening the PR. Thanks @sunu for making the change!

PR #2155, opened 2022-03-11
update about page w/ new form link

jlstro:
Update the Aleph about page to use the new form for FOO/BAR requests


Rosencrantz (approved):
Nice

PR #2158, opened 2022-03-14
Specify aleph api image specifically in helm charts

etymono:
I wanted to create & [deploy an image](https://github.com/ftmnl/pages) to override the default info pages, as i [read here in the docs](https://docs.alephdata.org/developers/technical-faq#can-i-customise-the-text-of-the-about-page) but the helm charts did not support overriding the API image only, as the API chart used the `global.image` config.

I tried overriding the `global.image` config with my own API image, but the UI deployment uses the same `global.image.tag` field, so that did not work.

This is the most simple fix i could think of, seperate `api.image` config values.

### DIFF
--- a/helm/charts/aleph/templates/api.yaml
+++ b/helm/charts/aleph/templates/api.yaml
@@ -20,7 +20,7 @@ spec:
       securityContext: {{- toYaml .Values.api.podSecurityContext | nindent 8 }}
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.global.image.repository }}:{{ .Values.global.image.tag }}"
+          image: "{{ .Values.api.image.repository }}:{{ .Values.api.image.tag }}"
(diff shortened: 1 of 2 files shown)


sunu:
Hey @monneyboi! What do you think about building an ui image additionally to match the api image tag?

If we're gonna use a separate tag, I would prefer to add that to the ui rather than the api. We use `global.image` for the api, worker and the upgrade job. So I would prefer if we keep those the same.

monneyboi:
> What do you think about building an ui image additionally to match the api image tag?
Feels like wasting CPU cycles to me

> I would prefer to add that to the ui rather than the api.
Ah yeah, that sounds reasonable, i'll update the pull request

Rosencrantz:
@monneyboi Hey Fella, did you ever come back to this?

Rosencrantz:
Closing due to inactivity...

PR #2199, opened 2022-04-05
Switch to rabbitmq based task queue

sunu:


aleph/worker.py
```diff
@@ -32,14 +33,25 @@
...
 log = structlog.get_logger(__name__)
 
 app = create_app(config={"SERVER_NAME": settings.APP_UI_URL})
 
 
-def op_index(collection, task):
-    sync = task.context.get("sync", False)
-    index_many(task.operation, collection, sync=sync, **task.payload)
+def op_index(collection_id, batch, worker=None):
+    collection = Collection.by_id(collection_id)
```
pudo, line 44: Are you still ensuring that all the documents in the whole batch are from the same collection? I would consider sending the collection as a dict as part of the indexing requests pretty soon after making this change, I think it'll be a perf game changer...

sunu: @pudo So we want to batch indexing tasks irrespective of collection?

PR #2223, opened 2022-04-20
Fix label in ensure_collection

robbarry:
When calling `aleph crawldir` with a `foreign_id` of an already existing collection, `ensure_collection` was overwriting the collection's name (`label`).

The assumption appeared to be that the path name would be used as the collection's label, but this isn't always the desired behavior.

This PR adds a new flag `-n` to `aleph crawldir` allowing the user to specify the name (`label`) of the collection.

If `-n` isn't called, `aleph crawldir` does not attempt to give the collection a label.

### DIFF
--- a/aleph/manage.py
+++ b/aleph/manage.py
@@ -61,8 +61,9 @@ def ensure_collection(foreign_id, label):
     authz = Authz.from_role(Role.load_cli_user())
     config = {
         "foreign_id": foreign_id,
-        "label": label,
     }
+    if label is not None:
+        config["label"] = label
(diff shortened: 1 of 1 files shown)


sunu (changes requested):
Hi @robbarry, the changes look good but I have one issue with it. With these changes, if we run `aleph crawldir /path/to/directory` to create a new collection, the collection won't have a label - which isn't ideal.

We could check whether the collection exists or not, and either force the user to provide a name or use `path.name` as the label if a name isn't provided explicitly.

PR #2246, opened 2022-05-01
Pass is_admin to Role#load_or_create() based on ALEPH_ADMINS

msmych:
* Extracted logic that checks whether an email is admin from `Role#load_or_create()` to `settings#is_admin()`. Not sure if functions are allowed/welcome there, otherwise it can be moved to a more appropriate place. Reason for extraction: not to mix persistence logic with business logic, especially when there's already `is_admin` param available.
* Changed `is_admin` param default value from `None` to `False`, moved `role.is_admin` assignment to role creation block
* Passing `is_admin` param from API and OAuth


aleph/settings.py
```diff
@@ -64,6 +64,10 @@
 # Assumes a comma-separated list.
 ADMINS = env.to_list("ALEPH_ADMINS")
 
+def is_admin(email):
```
sunu, line 67: I think we should move this to `logic/roles.py` and rename it to something like `is_auto_admin` to be more explicit. We try not to put app logic in settings directly.

sunu (commented):
Hi @msmych, it looks good to me from a quick look. I have added a nit pick. Let me know what you think

aleph/settings.py
msmych: It's a bit tricky — this causes circular dependency in imports: `aleph.core → aleph.oauth → core.logic.roles → aleph.core`

I think the right solution would be splitting `aleph.core` into core that would provide things like `db` and `cache` that would be imported in different components, and `aleph.app` that will only import everything needed to start the app.

For now moved `is_auto_admin` to `aleph.util`

msmych:
Updated and commented

Rosencrantz:
@sunu Are you happy with this now?

sunu (approved):

PR #2260, opened 2022-05-06
Add config file examples for traefik/keycloak/minio setup

catskillmarina:
This is my sample config for aleph using traefik, keycloak and minio.


sunu:
Thank you so much for putting this together @catskillmarina!

I think we can get rid of the license, dockerignore and gitignore files since they are pretty much the same as the ones in the root of the repo? I can do that clean up after merging this. Thanks again!

Issue #2093, opened 2022-01-01
No workers present on fresh docker launch, 'make worker' errors


straykat914:
Running a fresh development build of Aleph. 
```make all``` pulls and builds with no noticeable errors.  Web UI loads fine.

Uploading data, I encountered the common 'Stuck at 67%' issue.

Attempted to run ```make worker``` and received this traceback:

```
~/aleph (main) > make worker
docker-compose -f docker-compose.dev.yml up -d --remove-orphans \
        postgres elasticsearch ingest-file \
        convert-document
aleph_convert-document_1 is up-to-date
aleph_redis_1 is up-to-date
aleph_elasticsearch_1 is up-to-date
aleph_postgres_1 is up-to-date
aleph_ingest-file_1 is up-to-date
docker-compose -f docker-compose.dev.yml run -p 127.0.0.1:5679:5679 --rm app python3 -m debugpy --listen 0.0.0.0:5679 /usr/local/bin/aleph worker
Creating aleph_app_run ... done
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/dist-packages/debugpy/__main__.py", line 45, in <module>
    cli.main()
  File "/usr/local/lib/python3.8/dist-packages/debugpy/server/cli.py", line 444, in main
    run()
  File "/usr/local/lib/python3.8/dist-packages/debugpy/server/cli.py", line 285, in run_file
    runpy.run_path(target_as_str, run_name=compat.force_str("__main__"))
  File "/usr/lib/python3.8/runpy.py", line 264, in run_path
    code, fname = _get_code_from_file(run_name, path_name)
  File "/usr/lib/python3.8/runpy.py", line 234, in _get_code_from_file
    with io.open_code(decoded_path) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/bin/aleph'
ERROR: 1
make: *** [Makefile:30: worker] Error 1
```

urchnk:
@straykat914 I've got the same error. Could you tell me how you resolved it?

sunu:
@urchnk Alternatively you can try running `make shell` and run `aleph worker` in the shell you get. Let me know if that works for you

urchnk:
@sunu It helped, thanks a lot!

PR #2317, opened 2022-06-08
Fix worker target in Makefile

tillprochaska:
Currently, running `make worker` fails as the aleph CLI is installed at `/usr/bin/aleph` (not `/usr/local/bin/aleph`). This has previously been reported in #2093.

I’m not really sure though what has changed since the commit that introduced the absolute path to `/usr/local/bin/aleph` (https://github.com/alephdata/aleph/commit/f2824360c7032e98bb9944710feb978641a93181). I quickly looked through the history of files that might be relevant (setup.py, Dockerfile + compose config, …), but didn’t find anything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ web: services
 	$(COMPOSE) up api ui
 
 worker: services
-	$(COMPOSE) run -p 127.0.0.1:5679:5679 --rm app python3 -m debugpy --listen 0.0.0.0:5679 /usr/local/bin/aleph worker
+	$(COMPOSE) run -p 127.0.0.1:5679:5679 --rm app python3 -m debugpy --listen 0.0.0.0:5679 -c "from aleph.manage import cli; cli()" worker
 
 tail:
 	$(COMPOSE) logs -f


sunu:
I'm not sure what's causing the difference but I have aleph installed at `/usr/local/bin/aleph`:

image

tillprochaska:
Just curious, do you get that as well when you start a container using an image built in CI?

```
docker run -it alephdata/aleph:3.12.3 which aleph
```

sunu:
That gives me `/usr/bin/aleph`. My first guess was that it's caused by differences between arm64 vs amd64 builds. But doesn't seem like that's true.

tillprochaska:
Yes, I thought of that as well and had also tried this out on an Intel, but with the same result. I also asked Steve to try this out, as he’s the one who committed the absolute path initially, but it’s `/usr/bin/aleph` for him as well.

Maybe using shell interpolation to run `which aleph` instead of using a static absolute path is the solution here? Although it would sure look a little ugly/be less readable.

tillprochaska:
@Rosencrantz Thanks for approving - I’ve converted this back to a draft though. While this change would fix the issue on my machine, it would break the target for sunu (and most likely others).

tillprochaska:
@sunu, I’ve considered a couple of different fixes for this seemingly easy-to-fix problem now:

1. Use shell interpolation to do something like `docker compose run app bash -c "$(which aleph)"`. Honestly, I just gave up on this because it involves shell interpolation on multiple levels, because I didn’t manage to get it right and it was absolutely unreadable.

2. The fix that’s currently committed. It’s still a workaround, but it doesn’t rely on the absolute path and is fairly easy to understand.

3. Probably there would be some way to specify the install location for console scripts, but I was afraid of potential unexpected side effects.

sunu (approved):
I think that's a reasonable workaround for now

PR #2322, opened 2022-06-12
Add LinkMenuItem, LinkButton to replace inaccessible workarounds

tillprochaska:
Unfortunately, Blueprint makes it quiet hard to render valid, accessible HTML markup with its `AnchorButton` and `MenuItem` components. Currently, we’ve been using different workarounds (e.g. using `onclick` handlers rather than links or nesting buttons withink anchor tags), which e.g. makes it impossible to copy the URL or open a link in a new tab, and breaks assistive technology.

Fortunately, with recent versions of React Router, it’s become easier to do this in a way that doesn’t produce inaccessible markup: The `useLinkClickHandler` and `useHref` tags provide the behavior needed to support client-side navigation with existing components like `AnchorButton` and `MenuItem`.

The `LinkMenuItem`/`LinkButton` components behave like React Router’s `Link` component and render like their `MenuItem`/`AnchorButton` Blueprint counterparts.

I’ve started using these two components in the dashboard sidebar menu, the pages sidebar menu and the top navbar menu. There are a few more locations where we currently make use of workarounds, but I’d prefer to keep this PR rather small.

https://user-images.githubusercontent.com/1512805/173228953-e93fc22a-ecc7-432d-9cad-1975b9bf6cb9.mov

### DIFF
--- a/ui/src/components/Dashboard/Dashboard.jsx
+++ b/ui/src/components/Dashboard/Dashboard.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { Classes, Menu, MenuItem, MenuDivider } from '@blueprintjs/core';
+import { Classes, Menu, MenuDivider } from '@blueprintjs/core';
 import { defineMessages, injectIntl, FormattedMessage } from 'react-intl';
-import { Count, Skeleton, AppItem } from 'components/common';
+import { Count, Skeleton, AppItem, LinkMenuItem } from 'components/common';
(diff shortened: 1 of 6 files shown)


Rosencrantz (commented):
My question would be, is there a specific reason why things were done in this way to start with?

sunu (approved):
This looks great! "Open link in new tab" is such a time saver.

On a related note, can we run automated test on Aleph to find out more accessibility issues like these? IIRC there was one called lighthouse.

Issue #2327, opened 2022-06-13
BUG: xref page can result in an (almost) infinite load situation


Rosencrantz:
**Describe the bug**
When I load the XREF page https://aleph.occrp.org/datasets/5707#mode=xref Aleph keeps sending API calls to get the next page and never stops requesting the subsequent page until all pages have been requested.

Unfortunately although the response suggests that there are a large number of results (10,000) no results are actually returned as a result of the XHR requests being made.

**To Reproduce**
1. Go to https://aleph.occrp.org/datasets/5707#mode=xref
2. See that no results are displayed
3. Open the developer toolbar and the network tab
4. Notice that there are ongoing requests being made, but that they have no results

**Expected behavior**
Either results should be returned and displayed or no results are returned and the requests should stop

**Additional context**
This does not seem to occur with all datasets.

Rosencrantz:
Closed as Stale. This issue has changed/evolved.

PR #2361, opened 2022-06-29
Raise a 500 error when we're expecting results but the results array is empty

sunu:
ref: https://github.com/alephdata/aleph/issues/2327 and https://github.com/alephdata/aleph/pull/2328

### DIFF
--- a/aleph/views/serializers.py
+++ b/aleph/views/serializers.py
@@ -1,6 +1,7 @@
 import logging
 from pprint import pprint, pformat  # noqa
 from flask import request
+from flask_babel import gettext
 from pantomime.types import PDF, CSV
 from banal import ensure_list
 from followthemoney import model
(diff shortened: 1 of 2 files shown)


aleph/views/serializers.py
```diff
@@ -88,6 +89,13 @@ def jsonify_result(cls, result, extra=None, **kwargs):
         data = result.to_dict(serializer=cls)
         if extra is not None:
             data.update(extra)
+        if data.get("total") > 0 and not data.get("results"):
+            log.exception(f"Expected more results in the response: {data}")
+            data = {
+                "status": "error",
+                "errors": gettext("Failed to load expected results."),
```
tillprochaska, line 96: I was planning to update our error handling in the frontend to always display the error message we got from the API, so in that case it would only display "Failed to load expected results.". Can we update the message to make it clearer that there should be results, but we haven’t been able to load them due to a technical problem?

sunu: In that case, what do you think about something like: "We found x results, but could not load them due to a technical problem. Please check back later and if the problem persists contact an Aleph administrator"?

tillprochaska (approved):
Tested it while working on the respective frontend changes and it works as expected! :)

aleph/views/serializers.py
sunu: Yes, totally!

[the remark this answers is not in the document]
sunu: I propose we include `message` in all error responses and use `errors` to pass on validation errors only. I guess that's the way it was intended but I was misled by the [example](https://github.com/alephdata/aleph/blob/bf63ad80b7a7f2b593f9ebfab610645d87f8ad80/aleph/views/base_api.py#L244) I copied.

I'll make a separate PR for that.

PR #2367, opened 2022-06-30
Display api error messages

tillprochaska:
Currently, the `ErrorSection` component displays a generic error message (e.g. "Request failed with status code 500" even if the response contains a more specific message that provides helpful information to the end-user (see #2361). I’ve updated the component to display the generic message only as a fallback if the response doesn’t contain an error message.


Rosencrantz:
@tillprochaska Can. you provide us with some steps to reproduce this locally?

tillprochaska:
> Can. you provide us with some steps to reproduce this locally?

Sure! One way to test this is following the steps outlined in https://github.com/alephdata/aleph/pull/2328#issue-1269334449 to intercept API responses and modify them to return error messages. You need to adjust the response code to be erroneous (e.g. 500) and the response body:

```json
{
  "status": "error",
  "message": "This error message will be displayed on the frontend."
}
```

Another (possibly slightly easier) way is to adjust the backend code to trigger error responses:

### XREF

1. After [this line](https://github.com/alephdata/aleph/blob/main/aleph/views/xref_api.py#L62) in `xref_api.py` add the following:
   
   ```py
   result.total = 10
   result.results = []
   ```
2. Navigate to the XREF page in an investigation.
3. You should see an error message.

### Entity Sets

1. After [this line](https://github.com/alephdata/aleph/blob/main/aleph/views/entitysets_api.py#L81) in `entitysets_api.py` add the following:

   ```py
   result.total = 10
   result.results = []
   ```
2. Navigate to the "Network diagrams", "Timelines", or "Lists" section in an investigation.
3. You should see an error message.

### Notifications

1. After [this line](https://github.com/alephdata/aleph/blob/main/aleph/views/notifications_api.py#L36) in `notifications_api.py` add the following:

   ```py
   result.total = 10
   result.results = []
   ```
2. Navigate to the Aleph dashboard/home page.
3. You should see an error message.

tillprochaska:
Additional note: This PR concerns compoennts that already handle API errors using the `ErrorSection` component.  There are some views/components that do not handle erroneous API responses at all (e.g. the alerts index) and will simply render an empty list. I would like to change that, ideally in a generic, consistent way. But that’s probably something for a separate PR.

Rosencrantz:
Thanks for the heads up. Looks to be working. @tillprochaska Can you fix up the merge conflicts?

Rosencrantz (approved):

tillprochaska:
@Rosencrantz Done!

tillprochaska:
Hang on, need to fix the formatting issues flagged by Prettier 🙃

PR #2369, opened 2022-06-30
Navigate to entites referenced in entity tables

### LINKED ISSUE #2338: BUG: Quick links to associated entities in entity tables/table editor
In entity tables, it is already possible to navigate to associated entities. For example, when creating new `Airplane` entities, users can navigate to the owner entity by clicking on it.

Screenshot 2022-06-17 at 11 41 11

This has been explicitly disabled for entities that are not a `Thing`. For example, it’s not possible to navigate to the payer/beneficiary entity of a `Payment`:

Screenshot 2022-06-17 at 11 45 43

The current implementation explicitly disables this feature for entities that do not inherit from `Thing`, i.e. there’s probably a reason we don’t do that right now. It has been requested in the community slack to make associated entities clickable for all entities.

**Describe alternatives you've considered**
-/-

**Additional context**
* [Slack conversation](https://alephdata.slack.com/archives/C01AEDV8UF3/p1655394289487809)
* [Relevant lint in `EntityTable` that disables/enables quick navigation](https://github.com/alephdata/aleph/blob/main/ui/src/components/EntityTable/EntityTable.jsx#L186)

tillprochaska:
Previously, we only linked to referenced entities in case the *referencing* entity is a thing.

As discussed in #2338, we actually considered this a bug. The original intent of this condition was probably to check if the *referenced* entity is a thing. Non-thingy entity views do not offer a lot of additional information and are mostly blank, so it doesn’t make a lot of sense to link to them.

That doesn’t seem something that’s possible right now with FtM – referenced entities are always things (see https://github.com/alephdata/aleph/issues/2338#issuecomment-1170420470), so I devided to simply render links to referenced entities in all cases.

Closes #2338 

Screenshot 2022-06-30 at 13 20 07

### DIFF
--- a/ui/src/components/EntityTable/EntityTable.jsx
+++ b/ui/src/components/EntityTable/EntityTable.jsx
@@ -183,7 +183,7 @@ export class EntityTable extends Component {
   render() {
     const { collection, entityManager, query, intl, result, schema, isEntitySet, sort, updateStatus, writeable } = this.props;
     const { selection } = this.state;
-    const visitEntity = schema.isThing() ? this.onEntityClick : undefined;
+    const visitEntity = this.onEntityClick;
     const showEmptyComponent = result.total === 0 && query.hasQuery();
     const selectedEntities = selection.map(this.getEntity).filter(e => e !== undefined);
 


tillprochaska:
I think I’ve now understood why this condition was implemented. The entity tables do not only contain links to referenced entities, they also link to the current entity in the table header:

Screenshot 2022-06-30 at 13 24 39

In the case of non-thing entities, this will link to a page that is mostly empty. However, it does provide some additional information and it doesn’t result in an error, so I think this would actually be fine.

Screenshot 2022-06-30 at 13 27 11

tillprochaska:
I have also thought about opening referenced entities in a preview drawer so users don’t lose context while viewing more information. This is a pattern we already use in multiple other places.

https://user-images.githubusercontent.com/1512805/176669777-94a5715b-1a04-42b4-862d-f8e67d4fe294.mov

This is easy to implement. However, the entity table currently renders that small expand icon next to the link. In all other places, this icon indicates that clicking it will take you to the full entity page (rather than the preview), so we’d have to remove/change that before implementing it. As the table is part of `react-ftm`, that would require making this change there first, then bumping the `react-ftm` version in Aleph etc., so I’ll take a look at this separately.

Rosencrantz:
I think that this user experience needs a re-think. I don't equate that expand with showing me more information on the current entity. And the page, once it is show, is mostly a waste of space.

I agree that we should have the entity details displayed in a better way than they are currently. But perhaps this is something that we could raise as a feature request and deal with seperately? Unless it's a really quick fix to sort out a more sensible UX?

tillprochaska:
> I agree that we should have the entity details displayed in a better way than they are currently. But perhaps this is something that we could raise as a feature request and deal with seperately? Unless it's a really quick fix to sort out a more sensible UX?

I agree that it could use some overall improvements. I’m not sure if your comment ware a reply to https://github.com/alephdata/aleph/pull/2369#issuecomment-1171123409 or to the PR in general? The comment was really just an idea, nothing I'd want to implement in this PR.

The ability to navigate to a referenced entity is already implemented. However, it doesn’t work for some entity types. Consider the following example. When viewing vehicle entities, users are able to click on the referenced person entity. When viewing associate entities, users aren’t able to click on the referenced entities -- even though they reference the exact same entity type.

https://user-images.githubusercontent.com/1512805/176684609-70c6927d-d802-481e-a611-12883e4e6925.mov

Rosencrantz:
@tillprochaska Just a reply, not in relation to the pull request.

Rosencrantz (approved):

PR #2378, opened 2022-07-04
Update to `react-scripts@5.0.1`

tillprochaska:
The only relevant breaking change in `react-scripts` v5 seems to be the fact that Webpack 5 doesn’t ship with polyfills for built-in Node.js modules anymore. We’re using some of them (transitively), but as far as I can seem, installing the polyfills manually seems to be enough.


ui/package.json
```diff
@@ -21,7 +21,9 @@
     "@types/react": "^17.0.1",
     "@types/react-dom": "^17.0.1",
     "axios": "^0.25.0",
+    "buffer": "^6.0.3",
     "classnames": "^2.2.6",
+    "crypto": "npm:crypto-browserify@^3.12.0",
```
tillprochaska, line 26: This is just an alias syntax, installing the `crypto-browserify` module from `npm` as `crypto` (the name of the built-in node module it’s polyfilling).

Rosencrantz (approved):

Issue #2370, opened 2022-06-30
BUG: 500 response if limit is set to zero


tillprochaska:
**Describe the bug**
#2361 implemented a change that returns an error response if the total number of results is larger than zero, but the list of results is empty.

There is however a legitimate situation where such a response is expected. When an API consumer explicitly sets `limit=0`. Aleph’s PDF viewer [does that](https://cs.github.com/alephdata/aleph/blob/bf63ad80b7a7f2b593f9ebfab610645d87f8ad80/ui/src/viewers/PdfViewer.jsx#L255) in order to get the total number of pages without actually requesting any pages.

**To Reproduce**
1. Create an investigation.
2. Upload a document.
3. Click on "Documents" in the sidebar.
4. Open the network tab in browser dev tools-
5. Click on the document. You should see an infinite loading spinner.
6. You should be able to see a request to the `entities` endpoint with `limit=0`.

**Expected behavior**
The API should not return an error response in this case as an empty list of results is actually expected.

tillprochaska:
I can think of another similar situation: If a query has 100 results and an API consumer sends a request with e.g. `offset=200`, that would also return an empty list of results.

sunu:
So we have 2 legitimate cases where empty results should be allowed with a non-zero total - when `limit==0` and when `offset>=total`. I can introduce those checks.

tillprochaska:
@sunu, yes, there might be more cases though (but I couldn’t think of any other so far).

Rosencrantz:
@Rosencrantz Can you verify whether this issue is now resolved

PR #2380, opened 2022-07-05
Don't return a 500 response in cases where empty result is expected

sunu:
fixes https://github.com/alephdata/aleph/issues/2370

### DIFF
--- a/aleph/views/serializers.py
+++ b/aleph/views/serializers.py
@@ -89,18 +89,24 @@ class Serializer(object):
         data = result.to_dict(serializer=cls)
         if extra is not None:
             data.update(extra)
+        # Sometimes part of the result might be missing because of some
+        # inconsistency (eg: a failed re-indexing). We try to spot those
+        # inconsistencies.
         total = data.get("total", 0)
+        limit = data.get("limit", 0)
+        offset = data.get("offset", 0)
         if total > 0 and not data.get("results"):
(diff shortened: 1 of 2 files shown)


aleph/views/serializers.py
```diff
@@ -89,18 +89,24 @@ def jsonify_result(cls, result, extra=None, **kwargs):
...
-            }
-            return jsonify(data, status=500)
+            if not (limit == 0 or offset >= total):
+                log.exception(f"Expected more results in the response: {data}")
+                data = {
+                    "status": "error",
+                    "message": gettext(
+                        f"We found {total} results, but could not load them due "
+                        "to a technical problem. Please check back later and if "
+                        "the problem persists contact an Aleph administrator"
```
tillprochaska, line 106: This isn’t directly related to this PR, but I noticed that we need to use `gettext` placeholders here. Otherwise this message can’t be extracted correctly.

I can also open a separate PR if you’d prefer that!

```suggestion
                        "We found %(total)d results, but could not load them due "
                        "to a technical problem. Please check back later and if "
                        "the problem persists contact an Aleph administrator"
```

tillprochaska (approved):
Looks good, thanks! 👍

aleph/views/serializers.py
sunu: Ah, right! makes sense. Thanks for spotting it!

PR #2382, opened 2022-07-05
Use gettext placeholder

tillprochaska:
@sunu Sorry, that’s another thing I wasn’t aware of when I reviewed that PR. 🙃 We can’t use format strings for messages to be translated and need to use gettext placeholders instead.

See https://github.com/alephdata/aleph/pull/2380#discussion_r913464101


sunu (approved):
Thanks for the quick fix @tillprochaska! I should have tested my original change with a different language  😄

PR #2388, opened 2022-07-07
Display highlights in document plaintext view mode

tillprochaska:
This PR adds support to highlight search terms in document text while in plaintext mode. It would be nice to support this in view mode as well, but this is likely going to be less straight-forward and something I’d look into separately.

https://user-images.githubusercontent.com/1512805/178830658-b1071061-1523-4d77-8de3-04d7b0f7b575.mov

Example document that can be used for testing: [test_markup.pdf](https://github.com/alephdata/aleph/files/9071458/test_markup.pdf)

- [x] ~If the document text contains line breaks, and a snippet spans multiple lines, then the snippet does not contain line breaks.~ Actually, this is not an issue at all. Snippets returned by ES do contain line breaks.
- [x] Adjust the number (and length) of highlights/snippets?
- [x] Test escpaing in `TextViewer`
- [x] Test `SearchHighlight`
- [x] Remove unused imports/linting issues

### DIFF
--- a/aleph/search/parser.py
+++ b/aleph/search/parser.py
@@ -132,10 +132,12 @@ class SearchQueryParser(QueryParser):
         # index.
         self.facet_names = set(self.getlist("facet"))
 
+        # Query to use for highlighting, defaults to the search query
+        self.highlight_text = self.get("highlight_text", self.text)
         # Include highlighted fragments of matching text in the result.
         self.highlight = self.getbool("highlight", False)
         self.highlight = self.highlight and settings.RESULT_HIGHLIGHT
-        self.highlight = self.highlight and self.text
+        self.highlight = self.highlight and self.highlight_text
         # Length of each snippet in characters
         self.highlight_length = self.getint("highlight_length", 120)
         # Number of snippets per document, 0 = return full document text.
--- a/ui/src/util/convertHighlightsToReactElements.js
+++ b/ui/src/util/convertHighlightsToReactElements.js
@@ -1,4 +1,5 @@
 import { createElement } from 'react';
+import { decode } from './htmlEntities';
 
 // Converts a string containing `<em>` tags to a list of
 // React elements without using `dangerouslySetInnerHTML`.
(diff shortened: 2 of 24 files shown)


aleph/search/parser.py
```diff
@@ -132,10 +132,12 @@ def __init__(self, args, authz, limit=None):
         # index.
         self.facet_names = set(self.getlist("facet"))
 
+        # Query to use for highlighting, defaults to the search query
+        self.highlight_text = self.get("highlight_text", self.text)
         # Include highlighted fragments of matching text in the result.
         self.highlight = self.getbool("highlight", False)
         self.highlight = self.highlight and settings.RESULT_HIGHLIGHT
-        self.highlight = self.highlight and self.text
+        self.highlight = self.highlight and self.highlight_text
```
sunu, line 140: So we basically want to be able to highlight terms even when a query is not defined. I think I didn't fully understand that during our discussion the other day.
This looks great @tillprochaska!

ui/src/viewers/PdfViewer.jsx
```diff
@@ -291,6 +291,7 @@ const mapStateToProps = (state, ownProps) => {
   const pageQuery = baseQuery
     .set('highlight', true)
     .set('highlight_text', queryText)
+    .set('highlight_count', 15)
```
tillprochaska, line 294: @sunu Do you think this can have a significant impact on ES performance? Highlights need to be generated based on a single page only and 15 fragments is not a huge number, but I’m not sure how expensive highlighting is.

ui/src/util/convertHighlightsToReactElements.js
```diff
@@ -0,0 +1,40 @@
+import { createElement } from 'react';
+import { decode } from './htmlEntities';
+
+// Converts a string containing `<em>` tags to a list of
+// React elements without using `dangerouslySetInnerHTML`.
+// This ensures that all other HTML markup is automatically
+// escaped by React.
```
tillprochaska, line 7: Before, we simply rendered the text contents of a document as a text node and React would take care of escaping any HTML markup. Now, our text contents will contain some HTML markup (the `em` tags for highlighting search terms) that we do not want to be escaped. Instead of using something like `dangerouslySetInnerHTML`, this converts the text into a list of React elements (for the `em` tags) and basic text nodes. This ensures that React still escapes any HTML markup contained in the original document text and only renders the `em` tags as actual highlights.

As this is a somewhat sensitive change, please make sure to review this thoroughly, just in case I missed something. Here’s a [PDF document](https://github.com/alephdata/aleph/files/9071458/test_markup.pdf) I used during development that might be helpful.

ui/src/viewers/PdfViewer.jsx
Rosencrantz: @sunu What are your thoughts on this?

Rosencrantz:
I am happy with the changes here. Thanks for taking the time to consider both security and performance. My only concern is that I struggled to get this running locally. This may be the poor internet connection I am currently on, so am going to try again (later).

Rosencrantz (approved):

PR #2398, opened 2022-07-12
Show message that public access is disabled

sunu:
On the home page, show a message that public access is currently disabled.

### DIFF
--- a/aleph/pages/home.en.md
+++ b/aleph/pages/home.en.md
@@ -4,8 +4,6 @@ title: 'Aleph'
 description: 'Search documents, companies and people'
 html: true
 samples: ["Vladimir Putin", "TeliaSonera"]
-warning_title: "Public access temporarily disabled"
-warning_body: "We have disabled public access to prevent Aleph from being overwhelmed by automated anonymous requests."
 ---
 
 <section className="HomeScreen__section">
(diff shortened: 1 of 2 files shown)


sunu:
Editing the markdown files to add this message feels a bit weird though. May be putting it in settings is a better move?

Also, looks like the access disabled message on the home screen (https://github.com/alephdata/aleph/blob/main/ui/src/screens/HomeScreen/HomeScreen.jsx#L27) is not being used.

sunu:
This message will show up like this fwiw
image

tillprochaska:
> Editing the markdown files to add this message feels a bit weird though. May be putting it in settings is a better move?

Yes, it’s actually planned to start working on something like that soon!

aleph/pages/home.en.md
```diff
@@ -4,6 +4,8 @@ title: 'Aleph'
 description: 'Search documents, companies and people'
 html: true
 samples: ["Vladimir Putin", "TeliaSonera"]
+warning_title: "Public access temporarily disabled"
+warning_body: "We have disabled public access to prevent Aleph from being overwhelmed by automated anonymous requests."
```
tillprochaska, line 8: Is this the correct file to edit? I always thought that aleph.occrp.org will use the site-specific contents from in [`/site/aleph.occrp.org/pages/home.en.md`](https://github.com/alephdata/aleph/blob/main/site/aleph.occrp.org/pages/home.en.md).

Rosencrantz: I'd add a date here for context, it can be frustrating not knowing when this started

sunu: Ok, I have updated the PR. Thanks for catching the mix up @tillprochaska!

Rosencrantz (approved):

PR #2407, opened 2022-07-18
Setup a RC release process

Rosencrantz:


sunu:
Looks good to me from a quick glance. I'll also give it a spin locally to test the flow and let you know.

What do you think about documenting the release process and flow here in the repo? We do have some documentation in the wiki but it's probably useful to put it next to the code too.

Rosencrantz:
@sunu I like the idea of adding release documentation here. I'll review what we've got in the wiki and add it to the README.

README.rst
```diff
@@ -25,8 +25,56 @@ Support
...
+
+As far as possible apply the rules of semantic versioning when determining the type of release to perform.
+
+### Technical process
+
+#### RC releases
+
+If you need to perform an RC release of Aleph, follow these steps:
+
+1. Ensure that the `CHANGELOG`` is up to date on the develop branch and that all outstanding PR's have been merged
```
sunu, line 63: mismatched quotes here

sunu (approved):

Issue #1500, opened 2020-12-03
Result highlights aren't being shown 


pudo:
Especially in page searches within a`Pages` entity.

kjacks:
I think this might be a backend issue? I'm noticing that even with the highlight query params set in the entities query, the API is returning a result without a `highlight` key on the Page entity.

![Uploading Screen Shot 2020-12-07 at 12.54.22.png…]()

pudo:
This now tries to highlight within `properties.*` and is more permissive by disabling analyser matches via `require_field_match`.

PR #2416, opened 2022-07-25
Monneyboi/search highlights

etymono:
This is supposed to fix search highlights. It's basically a kind of rollback of https://github.com/alephdata/aleph/issues/1500. I can't find the reason to use the `properties.*` and `require_field_match`, but i could overlook something.

There's a couple of things i'm unsure about:
1. `"term_vector": "with_positions_offsets"` - I think this was left after removing the fast vector highlighting? It's there in the collection index as well, why?
2. moving the `"search_quote_analyzer": "latin_index"` to `LATIN_TEXT`, i just couldn't see a reason why not using it on fingerprints aswell.

Please let me know if i've overlooked something :)


Rosencrantz:
@tillprochaska How does this compare to the work that you've been doing on highlights? Is it duplication, complimentary or something else? @sunu Would be good to understand the performance impact of these changes, if any.

monneyboi:
We're trading index size versus highlighting speed here it seems:
```
Term vectors. If term_vector information is provided by setting term_vector to with_positions_offsets in the mapping, the unified highlighter automatically uses the term_vector to highlight the field. It's fast especially for large fields (> 1MB) and for highlighting multi-term queries like prefix or wildcard because it can access the dictionary of terms for each document. The fvh highlighter always uses term vectors. 
```

https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html#offsets-strategy

Rosencrantz:
Closing because inactive.

PR #2459, opened 2022-08-14
Make the hide filter button persistent

Rosencrantz:
It's a really small change which makes the toggle filters sidebar button persistent within Aleph. Currently unless you are hovering over the very find line you would not know that this thing exists.


tillprochaska (approved):
Until now, I wasn't even aware that the sidebar is collapsible! I’ve pushed one minor change to add a label to the toggle button.

Issue #1940, opened 2021-08-25
Check npm audit and resolve any security vulnrabilities


Rosencrantz:

sam-heller:
Heyo,

PR For this guy is a far larger one than I expected it to be.  At a point right now where I'm potentially making major version upgrade choices, and my mother always taught me to not make major design decisions about a codebase for at least the first 3 months. 

Would very much appreciate any 👀 and direction anyone on the team could spare. 

I'm Sam btw!

Rosencrantz:
Hey @sam-heller. First off, very sorry for the slow slow response. Are you able to create a PR for the work that you have done thus far? Then we can take a look and perhaps collaborate on making some forward momentum?

tillprochaska:
@Rosencrantz Check out the conversation in the linked PR! https://github.com/alephdata/aleph/pull/2469

Rosencrantz:
Thanks till, apologies for the noise on this issue.

PR #2469, opened 2022-08-16
Security Updates for https://github.com/alephdata/aleph/issues/1940

sam-heller:
Various changes to deal with security vulnerabilities.
-Add yarn.lock for version pinning
-css-select pinned to ^4.1.3
-nth-check pinned to 2.1.1
-node-sass swapped out for sass
-scss-tokenizer pinned to 0.4.3


sam-heller:
Hey Folks! 

So I snagged the [NPM Audit ticket](https://github.com/alephdata/aleph/issues/1940) thinking it'd be a straightforward package upgrade and a great introduction ticket, and now I'm bogged down in sass parser wars and definitely would appreciate a call from ya'll on whether you want to push through and upgrade to blueprint 4.x  core or try to hack around the sides for now. 

I'm Sam by the way, nice to meet you and sorry my intro is so much butchery.

ui/src/screens/SystemStatusScreen/SystemStatusScreen.scss
```diff
@@ -1,15 +1,12 @@
...
+// .StatusTable {
+//   @extend .data-table;
+//   tr td,
+//   tr th {
+//     vertical-align: middle;
+//   }
+//   td.text {
+//     @include rtlSupportInvertedProp(padding, left, 0, null);
+//   }
+// }
```
sam-heller, line 12: Wasn't able to track down what in this guy was breaking things, so I commented it out for now in the interest of getting this up with eyes on it. Not staying this way.

tillprochaska:
> So I snagged the https://github.com/alephdata/aleph/issues/1940 thinking it'd be a straightforward package upgrade and a great introduction ticket, and now I'm bogged down in sass parser wars and definitely would appreciate a call from ya'll on whether you want to push through and upgrade to blueprint 4.x core or try to hack around the sides for now.

@sam-heller Hi Sam, welcome to the Aleph community and thanks for opening your first pull request in this repository, that’s really appreciated!

As you’ve noticed, upgrading some of the dependencies (e.g. Blueprint) unfortunately isn’t quite as easy as it seems at first glance. Also, we have a companion library, [react-ftm](https://github.com/alephdata/react-ftm), that is used in Aleph and for example also has a peer dependency on Blueprint.

We are currently in the process of moving react-ftm back into the main Aleph repository and are also upgrading Blueprint and some of our build tooling to the most recent versions in the same process.

That ticket probably should not have been labeled as "good first issue" due to the complexities involved in upgrading some of the dependencies, I’m sorry about that. If there’s another issue you’d like to work on, feel free to reach out so we can make sure it’s up to date and actually a good first issue!

sam-heller:
Roger that, glad I checked in before I got any further down this rabbithole!

Issue #2489, opened 2022-08-31
FEATURE: Return dates with timezone designators in API responses


tillprochaska:
**Is your feature request related to a problem? Please describe.**
The API currently formats dates as ISO strings *without* timezone designators, for example

```
2022-08-31T11:00:50.614817
```

This makes it slightly more difficult to parse them in the frontend, which has in the past led to bugs.

**Describe the solution you'd like**
It would be great if dates could be formated as ISO strings *with* timezone designators, for example:

```
2022-08-31T11:00:50.614817Z # or
2022-08-31T11:00:50.614817+00:00
```

**Describe alternatives you've considered**
-/-

**Additional context**
Maybe we could insert something like this in the `JSONEncoder` to ensure that dates are timezone-aware?

```python
if isinstance(obj, (datetime, date)):
    if obj.tzinfo is None:
        obj = obj.replace(tzinfo=timezone.utc)
    return obj.isoformat()
```

https://github.com/alephdata/aleph/blob/main/aleph/util.py#L36

PR #2490, opened 2022-08-31
Fix date helper

### LINKED ISSUE #2488: BUG: Dashboard/Notifications screen is blank in Safari
**Describe the bug**
When opening the dashboard in Safari (and there is at least one notification), the page briefly displays, but then turns blank within seconds.

**To Reproduce**
1. Make sure that there is at least one notification for your user account. (In order to trigger notifications, you can for example create an export of a search.)
2. Open the Dashboard in Safari 15.5.
3. You should quickly see the dashboard in its loading state, then the page turns blank.

**Expected behavior**
The dashboard should be displayed.

**Screenshots**
-/-

**Additional context**
Browser: Safari 15.5
OS: macOS Monterey

This seems to be related to the `FormattedRelativeTime` component from `react-intl` triggering too many rerenders. React throws an exception if a component triggers too many rerenders.

* I wasn’t able to reproduce the bug in 13.11.1
* I was able to reproduce the bug in 13.12.0+

tillprochaska:
Closes #2488. Related to #2489.

The Aleph API returns dates in UTC as ISO strings without timezone designators. Parsing a date in JS (`new Date(dateString)`) uses the local timezone, and there’s no way to explicitly specify a timezone to use.

We currently have a helper function that replaces the local timezone with UTC. It does this by simply appending "UTC" to the string representation of a date in order to override the time zone.

This does however results in a date string that has multiple time zones:

```js
const date = new Date()
console.log(`${date} UTC`);
// "Wed Aug 31 2022 20:01:20 GMT+0200 (Central European Summer Time) UTC"
```

Firefox and Chrome parse these date strings, even though they are invalid. Safari throws an error.

This PR replaces the helper implementation with a way to adjust the timezone using only standard, broadly supported date methods.

### DIFF
--- a/ui/src/util/convertUTCDateToLocalDate.js
+++ b/ui/src/util/convertUTCDateToLocalDate.js
@@ -1,3 +1,8 @@
+// The Aleph API returns dates as ISO strings without timezone
+// designators. When initializing new `Date` objects, those dates
+// will by default be initialized with the local timezone of the
+// user agent.
 export default function convertUTCDateToLocalDate(date) {
-  return new Date(`${date} UTC`);
+  const dateObj = new Date(date);
+  return new Date(dateObj.getTime() - dateObj.getTimezoneOffset() * 60 * 1000);
 }


ui/src/util/convertUTCDateToLocalDate.js
```diff
@@ -1,3 +1,8 @@
+// The Aleph API returns dates as ISO strings without timezone
+// designators. When initializing new `Date` objects, those dates
+// will by default be initialized with the local timezone of the
+// user agent.
 export default function convertUTCDateToLocalDate(date) {
-  return new Date(`${date} UTC`);
+  const dateObj = new Date(date);
+  return new Date(dateObj.getTime() - dateObj.getTimezoneOffset() * 60 * 1000);
```
tillprochaska, line 7: There was a similar implementation before that Kirk changed in https://github.com/alephdata/aleph/commit/f30763ea5924222ec21ac47d95e0592a45c6fb92.

But if I understand correctly, the reason to move away from the old implementation was that it would fail if the day was different in the local vs UTC timezone.

This implementation shouldn’t have this issue, as it uses timestamps.

sunu (approved):
Looks good to me!

Rosencrantz:
@tillprochaska Can you cherry-pick this over to the rollback branch please.

PR #2543, opened 2022-09-29
Handle table rows with incorrect number of columns

### LINKED ISSUE #2377: BUG: Skeleton in table previews displayed after loading has finished
**Describe the bug**
When displaying a table with only a few rows, skeletons are still displayed even after the data has been loaded.

**To Reproduce**
1. Create a new investigation.
2. Upload a table document with less than 4 rows (see example CSV file).
3. Wait until Aleph has processed the document.
2. Go to the "Documents" section in the investigation.
3. Open the table document.
4. Skeletons are displayed in the first 4 rows (if they aren’t replaced by rows of the table document).

**Expected behavior**
After the table data has been loaded, skeletons shouldn’t be displayed anymore.

**Screenshots**

https://user-images.githubusercontent.com/1512805/177121034-e8411af7-42f6-4f9e-9d31-a51e1d52e5df.mov

**Additional context**
* [Example CSV file](https://github.com/alephdata/aleph/files/9038000/example_table.csv) that can be used to reproduce the bug

tillprochaska:
**This PR is based on https://github.com/alephdata/aleph/pull/2478 as it includes changes to components in react-ftm. Merge https://github.com/alephdata/aleph/pull/2478 first, then change the PR target to develop.**

***

When ingesting CSV files (and probably also other tabular file formats), the resulting CSV file that’s stored in the archive contains a new line file ending.

When we request and parse the CSV file in the frontend, the CSV parser we use emits only a single empty cell for the last row. When we were trying to render other cells in that row, the cell value would always be "undefined" and a skeleton loader would be displayed indefinitely.

Instead, we now compare the number of rows that have been already loaded with the row index of the cell to be rendered.

Closes #2377

### DIFF
--- a/ui/src/viewers/TableViewer.jsx
+++ b/ui/src/viewers/TableViewer.jsx
@@ -30,10 +30,14 @@ class TableViewer extends Component {
   }
 
   renderCell(rowIndex, colIndex) {
-    const row = this.props.rows[rowIndex];
+    const { rows } = this.props;
+
+    const row = rows[rowIndex];
     const value = row ? row[colIndex] : undefined;
+    const loading = rowIndex >= rows.length;
+
     return (
-      <Cell loading={value === undefined}>
+      <Cell loading={loading}>
         <TruncatedFormat detectTruncation>{value || ''}</TruncatedFormat>
       </Cell>
     );


Rosencrantz (commented):
There seems to be a lot of file changes here. Is something odd going on with this PR?

tillprochaska:
Yeah, I rebased the target branch after opening this PR. Should be fixed now.

Rosencrantz (approved):

PR #2555, opened 2022-10-06
don't set env var prefix name as flask config setting

brassy-endomorph:
Fixes a bug where we were setting unwanted env vars the settings object

### DIFF
--- a/aleph/settings.py
+++ b/aleph/settings.py
@@ -5,10 +5,15 @@
 # defaults.
 import json
 import os
-from servicelayer import env
-from urllib.parse import urlparse
-from flask_babel import lazy_gettext
 from datetime import timedelta
+from json.decoder import JSONDecodeError
+from urllib.parse import urlparse
+
+import structlog
+from flask_babel import lazy_gettext
+from servicelayer import env
+
+log = structlog.get_logger(__name__)
 
 
 class Settings:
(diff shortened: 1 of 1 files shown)


aleph/settings.py
```diff
@@ -5,10 +5,15 @@
 # defaults.
 import json
 import os
-from servicelayer import env
+from datetime import timedelta
```
Rosencrantz, line 8: Is this being used?

aleph/settings.py
```diff
@@ -5,10 +5,15 @@
...
+from json.decoder import JSONDecodeError
 from urllib.parse import urlparse
+
+import structlog
 from flask_babel import lazy_gettext
-from datetime import timedelta
+from servicelayer import env
```
Rosencrantz, line 14: Is this required?

aleph/settings.py
```diff
@@ -154,7 +159,7 @@ def __init__(self) -> None:
         )
 
         # Publishing network diagram embeds
-        self.REACT_FTM_URL = "https://cdn.jsdelivr.net/npm/@alephdata/react-ftm@latest/dist/react-ftm-embed.js"  # noqa
```
Rosencrantz, line 157: Was this a conscious decision?

aleph/settings.py
brassy-endomorph: Yes, for things like

```python
self.ROLE_INACTIVE = timedelta(days=env.to_int("ALEPH_ROLE_INACTIVE", 6 * 30))
```

aleph/settings.py
brassy-endomorph: No not actually, but that was consistent with the style of the file before I first started editing it

aleph/settings.py
brassy-endomorph: No. `black` must have stripped out the `noqa`. I added it back.

PR #2572, opened 2022-10-11
Bookmarks MVP

tillprochaska:
A very simple, client-side only implementation of a bookmarks feature for Aleph. Allows users to bookmark entities and quickly view a list of all bookmarks anywhere in the app.

**To consider:**

* Bookmarks may not be a perfect/flexible-enough solution for some cases. For example, some users want to mark/tag entities as sighted so they know they don’t have to look at the entities again when it appears e.g. in the results for another search.
* Some users would like to add a list of relevant entities from *other* datasets to an investigation.
* Some users would like to add tags/bookmark entities/mark entities as sighted in a way that can be shared across everyone who works on an investigation.

**To do:**

- [x] Improve empty state wording to let users know what they can bookmark
- [x] Add feature flag/env variable to enable experimental features
- [x] Only show bookmarks if user is authenticated
- [x] Add placeholder API endpoints `POST /bookmarks`, `DELETE /bookmarks/:id` that emit logs containing the user ID.
- [x] Download/backup bookmarks as plaintext file

**Possible future improvements:**

- [ ] Add indicator in search results/entity listings ("in your bookmarks")

### DIFF
--- a/aleph/settings.py
+++ b/aleph/settings.py
@@ -202,6 +202,12 @@ class Settings:
         self.XREF_MODEL = env.get("FTM_COMPARE_MODEL", None)
 
         ###############################################################################
+        # Feature flags
+        self.ENABLE_EXPERIMENTAL_BOOKMARKS_FEATURE = env.get(
+            "ALEPH_ENABLE_EXPERIMENTAL_BOOKMARKS_FEATURE", False
+        )
+
+        ###############################################################################
         # Additional configurations
         string_prefix = env.get("ALEPH_STRING_CONFIG_PREFIX")
         json_prefix = env.get("ALEPH_JSON_CONFIG_PREFIX")
--- a/aleph/views/bookmarks_api.py
+++ b/aleph/views/bookmarks_api.py
@@ -1,0 +1,51 @@
+import logging
+from flask import Blueprint, request
+from aleph.views.util import jsonify, require
+
+
+log = logging.getLogger(__name__)
+blueprint = Blueprint("bookmarks_api", __name__)
+
+
+@blueprint.route("/api/2/bookmarks", methods=["POST"])
+def create():
+    """Creates a new bookmark for the user. This is currently a stub endpoint,
+    as the current version of the bookmarks features stores bookmarks client side.
+    ---
+    post:
+      summary: Create bookmark
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                type: object
+          description: OK
+      tags:
+        - Bookmarks
+    """
+    require(request.authz.logged_in)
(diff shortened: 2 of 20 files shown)


aleph/views/bookmarks_api.py
```diff
@@ -0,0 +1,51 @@
...
+            application/json:
+              schema:
+                type: object
+          description: OK
+      tags:
+        - Bookmarks
+    """
+    require(request.authz.logged_in)
+    log.info("User [%s]: Created bookmark", request.authz.role.id)
+    return jsonify({}, 200)
```
stchris, line 29: Minor thing, but we could consider return a `201` here. Likely we also want to return the `bookmark_id`. And perhaps the URL to that bookmark (and/or set the `Location` header to it).

(not sure how (in)consistent that is with other endpoints)

stchris (commented):
Left some minor remarks on the backend stub implementation. Happy to discuss and obviously I know very little about how other endpoints work, but it seems important to get the API right now.

aleph/settings.py
```diff
@@ -201,6 +201,12 @@ def __init__(self) -> None:
         # XREF Model Selection
         self.XREF_MODEL = env.get("FTM_COMPARE_MODEL", None)
 
+        ###############################################################################
+        # Feature flags
+        self.ENABLE_EXPERIMENTAL_BOOKMARKS_FEATURE = env.get(
```
Rosencrantz, line 206: Nitpick. The '_FEATURE' seems superfluous to me. "ENABLE_EXPERIMENTAL_BOOKMARKS" should be enought right?

Rosencrantz (approved):

aleph/views/bookmarks_api.py
tillprochaska: Thanks Chris!

> Minor thing, but we could consider return a 201 here.

Fair enough -- I think most of the other endpoints (e.g. entity creation) return `200`, but I agree that the better option here would be `201`.

> but it seems important to get the API right now.

I think we can change the API later on. To be honest I’m not really confident enough to commit on a stable API at the moment, especially given that the feature may change significantly or may even be removed completely based on user feedback. It’s an experimental feature though and has to be enabled explicitly, so maybe that’s ok?

aleph/settings.py
tillprochaska: Makes sense, I’ll push an update! :)

aleph/views/bookmarks_api.py
stchris: Totally fine. In that context feel free to skip the `201` change as well and check it in as it is, from my point of view 👍

stchris (approved):

Rosencrantz (approved):

PR #2711, opened 2022-11-29
Enable json logging

stchris:


Rosencrantz (approved):
Seems very straight forward.

stchris:
@Rosencrantz I mean ... it might work, I'm not sure why it was not enabled before though. But I think it can't hurt too much to give it a go.

PR #2716, opened 2022-11-30
Fix blank screen create folder

tillprochaska:
Before, after creating a folder, we’d redirect to a legacy route (`/documents/123`). This route was removed some time ago in ec9039db when documents were refactored to be plain entities. This PR fixes the redirect to target the correct route (`/entities/123`) instead.

While we had redirects in place for legacy routes, those seem to have been broken with the upgrade from React Router v5 to v6. I’ve also fixed those redirects. Now, if you visit `/documents/123` users will be correctly redirected to `/entities/123`.


Rosencrantz:
Long term... is there a way in which we can remove this. Also, I think maybe your comment above explaining why this exists would be a good thing to put into the LegacyRedirect file. What do you think?

tillprochaska:
> Long term... is there a way in which we can remove this

You mean the redirects from legacy routes to the current routes? We can remove them at any time, but that would potentially break links to Aleph that are out there in Wiki pages etc.

> I think maybe your comment above explaining why this exists would be a good thing to put into the LegacyRedirect file. What do you think?

Rosencrantz:
Not quite what I meant, I was wondering if there was ever a future where this feature would not be needed, and that all legacy routes had been updated

tillprochaska:
@Rosencrantz Ah, sorry misunderstood that. When I fixed the redirect that broke the folder creation flow, I also did a quick search through our code base for any other redirects that still use the legacy routes and couldn’t find any other.

So once this is merged, the redirects shouldn’t be necessary anymore for Aleph to work correctly, but I’d recommend we still keep them around so that deep links into Aleph from other apps are not broken.

I will however based on your feedback extract all of them into a separate file and add comments for the individual redirects to provide context for future reference.

tillprochaska:
@Rosencrantz I extracted the legacy redirects into a separate file so they don’t clutter the main router component.

Rosencrantz (approved):
Splendid

Issue #2742, opened 2022-12-05
FEATURE: Show the deployed FtM version on Aleph about page


uhhhuh:
**Is your feature request related to a problem? Please describe.**
(In)ability to quickly figure out what the currently deployed version of FtM is wrt the available entities etc.

**Describe the solution you'd like**
FtM version number alongside Aleph version on /pages/about

Rosencrantz:
@stchris Fixed this so closing with a 3.13.1 fix version.

PR #2743, opened 2022-12-05
Show FTM version on about page

stchris:
I accidentally implemented  #2742 while looking at how one would do that.

### DIFF
--- a/ui/src/components/common/AppItem.jsx
+++ b/ui/src/components/common/AppItem.jsx
@@ -23,7 +23,7 @@ class AppItem extends PureComponent {
     const ftm_message = (
       <FormattedMessage
         id="footer.aleph"
-        defaultMessage="followthemoney {version}"
+        defaultMessage="FollowTheMoney {version}"
         values={{
           version: app.ftm_version,
         }}
--- a/ui/src/components/common/AppItem.jsx
+++ b/ui/src/components/common/AppItem.jsx
@@ -20,7 +20,7 @@ class AppItem extends PureComponent {
         }}
       />
     );
-    const ftm_message = (
+    const ftmMessage = (
       <FormattedMessage
         id="footer.aleph"
         defaultMessage="FollowTheMoney {version}"
--- a/ui/src/components/common/AppItem.jsx
+++ b/ui/src/components/common/AppItem.jsx
@@ -20,14 +20,29 @@ class AppItem extends PureComponent {
         }}
       />
     );
-    return (
+    const ftm_message = (
+      <FormattedMessage
+        id="footer.aleph"
+        defaultMessage="followthemoney {version}"
+        values={{
+          version: app.ftm_version,
+        }}
+      />
+    );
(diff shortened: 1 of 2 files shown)


ui/src/components/common/AppItem.jsx
```diff
@@ -20,14 +20,29 @@ class AppItem extends PureComponent {
         }}
       />
     );
-    return (
+    const ftm_message = (
+      <FormattedMessage
+        id="footer.aleph"
```
stchris, line 25: This is probably not right, but I wasn't sure if I can freestyle here. would `id="footer.ftm"` work?

ui/src/components/common/AppItem.jsx
```diff
@@ -20,14 +20,29 @@ class AppItem extends PureComponent {
...
       <MenuItem
         className={Classes.TEXT_DISABLED}
         icon="code"
         text={message}
         href="https://docs.alephdata.org"
+      />,
+      <MenuItem
+        className={Classes.TEXT_DISABLED}
+        icon="code"
+        text={ftm_message}
```
tillprochaska, line 42:
```suggestion
        text={ftmMessage}
```

ui/src/components/common/AppItem.jsx
```diff
@@ -20,14 +20,29 @@ class AppItem extends PureComponent {
         }}
       />
     );
-    return (
+    const ftm_message = (
+      <FormattedMessage
+        id="footer.aleph"
+        defaultMessage="followthemoney {version}"
```
tillprochaska, line 26: This is absolutely nitpicking, but I think it would make sense if we tried to get a little more consistent about the casing we use for FtM. So far, we tend to use *FollowTheMoney* in the docs when talking about the data model and/or project as a whole.

```suggestion
        defaultMessage="FollowTheMoney {version}"
```

ui/src/components/common/AppItem.jsx
```diff
@@ -20,14 +20,29 @@ class AppItem extends PureComponent {
         }}
       />
     );
-    return (
+    const ftm_message = (
```
tillprochaska, line 23: I know it’s slightly confusing, but we use camelCase in JS land ;)

```suggestion
    const ftmMessage = (
```

ui/src/components/common/AppItem.jsx
tillprochaska: Was just going to comment about this! This is used as a key in translation files (e.g. [this one](https://github.com/alephdata/aleph/blob/main/ui/i18n/translations/raw/de.json)) (and also by Transifex). You can use any string you want as long as it’s unique, just try to stay consistent. `footer.ftm` should be good!

tillprochaska (approved):
Perfect, this is exactly how I had imagined it to be implemented :)

Rosencrantz (approved):
This is a really nice little change! Thanks for this.

PR #2744, opened 2022-12-05
Enable CORS for Elasticsearch API in development environment

tillprochaska:
This allows us to use tools like dejavue [1] to inspect index contents which is slightly more convenient than sending requests directly to the ES API.

I’ll eventually add this to the technical documentation or the Readme. For now, here’s what’s worked well for me:

1. Run `docker run -p 1358:1358 -d appbaseio/dejavu` to start dejavue
2. Open the web interface at http://localgost:1358
3. Enter the ES endpoint (`http://localhost:19200`) and the name of the ES index. Entity indices follow the follow the pattern `aleph-entity-schema-v1`. For example the index for the `LegalEntity` schema would be `aleph-entity-legalentity-v1`.
4. Click "Connect".

[1] https://github.com/appbaseio/dejavu

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -20,6 +20,10 @@ services:
     environment:
       - discovery.type=single-node
       - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
+      - "http.cors.enabled=true"
+      - "http.cors.allow-origin=*"
+      - "http.cors.allow-headers=*"
+      - "http.cors.allow-credentials=true"
     ports:
       - "127.0.0.1:19200:9200"
     volumes:


stchris:
> I didn’t do that so far because I’m not sure how often it would be used and other developers may have different preferences As long as we don’t start it automatically, it shouldn’t be a problem, right?

Yep, reasonable. Our `Makefile` targets only start specific containers, so it shouldn't hurt ™️ , but I'm fine either way. Having docs for how to set it up is also fine 👍

Rosencrantz (approved):
Good for dev... Maybe not so good for production :)

PR #2789, opened 2023-01-03
Run docs workflow only if docs directory as changes

tillprochaska:
Right now, the docs workflow is triggered every time a commit is pushed. This is unnecessary, can delay execution of other workflows, and can cause notification spam.

The `/docs` directory is completely self-contained (at least at the moment), i.e. it is fine to only run the docs workflow if the diff affects files in this directory.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,8 +1,10 @@
 name: Docs
 
 on: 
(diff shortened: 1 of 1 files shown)


stchris (approved):
LGTM, nice optimization!

PR #2790, opened 2023-01-03
Integrate dejavu as an ES UI into dev workflow

stchris:
Building on https://github.com/alephdata/aleph/commit/33df4c4638caf309eb5e2d45f15adaf9918fd998 this adds `dejavu` to the dev docker-compose setup and briefly documents it.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -29,6 +29,14 @@ services:
     volumes:
       - elasticsearch-data:/usr/share/elasticsearch/data
 
+  dejavu:
+    image: appbaseio/dejavu:3.6.0
+    container_name: dejavu
+    ports:
+        - '1358:1358'
+    links:
+        - elasticsearch
+
   redis:
     image: redis:alpine
     command: ["redis-server", "--save", "3600", "10"]
--- a/docs/src/pages/developers/installation.mdx
+++ b/docs/src/pages/developers/installation.mdx
@@ -152,6 +152,10 @@ You can also build the Aleph images locally. This could be useful while working 
 
 To build the image you can run `make build`, which will build the `alephdata/aleph` image (this will generate a production ready image).
 
+### Inspecting ElasticSearch
+
+At times it can be useful to inspect the contents of the ElasticSearch indices. The developer setup includes [dejavu](https://github.com/appbaseio/dejavu/stargazers), a web-based UI for ElasticSearch. In a local environment brought up as described above you should see dejavu under [http://localhost:1358](http://localhost:1358). You will need to pass it the local ES cluster URL (`http://localhost:19200`) and you can either inspect an individual index or use `*` to see the whole content.
+
 ## Production deployment
 
 <Callout>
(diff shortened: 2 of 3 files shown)


docker-compose.dev.yml
```diff
@@ -29,6 +29,14 @@ services:
     volumes:
       - elasticsearch-data:/usr/share/elasticsearch/data
 
+  dejavu:
+    image: appbaseio/dejavu:3.6.0
+    container_name: dejavu
+    ports:
+        - '1358:1358'
+    links:
+        - elasticsearch
```
tillprochaska, line 38: Is this required? Services can connect to other service names using the service name (i.e. `elasticsearch` in this case) by default. We use `depends_on` for other service to specify service dependencies and control startup order.


```suggestion
    depends_on:
        - elasticsearch
```

docs/src/pages/developers/installation.mdx
```diff
@@ -152,6 +152,10 @@ You can also build the Aleph images locally. This could be useful while working
 
 To build the image you can run `make build`, which will build the `alephdata/aleph` image (this will generate a production ready image).
 
+### Inspecting ElasticSearch
+
+At times it can be useful to inspect the contents of the ElasticSearch indices. The developer setup includes [dejavu](https://github.com/appbaseio/dejavu/stargazers), a web-based UI for ElasticSearch. In a local environment brought up as described above you should see dejavu under [http://localhost:1358](http://localhost:1358). You will need to pass it the local ES cluster URL (`http://localhost:19200`) and you can either inspect an individual index or use `*` to see the whole content.
```
tillprochaska, line 157:
```suggestion
At times it can be useful to inspect the contents of the ElasticSearch indices. The developer setup includes [dejavu](https://github.com/appbaseio/dejavu/stargazers), a web-based UI for ElasticSearch. In a local environment brought up as described above you should see dejavu under [http://localhost:1358](http://localhost:1358).

You will need to pass it the local ES cluster URL (`http://localhost:19200`). You can either inspect an individual index or use `*` to see the whole content. Index names for entity indices follow the pattern `aleph-entity-{{SCHEMA_NAME}}-{{VERSION}}`. For example, the index name for `LegalEntity` entities is `aleph-entity-legalentity-v1`.
```

tillprochaska (approved):
Thanks! 🙌

I’ve added a short paragraph on the naming pattern for entity indices, as I think a common use case is to inspect specific entities or schemata, and I had to look through the source code before to find out how exactly they are named. (For example, last time I needed this I was surprised that camel-case schema names like `LegalEntity` are converted to `legalentity` rather than `legal-entity`.)

docs/src/pages/developers/installation.mdx
tillprochaska: Oh, and this is probably a copy/paste mistake, but I think the GitHub URL to repo should probably the repo home page, not the stargazers page :)

stchris: Ugh it is, good catch. Thanks!

stchris:
> I’ve added a short paragraph on the naming pattern for entity indices, as I think a common use case is to inspect specific entities or schemata, and I had to look through the source code before to find out how exactly they are named. (For example, last time I needed this I was surprised that camel-case schema names like `LegalEntity` are converted to `legalentity` rather than `legal-entity`.)

Makes sense. Thanks for your careful review!

PR #2799, opened 2023-01-06
ingest-file 3.18.0 rc4

stchris:
Bump `ingest-file` to the latest RC to test the fix for https://github.com/alephdata/ingest-file/issues/398


Rosencrantz (approved):
There will also need to be updates to the deployment script, but I can help facilitate that

PR #2808, opened 2023-01-13
Change dependabot pip update check schedule from weekly to monthly

stchris:


Rosencrantz (approved):

Rosencrantz:
Direct to main?

stchris:
Whoops, sorry, I thought we configure dependabot from main, but it’s
probably not the case. Will redo for develop.

PR #2817, opened 2023-01-17
Reset signin modal state after closing or when switching between signin and signup

### LINKED ISSUE #2806: BUG:  Registering blocks a user from Singing In
**Describe the bug**
This bug was reproduced on Aleph running in development mode. 
When a user Registers using an e-mail on the landing page, they can no longer access the username & password modal when clicking Sign In. 

**To Reproduce**
1. Navigate to the Aleph landing page an a unauthenticated user.
2. Click `Sign In`
3. Click `Don't have account? Register!`
4. Fill in your email, click `Sign Up`
5. Close the modal.
6. Click `Sign In` again
7. What the user now sees in a modal titled `Check your inbox`. There is no way for the unauthenticated user to see the username & password modal again. 

Restarting the docker containers clears up this bug.

**Expected behavior**
After registering, when an unauthenticated user clicks `Sign In`, they expect to be able to see the username & password fields in the modal window.

**Aleph version**
3.13.0

**Screenshots**
image

tillprochaska:
Closes #2806. This allows users to switch between signin/signup any time and always resets the dialog state when closing and reopening it.

In addition to fixing the referenced bug, I’ve also sneaked in a few minor styling fixes.

https://user-images.githubusercontent.com/1512805/212868415-9c4b5080-5d09-45ce-add7-e5843ec0fb32.mov

### DIFF
--- a/ui/src/dialogs/AuthenticationDialog/AuthenticationDialog.jsx
+++ b/ui/src/dialogs/AuthenticationDialog/AuthenticationDialog.jsx
@@ -38,17 +38,20 @@ export class AuthenticationDialog extends Component {
   constructor(props) {
     super(props);
 
-    this.state = {
+    this.initialState = {
       submitted: false,
       firstSection: '',
       secondSection: 'hide',
     };
+
+    this.state = this.initialState;
 
     this.onLogin = this.onLogin.bind(this);
     this.onSignup = this.onSignup.bind(this);
     this.onRegisterClick = this.onRegisterClick.bind(this);
     this.onSignInClick = this.onSignInClick.bind(this);
     this.onOAuthLogin = this.onOAuthLogin.bind(this);
+    this.resetState = this.resetState.bind(this);
   }
 
   onOAuthLogin() {
@@ -86,12 +89,24 @@ export class AuthenticationDialog extends Component {
 
   onRegisterClick(e) {
     e.preventDefault();
-    this.setState({ firstSection: 'hide', secondSection: '' });
+    this.setState({
+      firstSection: 'hide',
+      secondSection: '',
+      submitted: false,
+    });
   }
 
   onSignInClick(e) {
     e.preventDefault();
-    this.setState({ firstSection: '', secondSection: 'hide' });
+    this.setState({
+      firstSection: '',
+      secondSection: 'hide',
+      submitted: false,
+    });
+  }
+
+  resetState() {
+    this.setState(this.initialState);
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


ui/src/dialogs/AuthenticationDialog/AuthenticationDialog.scss
```diff
@@ -22,8 +18,8 @@
   }
 
   .link-box {
-    margin-top: 2em;
-    margin-bottom: 2em;
-    // margin-left: 8.5em;
+    // TODO: This is a hacky way to center the link horizontally. We should
+    // refactor this and use flexbox to achieve that.
```
Rosencrantz, line 22: Any chance you could create an issue for that and reference it in this comment?

tillprochaska: To be honest, there are many cases similar to this one in our code. I’m not sure we will refactor them anytime soon (or at all), and the value of refactoring it is very limited as long as we do not plan to change this specific UI significantly.

If I open an issue for every case like this, we’d end up with a lot of issues that will see no activity for a very long time and add additional clutter to our issues list, so I’d prefer not to do that, but not sure if I’m missing something and am happy to discuss!

ui/src/dialogs/AuthenticationDialog/AuthenticationDialog.scss
```diff
@@ -1,12 +1,8 @@
 @import 'app/mixins.scss';
 
-.AuthenticationScreen {
+.AuthenticationScreen.AuthenticationScreen {
   width: 28em !important;
```
Rosencrantz, line 4: When I was a kid, !important was not considered best practice. Has this changed?

ui/src/dialogs/AuthenticationDialog/AuthenticationDialog.jsx
```diff
@@ -86,12 +90,24 @@ export class AuthenticationDialog extends Component {
...
+      submitted: false,
+    });
   }
 
   onSignInClick(e) {
     e.preventDefault();
-    this.setState({ firstSection: '', secondSection: 'hide' });
+    this.setState({
+      firstSection: '',
+      secondSection: 'hide',
```
Rosencrantz, line 104: Could firstSection / sectionSection be handled with a single toggle?

Rosencrantz (approved):

ui/src/dialogs/AuthenticationDialog/AuthenticationDialog.jsx
tillprochaska: Yes, probably, but note that I didn’t actually introduce this code in this PR (it’s just formatted differently), and I think refactoring it wouldn’t be worth it at this time.

Rosencrantz (approved):

PR #2832, opened 2023-01-23
Add renameuser CLI command

catileptic:
Usage: 

`aleph renameuser user@example.com Bob`

To rename the default `Aleph` user, the command is:

`aleph renameuser default CorpName`

### DIFF
--- a/aleph/logic/roles.py
+++ b/aleph/logic/roles.py
@@ -81,6 +81,17 @@ def create_user(email, name, password, is_admin=False):
     update_role(role)
     return role
 
+def rename_user(email, name):
+    """Rename an existing user"""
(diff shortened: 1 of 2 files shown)


Rosencrantz (commented):
Just a general thought, but what happens if a user tries to rename themselves to a name that is already in use in the system?

Rosencrantz:
> 

Yeah, ignore me. This comment was not thought through, I can see that what I am thinking about is email.

catileptic:
The name is only a label. The email is the unique identifier. So, duplicate names are fine.

Rosencrantz (commented):
Looks good. No comments.

PR #2839, opened 2023-01-27
Improve bookmarks warning messages

tillprochaska:
This PR…

1. improves the wording of the existing warning message to clarify that clearing browsing data or switching devices can cause bookmarks to get deleted,
2. adds a separate warning message popover when bookmarking something for the first time.

https://user-images.githubusercontent.com/1512805/215151100-a3802ed6-aec9-4792-937b-e9818b5043d5.mov

### DIFF
--- a/ui/src/components/common/BookmarkButton.tsx
+++ b/ui/src/components/common/BookmarkButton.tsx
@@ -1,8 +1,9 @@
-import { FC } from 'react';
+import { FC, useState } from 'react';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { injectIntl, FormattedMessage } from 'react-intl';
-import { Button, ButtonProps } from '@blueprintjs/core';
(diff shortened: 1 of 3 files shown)


Rosencrantz (approved):
Doesn't prevent merging but I couldn't see easily where you're storing the check for whether the message had been seen/dismissed.

tillprochaska:
> Doesn't prevent merging but I couldn't see easily where you're storing the check for whether the message had been seen/dismissed.

@Rosencrantz This is the line. It uses existing functionality to store certain UI configurations in local storage. It’s only been used to store whether users have dismissed the warning message when downloading source documents so far, but it can basically be used to store any config on the client.

https://github.com/alephdata/aleph/blob/feature/improve-bookmarks-warning-message/ui/src/components/common/BookmarkButton.tsx#L78

PR #2843, opened 2023-01-30
Implement server-side bookmarks

### LINKED ISSUE #2831: FEATURE: Persist bookmarks server-side
We have implemented an experimental bookmarks feature a while ago (see #2572). The current implementation stores bookmarks on the client-side. The feature has been received positively by users.

The current implementation has a major downside: bookmarks are stored in local storage. That means users can’t access bookmarks on other devices/in other browsers and users can delete all their bookmarks unintentionally by clearing their browsing data.

Bookmarks need to be stored on the server side.

- [ ] Bookmarks currently stored in local storage need to be migrated automatically, i.e. we probably need an API endpoint to bulk-add bookmarks. Bookmarks currently stored in local storage shouldn’t be overriden by bookmarks stored on the server, i.e. if a user has local bookmarks in multiple browsers and logs in, bookmarks from all browsers need to be migrated successfully.
- [ ] We need API endpoints to add and remove a bookmark for an entity.
- [ ] We need an API endpoint to list all bookmarks (probably with support for pagination).
- [ ] The `GET /entities/:id` endpoint needs to return whether an entity is bookmarked if the user is authenticated (this can be a request option)

**Additional context**
* We should focus on adding a server-side implementation for the feature that is currently available first. However, we should consider potential future enhancements of the feature (e.g. filtering search results for bookmarks, sharing bookmarks with other users, adding multiple separate bookmarks lists/tags, searching bookmarks, bookmarking collections etc.)
* I think there were ideas to make bookmarks FtM entities and/or to store bookmarks in ES.

tillprochaska:
The current experimental bookmarks feature in Aleph stores bookmark in local storage in the browser. This PR extends the feature to store bookmarks on the server, preventing a few common issues (for example users losing their bookmarks after clearing the browsing data). Closes #2831.

Specifically, there are three primary changes:

1. **API**: A new `/bookmarks` API endpoint to retrieve, create, and delete bookmarks.
2. **Frontend:** Changes to frontend to use the API instead of storing bookmarks locally.
3. **Migration:** New logic in the frontend to migrate bookmarks from client-side storage to the server.

### API

* Data is stored in a new `bookmark` table in Postgres. This table stores the role ID, entity ID and (to support efficient querying based on users permissions) the collection ID.
* There are API method to list bookmarks makes use of existing abstractions to fetch data from the database (incl. flexible offsets and limits) using the `DatabaseQuery` class, and merging it with entity data from the ElasticSearch index using the `Serializer` class. So it’s actually quite simple.

### UI

* Uses our existing data fetching logic based on Redux to handle data fetching and mutations. I had to extend this to support partial data invalidation, for details see f5ac9d5.

### Migration

* There’s a separate endpoint to batch create bookmarks specifically for the purpose of migrating existing bookmarks stored client-side. The idea is to remove any code related to this migration after a couple of months.
* When a user opens the bookmarks drawer and has local bookmarks that haven’t yet been migrated, a request to the migration endpoint is sent. If there have been errors while migrating bookmarks, users can download a list of bookmarks that couldn’t be migrated.

***

### How to test the migration

**Testing the happy path:**

1. Check out the `develop` branch.
2. Start Aleph.
3. Create a few entities.
4. Bookmark a few entities.
5. Stop Aleph.
6. Check out this branch.
7. Start Aleph.
8. Open the bookmarks drawer.
9. You should see a message notifying you that bookmarks now sync across devices. Confirm the message and you should see your bookmarks.
10. Clear your local storage.
11. Reload the page.
12. Open the bookmarks drawer.
13. Your bookmarks should still be there.

**Testing migration errors:**

1. Check out the `develop` branch.
2. Start Aleph.
3. Create a few entities.
4. Bookmark a few entities.
5. Now delete one or two of the bookmarked entities, but not all of them.
6. Stop Aleph.
7. Check out this branch.
8. Start Aleph.
9. Open the bookmarks drawer.
10. You should see a message about recent changes to bookmarks and that there was an issue with some of your bookmarks prompting you to download a list of affected bookmarks.
11. Download and confirm.
12. You should now be presented with a list of bookmarks that could be migrated successfully.
13. Clear your local storage.
14. Reload the page.
15. Open the bookmarks drawer.
16. Your bookmarks should still be there.

***

**Backend:**
- [x] Decide whether we want to have single `POST /bookmarks` endpoint that handles both creating a single bookmark as well as bulk creation (see comment below).
  * **Decision:** We want to have a separate endpoint for the purpose of migrating client-side bookmarks only. This endpoint can include specifics only relevant to this particular use case and we expect to remove it after some time.
- [x] `GET /bookmarks`: Endpoint should return bookmarks sorted by creation date, in descending order.
- [x] `POST /bookmarks`: Endpoint should validate `entity_id` when creating a new bookmark.
- [x] `POST /bookmarks`: Creating a bookmark for an entity that is already bookmarked shouldn’t raise an exception, but silently ignore the request or return a semantic status code (see review comment below).
- [x] Adjust `GET /entities/:entity_id` to return whether the entity is bookmarked if the user is authenticated
- [x] Optional: We should consider using bulk operations when creating database records for new bookmarks (see review comment below).
- [x] ~Optional: Delete bookmarks if entity is deleted. (Check entity set implementation)~
  * ~**Update:** Entity sets are very similar to bookmarks. I checked the current implementation and right now, if you delete an entity, corresponding entity set items aren’t deleted. They are obviously not returned from API responses and not displayed in the UI, but we keep the records in the database. I think it’s sensible to keep the bookmarks implementation similar for now and don’t think it’s a huge issue.~
  * **Update 2:** Actually we do clean up related records after deleting an entity. This is now also implemented for bookmarks.
- [x] Optional: Add proper response schemata in docstrings for API routes

**Frontend:**
- [x] Fetch bookmarks from backend API
- [x] Implement infinite scrolling for bookmarks list to support users with 100+ bookmarks. (We should expect users to accumulate many bookmarks over time.)
- [x] Send a one-time bulk create requests for bookmarks stored in local storage.

**Other:**
- [x] Check whether we use a foreign key constraint on the `entity_id` column? This will also help us get a better understanding of Aleph’s architecture.
  * The answer is no -- and it makes sense as there can be situations where entity data does exist in the index only and there is no corresponding record in the Postgres database.

### DIFF
--- a/aleph/logic/entities.py
+++ b/aleph/logic/entities.py
@@ -7,7 +7,7 @@ from followthemoney.types import registry
 from followthemoney.exc import InvalidData
 
 from aleph.core import db, cache
-from aleph.model import Entity, Document, EntitySetItem, Mapping
+from aleph.model import Entity, Document, EntitySetItem, Mapping, Bookmark
 from aleph.index import entities as index
 from aleph.queues import pipeline_entity, queue_task
 from aleph.queues import OP_UPDATE_ENTITY, OP_PRUNE_ENTITY
@@ -165,6 +165,7 @@ def prune_entity(collection, entity_id=None, job_id=None):
     if doc is not None:
         doc.delete()
     EntitySetItem.delete_by_entity(entity_id)
+    Bookmark.delete_by_entity(entity_id)
     Mapping.delete_by_table(entity_id)
     xref_index.delete_xref(collection, entity_id=entity_id)
     aggregator = get_aggregator(collection)
--- a/aleph/migrate/versions/c52a1f469ac7_create_bookmark_table.py
+++ b/aleph/migrate/versions/c52a1f469ac7_create_bookmark_table.py
@@ -1,0 +1,41 @@
+"""create bookmark table
+
+Revision ID: c52a1f469ac7
+Revises: 274270e01613
+Create Date: 2023-01-30 08:53:59.370110
+
+"""
+
+# revision identifiers, used by Alembic.
+from alembic import op
+import sqlalchemy as sa
+
+revision = "c52a1f469ac7"
+down_revision = "274270e01613"
+
+
+def upgrade():
+    op.create_table(
+        "bookmark",
+        sa.Column("id", sa.Integer(), primary_key=True),
+        sa.Column("created_at", sa.DateTime(), nullable=False),
+        sa.Column("role_id", sa.Integer(), sa.ForeignKey("role.id"), nullable=False),
+        sa.Column("entity_id", sa.String(length=128), nullable=False),
+        sa.Column(
+            "collection_id",
+            sa.Integer(),
+            sa.ForeignKey("collection.id"),
+            nullable=False,
+        ),
+        sa.UniqueConstraint("role_id", "entity_id"),
+    )
+
+    op.create_index(
+        op.f("ix_bookmark_role_id_collection_id_created_at"),
+        "bookmark",
+        ["role_id", "collection_id", "created_at"],
+    )
+
+
+def downgrade():
+    op.drop_table("bookmark")
--- a/aleph/model/bookmark.py
+++ b/aleph/model/bookmark.py
@@ -19,3 +19,9 @@ class Bookmark(db.Model, IdModel):
             "entity_id": self.entity_id,
             "collection_id": self.collection_id,
         }
+
+    @classmethod
+    def delete_by_entity(cls, entity_id):
+        query = db.session.query(Bookmark)
+        query = query.filter(Bookmark.entity_id == entity_id)
+        query.delete(synchronize_session=False)
--- a/aleph/model/bookmark.py
+++ b/aleph/model/bookmark.py
@@ -1,0 +1,21 @@
+from datetime import datetime
+from normality import stringify
+from aleph.core import db
+from aleph.model.common import ENTITY_ID_LEN, IdModel
+
+
+class Bookmark(db.Model, IdModel):
+    """A bookmark of an entity created by a user."""
+
+    created_at = db.Column(db.DateTime, default=datetime.utcnow)
+    role_id = db.Column(db.Integer, db.ForeignKey("role.id"))
+    collection_id = db.Column(db.Integer, db.ForeignKey("collection.id"))
+    entity_id = db.Column(db.String(ENTITY_ID_LEN))
+
+    def to_dict(self):
+        return {
+            "id": stringify(self.id),
+            "created_at": self.created_at,
+            "entity_id": self.entity_id,
+            "collection_id": self.collection_id,
+        }
--- a/aleph/tests/test_bookmarks_api.py
+++ b/aleph/tests/test_bookmarks_api.py
@@ -1,0 +1,224 @@
+import json
+import datetime
+from aleph.index.entities import index_entity
+from aleph.tests.util import TestCase, JSON
+from aleph.model import Bookmark
+from aleph.core import db
+
+
+class BookmarksApiTestCase(TestCase):
+    def setUp(self):
+        super(BookmarksApiTestCase, self).setUp()
+
+        self.role, self.headers = self.login()
+        self.collection = self.create_collection(self.role, label="Politicians")
+
+        data = {"schema": "Person", "properties": {"name": "Angela Merkel"}}
+        self.entity = self.create_entity(data=data, collection=self.collection)
+        index_entity(self.entity)
+
+    def test_bookmarks_index_auth(self):
+        res = self.client.get("/api/2/bookmarks")
+        assert res.status_code == 403, res
+
+        res = self.client.get("/api/2/bookmarks", headers=self.headers)
+        assert res.status_code == 200, res
+        assert res.json["total"] == 0, res.json
+
+    def test_bookmarks_index_results(self):
+        other_role, _ = self.login("tester2")
+        other_entity = self.create_entity(
+            data={"schema": "Person", "properties": {"name": "Barack Obama"}},
+            collection=self.collection,
+        )
+        index_entity(other_entity)
+
+        bookmarks = [
+            Bookmark(
+                entity_id=self.entity.id,
+                collection_id=self.entity.collection_id,
+                role_id=self.role.id,
+            ),
+            Bookmark(
+                entity_id=other_entity.id,
+                collection_id=self.entity.collection_id,
+                role_id=other_role.id,
+            ),
+        ]
+        db.session.add_all(bookmarks)
+        db.session.commit()
+
+        res = self.client.get("/api/2/bookmarks", headers=self.headers)
+        assert res.status_code == 200, res
+        assert res.json["total"] == 1, res.json
+        entity = res.json["results"][0]["entity"]
+        assert entity["properties"]["name"][0] == "Angela Merkel", res.json
+
+    def test_bookmarks_index_order(self):
+        other_entity = self.create_entity(
+            data={"schema": "Person", "properties": {"name": "Barack Obama"}},
+            collection=self.collection,
+        )
+        index_entity(other_entity)
+
+        old_bookmark = Bookmark(
+            entity_id=self.entity.id,
+            collection_id=self.entity.collection_id,
+            role_id=self.role.id,
+            created_at=datetime.date(2005, 11, 22),
+        )
+        new_bookmark = Bookmark(
+            entity_id=other_entity.id,
+            collection_id=other_entity.collection_id,
+            role_id=self.role.id,
+            created_at=datetime.date(2009, 1, 20),
+        )
+        db.session.add_all([old_bookmark, new_bookmark])
+        db.session.commit()
+
+        res = self.client.get("/api/2/bookmarks", headers=self.headers)
+        assert res.status_code == 200, res
+        assert res.json["total"] == 2, res.json
+        first = res.json["results"][0]["entity"]
+        second = res.json["results"][1]["entity"]
+        assert first["properties"]["name"][0] == "Barack Obama", first
+        assert second["properties"]["name"][0] == "Angela Merkel", second
+
+    def test_bookmarks_index_access(self):
+        other_role = self.create_user(foreign_id="other")
+        secret_collection = self.create_collection(other_role, label="Top Secret")
+
+        data = {"schema": "Person", "properties": {"name": ["Mister X"]}}
+        secret_entity = self.create_entity(data, secret_collection)
+        index_entity(secret_entity)
+
+        bookmark = Bookmark(
+            entity_id=secret_entity.id,
+            collection_id=secret_entity.collection_id,
+            role_id=self.role.id,
+        )
+        db.session.add(bookmark)
+        db.session.commit()
+
+        res = self.client.get("/api/2/bookmarks", headers=self.headers)
+
+        assert res.status_code == 200, res
+        assert res.json["total"] == 0, res.json
+        assert len(res.json["results"]) == 0, res.json
+
+    def test_bookmarks_create(self):
+        count = Bookmark.query.count()
+        assert count == 0, count
+
+        res = self.client.post(
+            "/api/2/bookmarks",
+            headers=self.headers,
+            data=json.dumps({"entity_id": self.entity.id}),
+            content_type=JSON,
+        )
+        assert res.status_code == 201, res
+        entity = res.json.get("entity")
+        assert entity["properties"]["name"][0] == "Angela Merkel", res.json
+
+        count = Bookmark.query.count()
+        assert count == 1, count
+        bookmark = Bookmark.query.first()
+        assert bookmark.entity_id == self.entity.id, bookmark.entity_id
+        assert bookmark.role_id == self.role.id, bookmark.role_id
+
+    def test_bookmarks_create_validate_access(self):
+        other_role = self.create_user(foreign_id="other")
+        secret_collection = self.create_collection(other_role, label="Top Secret")
+
+        data = {"schema": "Person", "properties": {"name": ["Mister X"]}}
+        secret_entity = self.create_entity(data, secret_collection)
+        index_entity(secret_entity)
+
+        res = self.client.post(
+            "/api/2/bookmarks",
+            headers=self.headers,
+            data=json.dumps({"entity_id": secret_entity.id}),
+            content_type=JSON,
+        )
+        assert res.status_code == 400, res
+        message = res.json["message"]
+        assert message.startswith("Could not bookmark the given entity"), message
+
+        count = Bookmark.query.count()
+        assert count == 0, count
+
+    def test_bookmarks_create_validate_exists(self):
+        invalid_entity_id = self.create_entity(
+            {"schema": "Company"}, self.collection
+        ).id
+
+        res = self.client.post(
+            "/api/2/bookmarks",
+            headers=self.headers,
+            data=json.dumps({"entity_id": invalid_entity_id}),
+            content_type=JSON,
+        )
+        assert res.status_code == 400, res
+        message = res.json["message"]
+        assert message.startswith("Could not bookmark the given entity"), message
+
+        count = Bookmark.query.count()
+        assert count == 0, count
+
+    def test_bookmarks_create_idempotent(self):
+        count = Bookmark.query.count()
+        assert count == 0, count
+
+        res = self.client.post(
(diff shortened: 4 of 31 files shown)


aleph/tests/test_bookmarks_api.py
```diff
@@ -0,0 +1,113 @@
...
+        db.session.add(bookmark)
+        db.session.commit()
+
+        res = self.client.get("/api/2/bookmarks", headers=self.headers)
+        assert res.status_code == 200, res
+        assert res.json["total"] == 1, res.json
+        entity = res.json["results"][0]["entity"]
+        assert entity["properties"]["name"][0] == "Angela Merkel", res.json
+
+    def test_bookmarks_index_read_permission(self):
```
tillprochaska, line 38: This test is currently failing. This case is relevant if a user previously had access to a collection, bookmarked and entity from the collection, then lost access to the collection.

Right now, the bookmark is still included in the response, even though the actual entity data won’t be present:

```jsonc
{
  // ...
  "total": 1,
  "results": [
    {
      "created_at": "2022-01-31T...",
      // No `entity` key
    }
  ]
}
```

This is ok, although it would be a bit more elegant if the results simply excluded bookmarks that reference entities the user doesn’t have access to. Not sure though if it’s worth the effort to implement this.

aleph/tests/test_bookmarks_api.py
```diff
@@ -0,0 +1,113 @@
...
+        entity = res.json[0]["entity"]
+        assert entity["properties"]["name"][0] == "Angela Merkel", res.json
+
+        count = Bookmark.query.count()
+        assert count == 1, count
+        bookmark = Bookmark.query.first()
+        assert bookmark.entity_id == self.entity.id, bookmark.entity_id
+        assert bookmark.role_id == self.role.id, bookmark.role_id
+
+    def test_bookmarks_create_validate_entity_id(self):
```
tillprochaska, line 78: This test is currently failing. Users can currently create bookmarks for random IDs without validations.

aleph/views/bookmarks_api.py
```diff
@@ -1,51 +1,88 @@
...
-    log.info("User [%s]: Created bookmark", request.authz.role.id)
-    return jsonify({}, 200)
+    require(request.authz.session_write)
+    data = parse_request("BookmarkBulkCreate")
+
+    bookmarks = [
+        Bookmark(entity_id=bookmark.get("entity_id"), role_id=request.authz.id)
+        for bookmark in data
+    ]
+    db.session.add_all(bookmarks)
```
tillprochaska, line 59: Could be implemented using SQLAlchemy’s [`bulk_save_objects`](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects). This feature is deprecated though, but SQLAlchemy 2 has an equivalent replacement (I think -- see docs).

tillprochaska: Apart from that, I’m not sure whether bulk creation should be handled by the default `POST /bookmarks` endpoint or whether it should be a separate endpoint (that we could potentially remove again some time after this feature has been released). Might lead to a cleaner API design and simplify semantics?

aleph/views/bookmarks_api.py
```diff
@@ -1,51 +1,88 @@
...
+    """
+    require(request.authz.logged_in)
+
+    query = Bookmark.query.filter_by(role_id=request.authz.id)
+    result = DatabaseQueryResult(request, query)
+    return BookmarkSerializer.jsonify_result(result)
+
+
 @blueprint.route("/api/2/bookmarks", methods=["POST"])
 def create():
```
tillprochaska, line 38: This endpoint should handle the request gracefully when the/a given entity is already bookmarked and simply ignore the entity. Alternatively, if there is a more semantic way to express this in HTTP, it would also be fine to return a non-2xx status code. In that case, we’d handle this explicitly on the frontend.

aleph/views/bookmarks_api.py
```diff
@@ -1,51 +1,88 @@
...
+    response = serializer.serialize_many(bookmarks)
+    return jsonify(response, status=201)
 
 
-@blueprint.route("/api/2/bookmarks/<bookmark_id>", methods=["DELETE"])
-def delete(bookmark_id):
-    """Delete a bookmark. This is currently a stub endpoint, as the current version
-    of the bookmarks features stores bookmarks client side.
+@blueprint.route("/api/2/bookmarks/<entity_id>", methods=["DELETE"])
+def delete(entity_id):
```
tillprochaska, line 68: This endpoint should handle the request gracefully when the given bookmark doesn’t exist. Returning a 404 status code would be fine as well, we can handle this case explicitly on the frontend.

aleph/model/bookmark.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime
+from normality import stringify
+from aleph.core import db
+from aleph.model.common import ENTITY_ID_LEN, IdModel
+
+
+class Bookmark(db.Model, IdModel):
```
tillprochaska, line 7: `aleph.model.common.DatedModel` defines `created_at` and `updated_at` timestamps and also includes a few utility methods. However, bookmarks don’t have an `updated_at` timestamp.

tillprochaska:
@catileptic I’ve thought about the bulk endpoint: Right now, the default `POST /bookmarks` endpoint supports bulk creation because, this way, we can support both creating a single and multiple bookmarks using a single endpoint.

Bulk creation is only required for a one-time migration of client-side bookmarks. Apart from that, the bulk endpoint will need to silently ignore entities that are already bookmarked or that do not exist/are not accessible to the user. It will need to support a user-provided `created_at` timestamp, even though that doesn’t make in cases except for the one-time migration. Maybe it will need to have a few more quirks because we haven’t considered other edge cases yet.

Maybe it’s better to have the default `POST /bookmarks` endpoint create a single bookmark, handling requests in a simple and semantic way, without trying to support the use case of the one-time migration.

Then, we could have a separate `POST /bookmarks/bulk` (or similar) endpoint that handles bulk creation for the purpose of the one-time migration. This endpoint can be a bit weird and handle edge cases specific to the migration in a way that’s convenient for us (even though it may not be the most semantic or clean way). After some time, we can simply remove the endpoint and the complexity it introduces.

aleph/migrate/versions/c52a1f469ac7_create_bookmark_table.py
```diff
@@ -0,0 +1,33 @@
...
+down_revision = "274270e01613"
+
+
+def upgrade():
+    op.create_table(
+        "bookmark",
+        sa.Column("id", sa.Integer(), primary_key=True),
+        sa.Column("created_at", sa.DateTime(), nullable=False),
+        sa.Column("role_id", sa.Integer(), sa.ForeignKey("role.id"), nullable=False),
+        sa.Column("entity_id", sa.String(length=128), nullable=False),
```
tillprochaska, line 23: Maybe we should also track the `collection_id` the entity belongs to here? This would allow us to filter by collection efficiently (a use case we will likely want to support in the future, but not 100% sure) and it would allow us to select only bookmarks in collections a user has access to (see https://github.com/alephdata/aleph/pull/2843/files#r1092250155).

This may be a bit of a premature optimization, but would save us the manualy work to populate this information for existing bookmarks in the future.

aleph/views/bookmarks_api.py
```diff
@@ -123,3 +124,57 @@ def delete(entity_id):
...
+                      format: entity-id
+    """
+    require(request.authz.session_write)
+    data = parse_request("BookmarkMigrate")
+    values = []
+    errors = []
+
+    for bookmark in data:
+        try:
+            entity = get_index_entity(bookmark.get("entity_id"))
```
tillprochaska, line 161: Could instead use `entities_by_ids` from `aleph.index.entities` to get all entities in a single ES request, but would need to manually check permissions.

aleph/logic/entities.py
```diff
@@ -165,6 +165,7 @@ def prune_entity(collection, entity_id=None, job_id=None):
     if doc is not None:
         doc.delete()
     EntitySetItem.delete_by_entity(entity_id)
+    Bookmark.delete_by_entity(entity_id)
```
tillprochaska, line 168: Currently, this isn’t covered by tests. I was trying to find existing tests for this method without success so far, and I guess it doesn’t make a lot of sense to test `prune_entity` in isolation? Anyone knows whether this is already covered as part of a bigger integration-style test somewhere?

aleph/tests/util.py
```diff
@@ -223,7 +224,7 @@ def setUp(self):
         if not hasattr(SETTINGS, "_global_test_state"):
             SETTINGS._global_test_state = True
             destroy_db()
-            db.create_all()
+            flask_migrate.upgrade()
```
tillprochaska, line 227: Previously, we created the DB schema based on the model classes. I know that some people use Alembic in a way where the model classes are the only source of truth and contain all information required to generate the schema, and then auto-generate alembic revisions based on the model classes. We don’t seem to do that though (at least not consistently), as many indexes etc. are only defined in revision files, so we ended up with inconsistent DB schemata in test and dev/prod environemnts.

Not sure if there is any reason not to use the actual migrations (except for maybe test performance and historic reason)?

I tested if this has an effect on how long it takes to execute tests but couldn’t find a significant difference.

aleph/views/bookmarks_api.py
```diff
@@ -176,5 +176,6 @@ def migrate():
         index_elements=[Bookmark.role_id, Bookmark.entity_id],
     )
     db.session.execute(stmt)
+    db.session.commit()
```
tillprochaska, line 179: I accidentally removed this line in a previous commit or rebase I think. I was a little surprised the existing test didn’t catch this. When running this in development, if the session is not committed, no bookmarks are created (expected). However, in a test environment, asserting that bookmarks exist does not fail.

Pretty sure the cause is related to SQLAlchemy’s session handling, autocommit/transaction isolation config or something like that, but wasn’t able to isolate it yet.

aleph/views/bookmarks_api.py
Rosencrantz: Is this following similar conventions in other parts of the codebase? What I'm saying is that if other areas are manually checking the permissions it might be worth following the convention, if not, then that's a good indicator that perhaps this is the way to go

aleph/views/bookmarks_api.py
```diff
@@ -1,51 +1,181 @@
...
-    return jsonify({}, 200)
+    require(request.authz.session_write)
+    data = parse_request("BookmarkCreate")
+    entity_id = data.get("entity_id")
+
+    try:
+        entity = get_index_entity(entity_id, request.authz.READ)
+    except (NotFound, Forbidden):
+        raise BadRequest(
+            "Could not bookmark the given entity as the entity does not exist or you do not have access."
```
Rosencrantz, line 81: Perhaps: "We're sorry but we are unable to bookmark this entity for you" - But if this happens are we logging/tracking it?

ui/src/components/common/BookmarksMigration.tsx
```diff
@@ -0,0 +1,130 @@
...
+  if (state === 'LOADING') {
+    return <Spinner className="BookmarksMigration__spinner" size={100} />;
+  }
+
+  if (state === 'DONE' && failed.length <= 0) {
+    return (
+      <NonIdealState
+        icon="star"
+        title="Changes to bookmarks"
+        description="We have recently rolled out enhancements for bookmarks! Bookmarks are now more reliable and sync across all your devices."
```
Rosencrantz, line 66: We've updated bookmarks. The bookmarks that you create will now synchronize across all your devices"

ui/src/components/common/BookmarksMigration.tsx
```diff
@@ -0,0 +1,130 @@
...
+  if (state === 'DONE' && failed.length >= 0) {
+    return (
+      <NonIdealState
+        icon="warning-sign"
+        title="Changes to bookmarks"
+        description={
+          <>
+            <p>
+              We have recently rolled out enhancements for bookmarks! This
+              required updating bookmarks to a new storage format.
```
Rosencrantz, line 90: Based on user feedback we've updated the bookmarks with a new way of storing the items that you select.

ui/src/components/common/BookmarksMigration.tsx
```diff
@@ -0,0 +1,130 @@
...
+            </p>
+            <p>
+              <strong>
+                Unfortunately, {failed.length} out of {succeeded.length}{' '}
+                bookmarks could not be updated to the new format.
+              </strong>
```
Rosencrantz, line 96: Unfortunately, x out of the y bookmarks that you had previously created could not be updated to the new format

ui/src/components/common/BookmarksMigration.tsx
```diff
@@ -0,0 +1,130 @@
...
+            </p>
+            <p>
+              Most likely that’s because the data you bookmarked has been
+              removed from Aleph. Download a copy of the affected bookmarks
+              to proceed.
```
Rosencrantz, line 101: The most likely reason for this is that the data you bookmarked has now been removed from Aleph. To proceed, please download a copy of the affected bookmarks.

aleph/migrate/versions/c52a1f469ac7_create_bookmark_table.py
```diff
@@ -0,0 +1,41 @@
...
+import sqlalchemy as sa
+
+revision = "c52a1f469ac7"
+down_revision = "274270e01613"
+
+
+def upgrade():
+    op.create_table(
+        "bookmark",
+        sa.Column("id", sa.Integer(), primary_key=True),
```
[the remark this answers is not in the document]
tillprochaska: An integer, primary key column will auto increment by default:

> The default value is the string "auto", which indicates that a single-column (i.e. non-composite) primary key that is of an INTEGER type with no other client-side or server-side default constructs indicated should receive auto increment semantics automatically. Other values include True (force this column to have auto-increment semantics for a [composite primary key](https://docs.sqlalchemy.org/en/14/glossary.html#term-composite-primary-key) as well), False (this column should never have auto-increment semantics), and the string "ignore_fk" (special-case for foreign key columns, see below).

https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column.params.autoincrement

When you check the table schema of the `bookmark` table, you should see that the `id` column has a default value of `nextval('bookmark_id_seq'::regclass)`.

aleph/views/bookmarks_api.py
tillprochaska: Yes and no. For singular resources, we do rely on these helpers to check permissions and throw appropriate errors in case of missing permissions etc.

For lists of many resources, we usually ensure that only resources the current user can access are returned at the query level, and we have another last-resort check when serializing resources in API responses.

This case is a little different though -- we’re not processing data from the database, but from a user. And we can’t rely on the serializer-level check because we actually need to know about the issues in order to show a message to the user.

The current solution is the least complex, but I’m not sure if it might lead to issues when migrating huge lists of bookmarks due to the O(n) ES requests. So far it hasn’t been a problem locally, and I’ll make sure to test it on staging once it’s out there.

aleph/migrate/versions/c52a1f469ac7_create_bookmark_table.py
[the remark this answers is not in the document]
Rosencrantz: That is what I thought. All things being equal I'd probably prefer a UUID which is prefixed with an identifier something like: "bmk|6ab23112beab459cab30c9ef1597a57e" Its random which means it non deterministic, but it also allows you to see easily what type of identifier it is (in this case a bookmark id). 

However, all the other tables we have use a simple incrementing integer value, and I think consistency is probably more important. All this is to say, this is fine.

Thanks for clarifying though.

aleph/logic/entities.py
stchris: Indeed, I'm not seeing any test coverage for this. And I agree it should go into an integration test. Shall we leave a TODO here?

aleph/views/bookmarks_api.py
stchris: Hm great point and I see no immediate answer to this. Anything wrapped in a session should be running in a transaction so I would expect this to require a commit. But perhaps the the default isolation level is autocommit? 🤔

stchris (approved):
Looks good to me, Till! Outside of the mysterious `commit` question, which I can't quite answer, everything seems good to go! 👏

PR #2921, opened 2023-03-06
Indicate optional fields when creating new timeline items

tillprochaska:
The end date of a timeline item is always optional. This adds "(optional)" after the respective form labels to make this clear for users. Additionally, this also adds proper validation to required fields (i.e. you can’t create a timeline item without a start date).

Screen Shot 2023-03-06 at 07 41 06

### DIFF
--- a/ui/src/components/Timeline/Timeline.test.tsx
+++ b/ui/src/components/Timeline/Timeline.test.tsx
@@ -255,7 +255,7 @@ it('allows creating new items', async () => {
   // 'Event' is the default schema
   expect(screen.getByRole('textbox', { name: 'Name' })).toBeInTheDocument();
   expect(screen.getByRole('textbox', { name: 'Start date' }));
-  expect(screen.getByRole('textbox', { name: 'End date' }));
+  expect(screen.getByRole('textbox', { name: 'End date (optional)' }));
 
   // Change schema to 'Company'
   await userEvent.click(screen.getByRole('button', { name: 'Type' }));
@@ -266,7 +266,9 @@ it('allows creating new items', async () => {
   const incorporation = screen.getByRole('textbox', {
     name: 'Incorporation date',
   });
-  const dissolution = screen.getByRole('textbox', { name: 'Dissolution date' });
+  const dissolution = screen.getByRole('textbox', {
+    name: 'Dissolution date (optional)',
+  });
 
   expect(name).toBeInTheDocument();
   expect(incorporation).toBeInTheDocument();
--- a/ui/src/components/Timeline/Timeline.test.tsx
+++ b/ui/src/components/Timeline/Timeline.test.tsx
@@ -250,7 +250,9 @@ it('allows creating new items', async () => {
   await userEvent.click(screen.getByRole('button', { name: 'Add item' }));
 
   const dialog = screen.getByRole('dialog', { name: /Add new item/ });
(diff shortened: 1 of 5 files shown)


stchris (commented):
Looks like a nice usability addition! Out of curiosity: as I understand it this is only used in the `TimeLineCreateForm`, is there any way to make this a more generic thing we have on all our forms?

tillprochaska:
> Out of curiosity: as I understand it this is only used in the TimeLineCreateForm, is there any way to make this a more generic thing we have on all our forms?

Hmm, you mean the "optional" indicator? Or the validation for date formats?

As far as I know, most other forms (especially those related to creating entities or updating entity data) don’t have mixed required/optional fields. Either all fields are required or all are optional.

However, there are some other forms that are similar, for example when creating an investigation, all you really need to fill out is the title. It may be a good idea to add optional indicators there as well. (Although I’ll check with the data team before, not sure if they actually want people to fill out the description field…)

Screen Shot 2023-03-06 at 09 36 18

Regarding date format validation: We already have that (at least in all places that I know use date inputs), however not using native browser validation. That means this implementation is a bit inconsistent with some other UIs, but I think it’s a good trade-off in this case.

stchris:
> Hmm, you mean the "optional" indicator?

Sorry, yes, I mean the optional indicator. I think all of our create forms have some form of optional fields. Good point on checking with the data team.

tillprochaska:
> I think all of our create forms have some form of optional fields.

What forms are you thinking of? The ones that I’m aware of and that have optional fields are the "Create invesetigation/network diagram/timelines/list" forms.

stchris:
> > I think all of our create forms have some form of optional fields.
> 
> What forms are you thinking of? The ones that I’m aware of and that have optional fields are the "Create invesetigation/network diagram/timelines/list" forms.

Those are the ones I had in mind as well. Apologies for the lengthy discussion in here, happy to have it outside of this PR.

tillprochaska:
No worries, it’s good feedback!

ui/src/components/Timeline/Timeline.test.tsx
```diff
@@ -255,7 +255,7 @@ it('allows creating new items', async () => {
   // 'Event' is the default schema
   expect(screen.getByRole('textbox', { name: 'Name' })).toBeInTheDocument();
   expect(screen.getByRole('textbox', { name: 'Start date' }));
-  expect(screen.getByRole('textbox', { name: 'End date' }));
+  expect(screen.getByRole('textbox', { name: 'End date (optional)' }));
```
Rosencrantz, line 258: For this, I don't suppose we have the translation for optional for other languages do we?

tillprochaska: It will use translations (as soon as someone has translated this string in Transifex)

https://github.com/alephdata/aleph/pull/2921/files#diff-391f5ad77c4995b4373c13316eafa3d5b2987c7f6a7ebcb611eabfbde3323469R101-R106

Rosencrantz (approved):

tillprochaska:
Just pushed another small change that enables the submit button only when the form is valid which is a pattern we use for other forms as well and provides some additional guidance for users.

Uploading Screen Recording 2023-03-06 at 12.23.52.mov…

Rosencrantz (approved):

PR #2923, opened 2023-03-06
First end to end test

stchris:
How to review this: run `make e2e`. You can run the visual debugger by running `PWDEBUG=1 make e2e`.

TODOS

- [x] figure out the "disable host check" thing
- [x] make CI work
- [x] reduce the amount of events this gets triggered by
- [x] perhaps parametrize the BASE_URL in CI
- [x] have the test clean up after itself
- [x] use authentication instead of single user mode
- [x] have the tests run on separate (configurable) ElasticSearch & Postgres DBs
- [x] prefix test-created resources with `[e2e]`
- [x] fix the timestamp prefix when waiting for the ui container


docker-compose.dev.yml
```diff
@@ -152,6 +152,31 @@ services:
     environment:
       PORT: "8080"
       CI: "true"
+      DANGEROUSLY_DISABLE_HOST_CHECK: "true"
```
stchris, line 155: - [ ] I had to use this because otherwise I would not be able to make http calls from the `e2e` container.

docker-compose.dev.yml
```diff
@@ -152,6 +152,31 @@ services:
...
+
+  e2e:
+    build:
+      context: e2e
+    environment:
+      BASE_URL: "http://ui:8080"
+    volumes:
+      - "./e2e:/e2e"
+
+  worker:
```
stchris, line 167: Added this as a more comfortable way of getting a worker up from a docker compose context.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,23 @@
+name: End to end tests
+
+on:
+  push
```
stchris, line 4: - [ ] up for discussion, we probably want something more sensible here

docker-compose.dev.yml
tillprochaska: And as a last resort, we could override the the `allowedHosts` option in Webpack Dev Server: https://craco.js.org/docs/configuration/devserver/. As this isn’t exposed directly as a config option, it requires a hacky workaround using craco and will likely break in the future, so if something else works, I’d prefer that.

stchris: Ah so setting `ALEPH_UI_HOST` is enough, great to know that! I thought about the docker-compose override file as well, but I'd like to avoid it if it's not necessary. Thanks!

Makefile
```diff
@@ -99,5 +99,15 @@ translate: dev
 	npm run --prefix ui translate
 	pybabel compile -d aleph/translations -D aleph -f
 
-.PHONY: build services
+e2e: services
+	$(COMPOSE) up -d api ui worker
```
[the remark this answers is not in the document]
stchris: The reason is that the `web` target brings up the services in the foreground.

docker-compose.dev.yml
stchris: Perhaps this is https://github.com/facebook/create-react-app/issues/12304 ?

stchris: Meh, I doubled down and went with the docker compose override for now 🤷

e2e/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/playwright/python:v1.31.0-focal
+COPY entrypoint.sh .
+RUN apt update -y \
+    && apt install -y netcat wget \
+    && apt-get -qq -y autoremove \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN pip install pytest-playwright
+RUN curl -O https://raw.githubusercontent.com/eficode/wait-for/v2.2.3/wait-for && chmod +x ./wait-for
```
tillprochaska, line 9: It’s not an issue in this case, because the e2e workflow doesn’t have access to any secrets, so maybe this is a bit nitpicky, but executing bash scripts directly from the web always makes me a bit uneasy. Alternatively, we could just inline the bash script in the repo (it’s probably not going to be updated frequently)?

docker-compose.dev.yml
tillprochaska: Yeah, I played around with it for a bit yesterday trying to find a solution that doesn’t require ignoring the host check or hacky workarounds to adjust the configuration of the underlying Webpack dev server, but without success :/

docker-compose.dev.yml
```diff
@@ -155,6 +155,30 @@ services:
     env_file:
       - aleph.env
 
+  e2e:
+    build:
+      context: e2e
+    environment:
+      BASE_URL: "http://ui:8080"
+    volumes:
+      - "./e2e:/e2e"
```
tillprochaska, line 164: Maybe this could go into `docker-compose.e2e.yml` too, now that we have a separate config file?

e2e/test_sanity.py
```diff
@@ -0,0 +1,42 @@
...
+    page.get_by_role("button", name="New investigation").click()
+    page.get_by_placeholder("Untitled investigation").click()
+    page.get_by_placeholder("Untitled investigation").fill("[test] Don't look in here")
+    page.get_by_role("button", name="Save").click()
+    page.get_by_role("button", name="Upload documents").click()
+    page.get_by_label("Choose files to upload...").set_input_files(
+        "e2e/fixtures/random.odt"
+    )
+    page.get_by_role("button", name="Upload", exact=True).click()
+    page.locator("section").get_by_role("button", name="Close").click()
```
tillprochaska, line 26: Something like this may be a little more robust in case the markup changes or there are other `section` elements:

```suggestion
    page.get_by_role("dialog", name="Upload documents").get_by_role("button", name="Close").click()
```

tillprochaska (approved):
I added a few more comments below, but nothing important, feel free to ignore if you disagree :).

Also didn’t mention it before, but I tested this on my machine and `make e2e` did all the magic and it worked out of the box. (Although the test was failing, for the same reason it fails in CI).

One more suggestion, I often find it helpful to open a shell in the container then run commands directly in the container, because that sometimes makes it a faster to iterate, so instead of running `make e2e` repeatedly, when I’m actively working on tests, I’d prefer to run `make e2e-shell` (as a convenience), and then run `make e2e-local` in the container. Would it be ok if I added that to the Makefile?

stchris:
@tillprochaska I was looking at adding the `e2e-shell` target, but the e2e container does not have `make` installed. Would you like me to add it?

Just for comparison: here's my workflow when doing iterations during writing a test: I bring up Aleph with `make services` and then run `make e2e-local` inside the aleph virtualenv. The big advantage is that you can use the visual debugger and codegen tools this way.

tillprochaska:
> I bring up Aleph with make services and then run make e2e-local inside the aleph virtualenv.

Just to make sure I’m not misunderstanding you: You run `make e2e-local` in the `api` container? Or directly on your host machine?

> The big advantage is that you can use the visual debugger and codegen tools this way.

Yep, makes sense. I was going to look into doing this while running Playwright in Docker, but I assume this probably requires some fairly complex configuration, so probably not worth it.

Let’s just ignore my comment about this for now. I’ll try to figure out a workflow that works well for me first, and then I can always add something to the Makefile later :)

stchris:
> Just to make sure I’m not misunderstanding you: You run `make e2e-local` in the `api` container? Or directly on your host machine?

I'm running it on my host machine.

e2e/test_sanity.py
```diff
@@ -0,0 +1,47 @@
...
+    page.get_by_role("button", name="New investigation").click()
+    page.get_by_placeholder("Untitled investigation").click()
+    page.get_by_placeholder("Untitled investigation").fill(investigation_name)
+    page.get_by_role("button", name="Save").click()
+    page.get_by_role("button", name="Upload documents").click()
+    page.get_by_label("Choose files to upload...").set_input_files(
+        "e2e/fixtures/random.odt"
+    )
+    page.get_by_role("button", name="Upload", exact=True).click()
+    page.get_by_role("dialog", name="Upload documents").get_by_role("button", name="Close").click()
```
stchris, line 28: @tillprochaska I committed your more precise suggestion, but it seems to be less specific now, the test fails with: 

```
E       playwright._impl._api_types.Error: Error: strict mode violation: get_by_role("dialog", name="Upload documents").get_by_role("button", name="Close") resolved to 2 elements:
E           1) <button type="button" aria-label="Close" class="bp4-butt…>…</button> aka get_by_role("dialog", name="Upload documents").locator("div").filter(has_text="Upload documents").get_by_role("button", name="Close")
E           2) <button type="submit" class="bp4-button bp4-intent-prima…>…</button> aka locator("section").get_by_role("button", name="Close")
E
```

Would you ok with me rolling this change back?

tillprochaska: So the dialog actually has two close buttons:

image

So maybe try this?

```suggestion
    page.get_by_role("dialog", name="Upload documents").get_by_role("button", name="Close").first.click()
```

tillprochaska: Just to clarify why I’d prefer this over just a `section` selector, this is the section element in the dialog:

image

To be honest, I’m not entirely sure if it’s semantically appropriate to use a section element in this case (that’s another question), but it’s something that could likely change when we work on the dialog itself to improve the UI or when we need to upgrade Blueprint (the UI component library we’re using).

Apart from that, as `section` is a very generic HTML element, there’s a risk of false positives and this would also match any other close button within a section element, so if for example we added a new section to the investigation homepage and then displayed a popover with a close button to promote the new feature, it would falsely match the locator.

e2e/test_sanity.py
```diff
@@ -0,0 +1,47 @@
...
+    page.get_by_placeholder("Search this investigation").fill("superprecise")
+    page.get_by_placeholder("Search this investigation").press("Enter")
+    page.get_by_role("link", name="random.odt").click()
+    page.get_by_role("tab", name="Text").click()
+    page.locator("#page").click()
+    page.locator("#page").fill("14")
+    page.locator("#page").press("Enter")
+    page.get_by_role("button", name="Close").click()
+    page.get_by_role("link", name="[test] Don't look in here").click()
+    page.get_by_role("navigation").get_by_role("button").click()
```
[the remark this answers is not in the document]
stchris: It's the cog wheel on the top right of an investigation. I'll try to find a more specific selector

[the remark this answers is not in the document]
tillprochaska: Ah, ok, in this case, you may be out of luck, as the markup used for the button and drop-down doesn’t really have any helpful labels. It’s a known problem in many different places in the UI -- maybe just ignore it for now, and at some point I’ll find time to fix all the missing button labels :D

tillprochaska:
Oh, just noticed one more thing. When running the tests locally, they write to my local dev database and ES index -- I think it would be good to use a test database and index here as well, so it can be deleted/reset easily. For our current Python tests, we override the settings here:

https://github.com/alephdata/aleph/blob/main/aleph/tests/util.py#L32

https://github.com/alephdata/aleph/blob/main/aleph/tests/util.py#L29

e2e/test_sanity.py
stchris: Understood and agreed. I generally use the codegen tool and this is what it spat out. But overall this is good, because it's showing us where we could be more specific in our markup, right?

stchris:
> Oh, just noticed one more thing. When running the tests locally, they write to my local dev database and ES index -- I think it would be good to use a test database and index here as well, so it can be deleted/reset easily. For our current Python tests, we override the settings here:
> 
> https://github.com/alephdata/aleph/blob/main/aleph/tests/util.py#L32 https://github.com/alephdata/aleph/blob/main/aleph/tests/util.py#L29

I'll take this as an item into the discussion next week. To me it was practical at times to be able to open up my local Aleph installation and check what the tests have done, but having a clean env makes sense to me.

e2e/test_sanity.py
tillprochaska: > But overall this is good, because it's showing us where we could be more specific in our markup, right?

Yes, absolutely, that’s also why I added it to the agenda for our discussion next week. Basically, I think there are three ways to handle situations like this:

* Use whatever locator works at the moment or the codgen tool generates, even if it’s not perfect, but maybe add a comment to the test or track this in an issue. In the manual feature descriptions/tests I started some time ago, you’ll often see things like *"Click on link with checkmark icon (should have invisible label, should be button or radio input)"*.

* Fix the issue at the same time as you add a new E2E test. This would be cool, but there may be situations where that’s not possible because of priorities/time or because it would require more complex refactorings. In this case, a fix would be to add an label to the button that’s visually hidden, but still accessible to Playwright and assistive tech like screen readers etc.

* Similar to the previous option, but don’t bother with fixing the underlying issue properly, and simply add a `data-test-id` or `data-qa` attribute to the element and use that in tests.

.github/workflows/e2e.yml
Rosencrantz: Maybe this only runs when we merge to develop? Merging to develop suggests you have something ready to be completed?

Rosencrantz:
Might be good to have some cleanup in place for when the tests finish running?

.github/workflows/e2e.yml
stchris: That's what I'm thinking as well: on push / merge to `develop` and when we tag something.

PR #2935, opened 2023-03-08
Do not save timeline layout on initial load

tillprochaska:
The current implementation was using `useEffect` to trigger an API request (via the `onLayoutUpdate` callback) whenever the layout state (for example the color of an entity set item) changes. As `useEffect` will also execute the side effect on initial render, this caused an undesired API request directly after opening a timeline.

Instead of relying on `useEffect` to persist the layout state, this now triggers the `onLayoutUpdate` callback explicitly when the user performs certain actions (like changing the color of an item).

### DIFF
--- a/ui/src/components/Timeline/Timeline.test.tsx
+++ b/ui/src/components/Timeline/Timeline.test.tsx
@@ -51,8 +51,17 @@ describe('Sidebar', () => {
   it('allows changing entity color', async () => {
     const entities = [event1];
     const layout = { vertices: [] };
+    const onLayoutUpdate = jest.fn();
 
-    render(<Timeline {...defaultProps} entities={entities} layout={layout} />);
+    render(
+      <Timeline
+        {...defaultProps}
+        entities={entities}
+        layout={layout}
+        onLayoutUpdate={onLayoutUpdate}
+      />
+    );
+
     const rows = screen.getAllByRole('row');
     await userEvent.click(rows[1]);
 
@@ -67,6 +76,7 @@ describe('Sidebar', () => {
     await userEvent.click(swatches[3]);
     color = rows[1].style.getPropertyValue('--timeline-item-color');
     expect(color).toEqual(Colors.RED2);
+    expect(onLayoutUpdate).toHaveBeenCalledTimes(1);
   });
 
   it('does not allow changing entity color if not writeable', async () => {
--- a/ui/src/components/Timeline/Timeline.tsx
+++ b/ui/src/components/Timeline/Timeline.tsx
@@ -5,7 +5,7 @@ import { Colors } from '@blueprintjs/colors';
 import { Schema, Entity, Model } from '@alephdata/followthemoney';
 import c from 'classnames';
 import { TimelineRenderer, Layout, Vertex } from './types';
-import { TimelineItem } from './util';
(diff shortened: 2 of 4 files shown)


ui/src/components/Timeline/Timeline.tsx
```diff
@@ -64,13 +64,9 @@ const Timeline: FC<TimelineProps> = ({
   const sortedEntities = selectSortedEntities(state);
 
   const items = sortedEntities.map(
-    (entity) => new TimelineItem(entity, layout)
+    (entity) => new TimelineItem(entity, state.layout)
   );
 
-  useEffect(() => {
-    onLayoutUpdate(state.layout);
-  }, [state.layout, onLayoutUpdate]);
```
tillprochaska, line 72: This was the problematic section as the `onLayoutUpdate` callback was called not only when the layout had actually changed, but also on initial render.

ui/src/components/Timeline/Timeline.tsx
```diff
@@ -131,6 +127,9 @@ const Timeline: FC<TimelineProps> = ({
             writeable={writeable}
             onVertexChange={(vertex: Vertex) => {
               dispatch({ type: 'UPDATE_VERTEX', payload: { vertex } });
+              // State updates are executed asyncronously, so we need to compute the new layout explicitly
+              const newLayout = updateVertex(state.layout, vertex);
+              onLayoutUpdate(newLayout);
```
tillprochaska, line 132: We now call the callback explicitly after the user has layout-mutating actions (right now, the only such action is changing the color of an entity).

ui/src/components/Timeline/util.ts
```diff
@@ -266,3 +266,20 @@ export function useFormValidity(formRef: RefObject<HTMLFormElement>) {
...
+  const newVertices = [...vertices];
+
+  if (index < 0) {
+    newVertices.push(updatedVertex);
+  } else {
+    newVertices.splice(index, 1, updatedVertex);
+  }
+
+  return { vertices: newVertices };
+}
```
tillprochaska, line 285: This logic is now needed in two separate places, so I have extracted it from the state reducer into a separate helper function.

ui/src/components/Timeline/Timeline.test.tsx
```diff
@@ -67,6 +76,7 @@ describe('Sidebar', () => {
     await userEvent.click(swatches[3]);
     color = rows[1].style.getPropertyValue('--timeline-item-color');
     expect(color).toEqual(Colors.RED2);
+    expect(onLayoutUpdate).toHaveBeenCalledTimes(1);
```
tillprochaska, line 79: This expectation would fail without the fix (the callback would have been called twice).

Rosencrantz (approved):

Rosencrantz:
Just for my own understanding. This resolves an issue with an API call being made on initial render, when that isn't really required right? There is no actual user facing 'bug' per se?

tillprochaska:
> This resolves an issue with an API call being made on initial render, when that isn't really required right? There is no actual user facing 'bug' per se?

Mostly, however, when viewing a read-only timeline, the API call would be triggered as well (and fail due to lack of permissions) which would result in an "Error saving" status indicator being displayed in the top right corner.

PR #2938, opened 2023-03-09
Do not render delete and settings actions for read-only timelines, lists

tillprochaska:
Previously, we would display the settings and delete action buttons for read-only timelines and lists. This PR hides them and only displays them if the current user does actually have write access to a timelines or list.

| | Before | After |
| --- | --- | --- |
| Timelines | Screen Shot 2023-03-09 at 10 20 25 | Screen Shot 2023-03-09 at 10 20 43 |
| Lists | Screen Shot 2023-03-09 at 10 21 21 | Screen Shot 2023-03-09 at 10 21 03 |

### DIFF
--- a/ui/src/components/EntitySet/EntitySetManageMenu.jsx
+++ b/ui/src/components/EntitySet/EntitySetManageMenu.jsx
@@ -100,6 +100,10 @@ class EntitySetManageMenu extends Component {
       return this.renderExport();
     }
 
+    if (!entitySet.writeable) {
+      return null;
+    }
+
     if (isDiagram || isTimeline) {
       return (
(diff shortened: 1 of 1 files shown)


Rosencrantz (approved):
There really isn't much to look at here. I'm personally no fan of using null, but I think that is just my age showing.

tillprochaska:
> I'm personally no fan of using null, but I think that is just my age showing.

This is how you make a component render nothing in React :)

> In rare cases you might want a component to hide itself even though it was rendered by another component. To do this return null instead of its render output.

https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

PR #2953, opened 2023-03-22
Fix browser compatibility issue with translucent background in timeline item list

tillprochaska:
CSS doesn’t (yet) support altering the alpha channel of HEX colors. That’s why we’ve been using a hack to dynamically apply a translucent background color for timeline list items: instead of actually setting the background color, we rendered a pseudo element with a fully opaque background, then set `opacity` on that pseudo element.

This relied on absolutely positioning the pseudo element within the list item, and the item having a relative position. However, `position: relative` for table rows is undefined per the CSS spec, and it doesn’t render in the intended way in WebKit/Safari.

Instead, we now simply calculate a HEX color in JS land that approximates how the color would be rendered at a certain opacity on a white background.


Rosencrantz:
> CSS doesn’t (yet) support altering the alpha channel of HEX colors. That’s why we’ve been using a hack to dynamically apply a translucent background color for timeline list items: instead of actually setting the background color, we rendered a pseudo element with a fully opaque background, then set `opacity` on that pseudo element.
> 
> This relied on absolutely positioning the pseudo element within the list item, and the item having a relative position. However, `position: relative` for table rows is undefined per the CSS spec, and it doesn’t render in the intended way in WebKit/Safari.
> 
> Instead, we now simply calculate a HEX color in JS land that approximates how the color would be rendered at a certain opacity on a white background.

I am probably missing  some context here but if altering the alpha channel of the hex color is not possible, why not just use RGBA instead?

tillprochaska:
> I am probably missing some context here but if altering the alpha channel of the hex color is not possible, why not just use RGBA instead?

Users can define custom colors, so there’s no static set of colors. For that purpose, we use the color picker from react-ftm, which returns colors as a HEX string. It’s probably possible to change that to return RGB components, but that might have unintended side-effects in other places that also use the color picker (for example network diagrams). As we have a low to non-existent test coverage for network diagrams, I didn’t want to risk breaking things.

Alternatively, I could’ve manually converted the HEX string to the individual RGB components, but that woul’ve meant writing new code and tests whereas the `reduceTranslucentColor(hex: string, opacity: number): string` function was already there.

So the reasons are mostly not of a technical nature…

tillprochaska:
I think using RGBA colors in HEX notation might have actually worked, i.e. something like `background-color: var(--timeline-item-color)01`, where `01` is the alpha channel in HEX notation, but it would’ve been a bit ugly, too. Didn’t try it tbh.

tillprochaska:
Can’t wait until this is implemented in all modern browsers btw: https://www.w3.org/TR/css-color-5/#relative-colors

Which would allow changing the alpha channel of a random colors in a single line:

```css
--timeline-item-color: ##00ff00;
background-color: rgba(from var(--timeline-item-color) rgb / 0.1);
```

PR #2961, opened 2023-03-28
Add Sentry support

stchris:


Rosencrantz:
Nice. Is there anywhere else in the documentation that you think we should add information around Sentry for Aleph?

Issue #2673, opened 2022-11-04
Timelines


tillprochaska:
We would like to enhance the timelines feature to solve a number of problems:

- [ ] https://github.com/alephdata/followthemoney/pull/849
      Schemata that can be added to timelines are currently limited. Some of the schemata users frequently need in timelines (e.g. `Company`) can currently not be added to timelines. In order to be able to add other schemata to timelines, we need to add a new temporal extent feature to FtM.

- [ ] https://github.com/alephdata/aleph/pull/2629
      This will (eventually) lead to a more consistent UI (both implementation + UX) and simplify adding new entities to timelines.

- [ ] **Add existing entities to timeliens**
      Many users already use network diagrams to map out entities and relationships, but would like to be able to view the same entities in a timeline visualization as well.

- [ ] **Add timeline chart/Gantt-style visualization**

PR #2963, opened 2023-03-29
Show timeline items with invalid dates

tillprochaska:
**Note:** This PR is set up to be merged into the 3.14 release branch, but I think it would also be fine to merge it into develop and include it in the next release.

*** 

This displays timeline items (in list view) even if an item doesn’t have a valid date. Items without valid dates are displayed at the top of the list.

When creating new items through the timelines UI, dates are validated and users shouldn’t be able to add a new items without a valid date. However, that’s not the only way to add items to a timeline. Users may have created entities in network that don’t have dates yet (because dates aren’t a top priority when mapping out organization structures and relationships in a diagram). So when they add these existing entities to a timeline, we need to make users aware of the fact that they need to add dates to the entities in order to be able to visualize them.

Another case where this is relevant is when users have uploaded tabular data and created an entity mapping, but the source data contains invalid dates or uses an unrecognized date format.

Screen Shot 2023-03-30 at 00 08 52

Ref #2673

### DIFF
--- a/ui/src/components/Timeline/TimelineList/TimelineListItem.tsx
+++ b/ui/src/components/Timeline/TimelineList/TimelineListItem.tsx
@@ -64,11 +64,23 @@ const TimelineListItem = forwardRef<HTMLTableRowElement, TimelineListItemProps>(
         )}
       >
         <td className="TimelineListItem__date">
-          {start?.value}
-          <br />
-          <span className="TimelineListItem__property">
-            {start?.property.label}
-          </span>
+          {start ? (
+            <>
+              {start.value}
+              <br />
+              <span className="TimelineListItem__property">
+                {start?.property.label}
+              </span>
+            </>
+          ) : (
+            <div className="TimelineListItem__warning">
+              <Icon icon="warning-sign" color="white" />
+              <FormattedMessage
+                id="timeline.item.invalid_date"
+                defaultMessage="Invalid or no date"
+              />
+            </div>
+          )}
         </td>
         {showEndDate && (
           <td className="TimelineListItem__date">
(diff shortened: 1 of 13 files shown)


ui/src/components/Timeline/TimelineList/TimelineListItem.tsx
```diff
@@ -64,11 +64,23 @@ const TimelineListItem = forwardRef<HTMLTableRowElement, TimelineListItemProps>(
...
         <td className="TimelineListItem__date">
-          {start?.value}
-          <br />
-          <span className="TimelineListItem__property">
-            {start?.property.label}
-          </span>
+          {start ? (
+            <>
+              {start.value}
+              <br />
```
Rosencrantz, line 70: meh

Rosencrantz (approved):

ui/src/components/Timeline/state.ts
```diff
@@ -199,26 +198,30 @@ export function selectSelectedVertex(state: State): Vertex | null {
...
+        return 0;
+      }
+
+      if (aStart === null) {
+        return -1;
+      }
+
+      if (bStart === null) {
+        return 1;
+      }
```
Rosencrantz, line 217: I feel like this could be better or re-worked.

Rosencrantz (approved):

ui/src/components/Timeline/state.ts
tillprochaska: What would you prefer?

Rosencrantz: It's a good question. I didn't like the look of it, but I'm not honestly sure if I can come up with anything better in terms of being more terse, and easy to understand. It wouldn't be hard to make things more terse and reduce the number of if statements. But on reflection I think that the result would be harder for a future develop to understand. I was thinking of three seperate inline statements of the form:

value = 1;
value = (!a && !b) ? 0 : value;
value = (a && !b) ? -1 : value;
value - (!a && b) ? 1 : value; // this shouldn't be required because it's the only option left.
return value;

PR #3005, opened 2023-04-23
Add existing entities to timelines

tillprochaska:
This allows users to add existing entities of most schemata to timelines. Until now, this was only supported for interval entities (due to the limitations of the old timelines).

Note: Right now, when adding an existing entity that has now filled out date property (e.g. a company without an incorporation date), this entity wouldn’t be visible when displaying the timeline. This will be resolved as soon as #2963 is merged.

https://user-images.githubusercontent.com/1512805/233866664-f6d516b7-509b-4dcf-9bba-0610944141ad.mov

Ref #2673


Rosencrantz (approved):
I have no comments

PR #3006, opened 2023-04-24
Add additional confirmation UI before destructive actions

### LINKED ISSUE #2334: FEATURE: Confirmation before destructive actions
Before actually deleting data (investigations, network diagrams, entities, …) via the UI, we display a confirmation dialog like this:

Screenshot 2022-06-16 at 10 52 13

We see confirmation dialogs like this so often that we’re trained to simply click through them without actually reading and understanding what they’re trying to tell us.

A pattern that has stopped me from accidentally deleting data in the past is to request the user to type in the name of the thing they’re about to delete before you can continue.

It ensures that users know *what* data they’re about to delete, and it might also make them read any additional information in the dialog.

Screenshot 2022-06-16 at 10 53 02

**Describe alternatives you've considered**
-/-

**Additional context**
-/-

tillprochaska:
Recently, we had to handle multiple support cases of users deleting collections and entity sets by accident. In some cases, this was because users thought they were only deleting an entity within the investigation or entity set. Even though we do show a confirmation dialog before destructive actions, users often do not read the actual dialog message.

In order to prevent users from deleting data by accident, this PR adds an additional confirmation UI before deleting collections or entity sets and asks users to confirm the title of the collection/investigation.

The following video shows the flow to delete a collection:

https://user-images.githubusercontent.com/1512805/234068743-d6768c09-d2d4-47ad-8a0f-41ab9175cb99.mov

The flows for deleting entity sets (diagrams, lists, timelines, profiles) use a similar UI:

Screen Shot 2023-04-24 at 19 16 45

***

**To do:**

- [x] Extract common UI into `DeleteDialog` component to reduce redundancy
- [x] Adjust steps to delete investigation in e2e tests

Closes #2334

### DIFF
--- a/ui/src/dialogs/EntitySetDeleteDialog/EntitySetDeleteDialog.jsx
+++ b/ui/src/dialogs/EntitySetDeleteDialog/EntitySetDeleteDialog.jsx
@@ -1,5 +1,12 @@
 import React, { Component } from 'react';
-import { Alert, Intent } from '@blueprintjs/core';
+import {
+  Button,
+  Dialog,
+  DialogBody,
+  Intent,
+  InputGroup,
+  FormGroup,
+} from '@blueprintjs/core';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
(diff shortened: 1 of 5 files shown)


ui/src/dialogs/EntitySetDeleteDialog/EntitySetDeleteDialog.jsx
```diff
@@ -30,51 +22,118 @@ class EntitySetDeleteDialog extends Component {
...
+          />
+        )}
+        {type === 'profile' && (
+          <FormattedMessage
+            id="entityset.delete.confirm.profile"
+            defaultMessage="Delete this profile."
+            values={{ type }}
+          />
+        )}
+      </>
```
Rosencrantz, line 108: Nitpick. To me it would be better to create a set of values and then make only one <FormattedMessage /> element.

Rosencrantz (approved):

PR #3008, opened 2023-04-24
Push Docker images for tags only

tillprochaska:
We have implemented a process that uses release candidates for pre-release testing and deployments, so there’s no need anymore to push Docker images for every commit. This is wasteful in terms of bandwidth and storage resources and clutters the image repository.

@catileptic @Rosencrantz @stchris This is something we have discussed a few times before, tbh I’m not sure anymore what exactly our conclusion was. Let me know if you have concerns or in case we do actually still depend on images built from commits that are not tagged.

### DIFF
--- a/.github/workflows/build-ui.yml
+++ b/.github/workflows/build-ui.yml
@@ -37,6 +37,7 @@ jobs:
         run: make test-ui
 
       - name: Build production image
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
         run: docker build -t ghcr.io/alephdata/aleph-ui-production:${GITHUB_SHA} -f ui/Dockerfile.production ui
 
       - name: Push docker image (tagged)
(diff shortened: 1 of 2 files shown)


stchris (approved):
Totally on board with this, thanks for making this change! The build seems to fail but it must be a glitch, will trigger a re-run.

stchris (changes requested):
Oh actually the build is failing because the first push job had the `docker login` statement we need so you might need to move that down to the next one.

tillprochaska:
Ha, thanks for catching this! I opened the PR with too much confidence and then didn't bother to check the CI status 😄

stchris (approved):

tillprochaska:
Alright, thanks again for catching my mistakes. I’ve fixed the workflow and tested it to make sure it will publish images for tagged commits successfully ;)

https://github.com/alephdata/aleph/pkgs/container/aleph/88237955?tag=test-updated-workflows
https://github.com/alephdata/aleph/pkgs/container/aleph-ui-production/88238718?tag=test-updated-workflows

Rosencrantz (approved):

PR #3010, opened 2023-04-25
Enable mtls / client cert configuration for elasticsearch connections

simonwoerpel:
This PR enables client certificates / mtls for the elasticsearch connection, as it is not possible via the connection string uri as it is for all the other services.

By default (when these env vars are not set) it behaves like before


stchris (commented):
This looks good to me, @simonwoerpel ! Thanks for working on this! I would love to add a line about this to the production env docs. Do you want to do that or shall I?

simonwoerpel:
where do I find the production env docs? you mean here: https://docs.aleph.occrp.org/developers/installation/#configuration ?

stchris (approved):
Awesome contribution. Thanks @simonwoerpel ! 🙏

PR #3013, opened 2023-04-26
Update Docker image tags to use GHCR repo

tillprochaska:
Right now, UI tests build the Docker image twice because the tags do not match, causing CI runs to take 2-3 mins longer than necessary.

### DIFF
--- a/contrib/aleph-traefik-minio-keycloak/docker-compose.yml
+++ b/contrib/aleph-traefik-minio-keycloak/docker-compose.yml
@@ -65,7 +65,7 @@ services:
       - "traefik.enable=false"
 
   worker:
-    image: alephdata/aleph:${ALEPH_TAG:-3.12.2}
+    image: ghcr.io/alephdata/aleph:${ALEPH_TAG:-3.12.2}
     command: aleph worker
     restart: on-failure
     links:
@@ -90,7 +90,7 @@ services:
       - "traefik.enable=false"
 
   shell:
-    image: alephdata/aleph:${ALEPH_TAG:-3.12.2}
+    image: ghcr.io/alephdata/aleph:${ALEPH_TAG:-3.12.2}
     command: /bin/bash
     depends_on:
       - postgres
@@ -110,7 +110,7 @@ services:
       - "traefik.enable=false"
 
   api:
-    image: alephdata/aleph:${ALEPH_TAG:-3.12.2}
+    image: ghcr.io/alephdata/aleph:${ALEPH_TAG:-3.12.2}
     command: gunicorn -w 6 -b 0.0.0.0:8000 --log-level debug --log-file - aleph.wsgi:app
     expose:
       - 8000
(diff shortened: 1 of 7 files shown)


contrib/keycloak/docker-compose.dev-keycloak.yml
```diff
@@ -16,7 +16,7 @@ services:
   elasticsearch:
     build:
       context: services/elasticsearch
-    image: alephdata/aleph-elasticsearch:${ALEPH_TAG:-latest}
+    image: ghcr.io/alephdata/aleph-elasticsearch:${ALEPH_TAG:-latest}
```
tillprochaska, line 19: I think we might need to pin this to a commit hash tag (because we don’t have versioned releases for this image)?

helm/charts/aleph/values.yaml
tillprochaska, line 1: Decided to update the values here too. This should be the only file that specifies the repos, all templates just reference the values. But I’m not familiar with Helm, so maybe I’m missing something!

stchris (approved):
Overall this looks good to me, thanks for this change! I think the only comment I have is that we should update this for the `3.14.x` releases and not just for `develop` and future releases!

stchris:
> So we should merge this into main and create a new 3.14.x release with these changes?

I think this change should go into `develop` and then we should ship it with `3.14.1` (for which there is a release branch available).

Github: comment threads when?

contrib/aleph-traefik-minio-keycloak/docker-compose.yml
```diff
@@ -65,7 +65,7 @@ services:
       - "traefik.enable=false"
 
   worker:
-    image: alephdata/aleph:${ALEPH_TAG:-3.12.2}
+    image: ghcr.io/alephdata/aleph:${ALEPH_TAG:-3.12.2}
```
Rosencrantz, line 68: Whilst we are here can we update the tags to be 3.14.0?

contrib/keycloak/docker-compose.dev-keycloak.yml
Rosencrantz: Yeah, I think that we should perhaps get rid of latest and pin to a version. Might even be something that we can automate in time (we don't need to do that now though)

contrib/aleph-traefik-minio-keycloak/docker-compose.yml
tillprochaska: Right! In that case I’d also add the file to the bump2version config, ok?

I was just thinking that maybe it’s a good idea for these example configuration to be actually a "frozen" snapshot, i.e. it worked with 3.12.2 but we don’t really maintain or test these -- so it might be better to not update it at all (which includes reverting the new repo, because 3.12.2 is still available on Docker Hub).

tillprochaska:
@Rosencrantz @stchris As discussed earlier, I have made the following changes:

* Updated all references to outdated Aleph versions (3.12.2) to the latest Aleph version 3.14.0.
* Adjusted the bump2version configuration to update the Aleph version in Compose files in the `contrib/` directory.
* Fixed the image tag of the ES image in the Keycloak example.

Rosencrantz (approved):

stchris (approved):
LGTM!

tillprochaska:
Closing as all the changes are in main already, I think

PR #3014, opened 2023-04-26
Allow users to add timeline items with times

### LINKED ISSUE #3009: BUG: Timelines do not support adding datetime values
**Describe the bug**
FtM does support times for date properties (see [docs](https://followthemoney.tech/explorer/types/date/)), but when adding items to a timeline in Aleph, you can’t add a time.

Additionally, when a timeline contains an item with a datetime value, an unhandled error occurs.

**To Reproduce**
1. Create a timeline
2. Click on "Add item".
3. Try to enter "2023-01-01T00:00:00" in the "Start date" field.
4. The validation fails, the "Add" button is disabled.
5. Change start date to "2023-01-01".
6. Click "Add".
7. In the entity sidebar, change the value to "2023-01-01T00:00:00".
8. Switch to the chart view.
9. Page turns blank due to an unhandled date parsing error.

**Expected behavior**
* Entering datetime values when adding new timeline items should be possible.
* Switching to the chart view should be possible.

**Aleph version**
Latest

tillprochaska:
FtM supports times in date properties (see [docs](https://followthemoney.tech/explorer/types/date/)), but when adding items to a timeline in Aleph, you can’t add a time.

Additionally, if a timeline contains an item with a datetime value, an unhandled error occurs when switching to the chart view.

This PR fixes both issues. When adding a new timeline item, the form now allows entering values like `2023-05-02T00:00:00` or `2023-05-02 00:00`.

When parsing those dates, we now simply ignore the time part. This also ensures that the chart view renders successfully.

Closes #3009

**To do:**

- [x] Use the same date regex for HTML `pattern` attribute in form an for parsing in `ImpreciseDate`?

### DIFF
--- a/ui/src/components/Timeline/TimelineItemCreateForm.test.tsx
+++ b/ui/src/components/Timeline/TimelineItemCreateForm.test.tsx
@@ -149,6 +149,22 @@ it('supports a variety of different date formats', async () => {
   await userEvent.type(input, '2022-01-01');
   expect(input.checkValidity()).toBe(true);
 
+  await userEvent.clear(input);
+  await userEvent.type(input, '2022-01-01T00:00:00');
+  expect(input.checkValidity()).toBe(true);
+
+  await userEvent.clear(input);
+  await userEvent.type(input, '2022-01-01T00:00');
+  expect(input.checkValidity()).toBe(true);
+
+  await userEvent.clear(input);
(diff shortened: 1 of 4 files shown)


Rosencrantz (approved):
Looks fine to me. Thanks Till

PR #3052, opened 2023-05-11
Release/3.14.2

stchris:


aleph/index/xref.py
```diff
@@ -89,7 +90,13 @@ def iter_matches(collection, authz):
...
     ]
     query = {"query": {"bool": {"filter": filters}}, "_source": XREF_SOURCE}
-    for res in scan(es, index=xref_index(), query=query):
+    for res in scan(
+        es,
+        index=xref_index(),
+        query=query,
+        scroll=SETTINGS.XREF_SCROLL,
+        size=SETTINGS.XREF_SCROLL_SIZE,
+    ):
```
tillprochaska, line 99: @stchris I’m not entirely sure, but isn’t the relevant method `iter_entities`?

https://github.com/alephdata/aleph/blob/main/aleph/index/entities.py#L59

stchris: You're very much right. For consistency we might leave this in, because it also relates to xref operations.

aleph/index/entities.py
```diff
@@ -64,6 +64,8 @@ def iter_entities(
     excludes=None,
     filters=None,
     sort=None,
+    es_scroll="5m",
```
catileptic, line 67: Do you suppose, for consistency, we could reference values set in `SETTINGS` here as well? Perhaps have `ES_SCROLL` and `ES_XREF_SCROLL`?

aleph/index/xref.py
tillprochaska, line 99: My understanding of the problem hypothesis is as follows:

* In order to compute xref matches, we fetch all entities in a collection in batches of 1000 entities max.
* For every entity in a batch, we try to find matches by sending a request to ES that searches only in relevant indexes, but in all collections. That means as the amount of data in an Aleph instance increases, those requests likely take longer.
* If sending those requests to find entity matches takes longer than 5m (the default scroll value), the search context is lost when trying to fetch the next batch and the error occurs.

Is that correct? In that case, a larger scroll value (or a smaller scroll size) is only required when computing xref matches.

As far as I can see `iter_matches` is used to iterate on pre-computed xref matches (in order to export them), so it’s very similar to streaming/exporting any other set of data from ES and should work well with lower/default values.

tillprochaska: So the short version of this comment is: Shouldn’t this method use the same scroll values as the other methods that are used to export/stream entities instead of the scroll values used for computing xref matches?

stchris: You are entirely right on your assumptions. I made this change for consistency, but this made me reconsider and take this one change back, because there's no need to do this. Thanks, @tillprochaska !

PR #3056, opened 2023-05-12
Add service annotations to the helm chart

monneyboi:
When running Aleph in the google cloud with [container native loadbalancing](https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing), the healthcheck of the default `BackendConfig` google cloud generates won't work (because of the non-standard name `api` of the service port if i remember correctly.).

This can be solved in multiple ways:
- I think renaming the port to `http` would make the defaults work, but i haven't tested this.
- Making it possible for the user to set `annotations` on the service. Which seems like a good idea anyway.

### DIFF
--- a/helm/charts/aleph/templates/api.yaml
+++ b/helm/charts/aleph/templates/api.yaml
@@ -149,6 +149,12 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+{{- if .Values.api.serviceAnnotations }}
+  annotations:
+  {{- range $key, $value := .Values.api.serviceAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+{{- end }}
   name: {{ .Values.global.namingPrefix }}-api
   labels:
     app: {{ .Values.global.namingPrefix }}-api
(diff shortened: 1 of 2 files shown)


helm/charts/aleph/templates/api.yaml
```diff
@@ -149,6 +149,12 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+{{- if .Values.api.serviceAnnotations }}
```
ps-occrp, line 152: If you specify default value it will always be there, you don't need to check that it exists.

helm/charts/aleph/templates/api.yaml
```diff
@@ -149,6 +149,12 @@ spec:
...
+  annotations:
+  {{- range $key, $value := .Values.api.serviceAnnotations }}
```
ps-occrp, line 154: Can you implement it like helm does it [here](https://github.com/helm/helm/blob/main/pkg/chartutil/create.go#L262C1-L265C13)?

ps-occrp (changes requested):

PR #3089, opened 2023-05-23
Use Ruff for linting

stchris:
https://github.com/charliermarsh/ruff replaces flake8 + black (and can also auto-fix lots of things).

```
$ time make lint
ruff check .
make lint  0.01s user 0.03s system 47% cpu 0.091 total
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,11 +35,8 @@ jobs:
       - name: Install development dependencies
         run: make dev
 
-      - name: Check code formatting
-        run: make format-check
-
-      - name: Run linter
-        run: make lint
+      - name: Run linter (ruff)
+        uses: chartboost/ruff-action@v1
 
       - name: Run aleph tests
         run: make ALEPH_TAG=${GITHUB_SHA} test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,8 +35,11 @@ jobs:
       - name: Install development dependencies
         run: make dev
 
-      - name: Run linter (ruff)
-        uses: chartboost/ruff-action@v1
+      - name: Check code formatting
+        run: make format-check
+
+      - name: Run linter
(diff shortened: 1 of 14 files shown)


.github/workflows/build.yml
```diff
@@ -35,11 +35,8 @@ jobs:
       - name: Install development dependencies
         run: make dev
 
-      - name: Check code formatting
-        run: make format-check
-
-      - name: Run linter
-        run: make lint
+      - name: Run linter (ruff)
+        uses: chartboost/ruff-action@v1
```
tillprochaska, line 39: What is the advantage of using this action over just `make lint` (which will now call ruff, too)?

Makefile
```diff
@@ -27,20 +27,14 @@ test-ui:
...
 lint:
-	flake8 aleph/
+	ruff check .
 
 lint-ui:
 	$(UIDOCKER) npm run lint
 
-format:
-	black aleph/
-
```
tillprochaska, line 37: Hm, but black doesn’t replace a formatter, right?

> As a project, Ruff is designed to be used alongside Black and, as such, will defer implementing stylistic lint rules that are obviated by autoformatting.

https://beta.ruff.rs/docs/faq/#is-ruff-compatible-with-black

tillprochaska (approved):
Left two comments/questions, but in general, I like ruff and am fine with whatever tooling everyone else prefers :)

.github/workflows/build.yml
stchris: None, I guess. I will change it to be consistent, making sure we use the same version we have pinned.

Makefile
stchris: You are totally right, my bad. Reverting this.

stchris:
Thanks for your careful review and great comments. I've made the changes and adapted the corresponding `ingest-file` PR as well. 👍

Issue #2268, opened 2022-05-11
Highlight text in search results dissappears when opening the result


monneyboi:
When i search for something through the main search, i get nice results with highlights:
![image](https://user-images.githubusercontent.com/86304937/167844971-3c52a616-ccea-401a-9a12-f6c7df5727e9.png)

Though when i open one of these results, the highlight disappears:
![image](https://user-images.githubusercontent.com/86304937/167845012-7e9e3930-94d3-4f82-a3dd-d81a6d92566f.png)

Seems like an UI bug?

PR #3093, opened 2023-05-24
Fix search highlights disappearing when opening entity previews

tillprochaska:
When searching, we keep the search results in the Redux store in a structure similar to the following:

```json
{
  "entities": {
    "abc123": {
      "id": "abc123",
      "schema": "Company",
      "highlight": "<em>Twitter</em> is an online social media and social networking service."
    }
  },
  "results": {
    "entities?filter:schemata:Company&highlight=true&limit=30&q=twitter": {
      "total": 1,
      "page": 1,
      "results": ["abc123"]
    }
  }
}
```

After clicking on one of the search results, we load detailed entity data and use the detailed data to replace the (partial) entity data from the initial search request. However, the response from the second request does not contain the search highlight, i.e. the highlight is removed from the state when loading additional entity data.

We can work around this by merging data instead of replacing data. It does still feel a bit like a hacky workaround, but a proper solution would require making significant changes to our state management solution (probably replacing it with an off-the-shelf solution).

Fix https://github.com/alephdata/aleph/issues/2268

### DIFF
--- a/ui/src/app/variables.scss
+++ b/ui/src/app/variables.scss
@@ -47,7 +47,7 @@ $aleph-sidebar-width: $aleph-grid-size * 35;
 $aleph-sidebar-sm-width: $aleph-grid-size * 20;
 $aleph-sidepane-md-width: $aleph-grid-size * 27;
 
-$aleph-text-highlight-color: yellow;
+$aleph-text-highlight-color: rgba($gold5, 0.5);
 $aleph-visited-link-color: rgba($aleph-link-color, 0.6);
 $warning-border-color: #faebcc;
 $warning-background-color: #fcf8e3;
--- a/ui/src/reducers/util.js
+++ b/ui/src/reducers/util.js
@@ -31,7 +31,10 @@ export function loadComplete(data) {
 }
 
 export function objectLoadComplete(state, id, data = {}) {
-  return { ...state, [id]: loadComplete(data) };
+  const oldData = state[id] || {};
+  const newData = loadComplete(data);
+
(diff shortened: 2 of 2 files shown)


ui/src/app/variables.scss
```diff
@@ -47,7 +47,7 @@ $aleph-sidebar-width: $aleph-grid-size * 35;
 $aleph-sidebar-sm-width: $aleph-grid-size * 20;
 $aleph-sidepane-md-width: $aleph-grid-size * 27;
 
-$aleph-text-highlight-color: yellow;
+$aleph-text-highlight-color: rgba($gold5, 0.5);
```
Rosencrantz, line 50: Is this really $gold5 or is it actually $dirtyMustard?  Joking.

ui/src/reducers/util.js
```diff
@@ -31,7 +31,10 @@ export function loadComplete(data) {
 }
 
 export function objectLoadComplete(state, id, data = {}) {
-  return { ...state, [id]: loadComplete(data) };
+  const oldData = state[id] || {};
+  const newData = loadComplete(data);
+
+  return { ...state, [id]: { ...oldData, ...newData } };
```
Rosencrantz, line 37: It's difficult to see, but I would assume there is no chance of duplication or overwriting here?

Rosencrantz (approved):

ui/src/reducers/util.js
Rosencrantz: To answer my own question... The spread operator will, in the case of duplicate items in an object, take the value from the 2nd object.

PR #3094, opened 2023-05-25
Improved API key handling

tillprochaska:
This PR includes a number of changes related to API keys:

* New users do not get an API key by default anymore. Instead, they have to generate one first if they want to use the Aleph API.
* User can regenerate their API key which invalidates the old API key.
* Users can see their API key once after they’ve (re-)generated it, but not again after that.
* Users receive an email notification when an API key is (re-)generated.
* New API keys automatically expire after 90 days. Users will receive an email notification 7 days before their key will expire and another notification when it has expired.
* Existing API keys (i.e. keys generated before rolling out this change) do not expire automatically, but there is a CLI command `aleph reset-api-key-expiration` to set the expiration date for these keys.

***

Some additional information about implementation decisions:

* I’ve refactored the settings screen, both with regards to the visual appearance as well as the implementation. The settings screen was a little cluttered before, and this PR as well as a few follow-up PRs to come will add even more settings to the screen. This refactoring prepares for that.

* It’s quite common to allow users to generate multiple API keys for different purposes, and to be able to revoke/expire them individually. While this has advantages, it also adds complexity to the implementation. Assuming that most users do indeed use the API key as a personal API key to e.g. programmatically search Aleph for a data analysis etc. rather than integrating the API key into ten different external services, I’ve decided to avoid this complexity, at least for now.

***

https://github.com/alephdata/aleph/assets/1512805/98eb67ca-af97-495d-a4b8-62fc194bd879

***

**To do:**

- [x] Right now, the settings UI displays "Reset API key" even for new users who do not have an API key. The settings UI should handle that case (something along the lines of "You currently do not have an API key. Generate an API key in order to use the Aleph API.")
- [x] The API endpoint and some methods in the code base are currently named "reset_api_key" or similar. It would probably make sense to rename them to "generate_api_key" as these endpoints/methods are also used to set an initial API key.


tillprochaska:
Related idea: Is it necessary to create an API key for every user by default? Most users probably do not use the API at all. We could reuse the logic to reset to let users create an initial key on demand.

tillprochaska:
@Rosencrantz As mentioned before I still need to double-check a few technical implementation details, but let me know what you think about the UX.

tillprochaska:
I’ve converted this back to a draft as I started implementing additional changes on this branch to avoid merge conflicts. However, all changes on this branch are ready for review and I’d appreciate a review of what’s already there.

aleph/tests/test_role_model.py
```diff
@@ -98,3 +100,33 @@ def test_role_by_api_key_empty(self):
...
+            print(role_.api_key_expires_at)
+            role = Role.by_api_key(role_.api_key)
+            assert role is not None
+            assert role.id == role_.id
+
+        with time_machine.travel("2024-03-31T00:00:00Z"):
+            role = Role.by_api_key(role_.api_key)
+            assert role is None
+
+    def test_role_by_api_key_legacy_without_expiration(self):
```
Rosencrantz, line 121: Will this test have an expiry date after which it wouldn't make sense to keep it any longer?

aleph/migrate/versions/d46fc882ec6b_api_key_expiration.py
```diff
@@ -0,0 +1,30 @@
...
+
+Revision ID: d46fc882ec6b
+Revises: c52a1f469ac7
+Create Date: 2024-05-02 11:43:50.993948
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
```
Rosencrantz, line 11: What are revision identifiers used for, something to do with upgrade tasks for the DB?

aleph/logic/api_keys.py
```diff
@@ -0,0 +1,108 @@
...
+    role.api_key_expiration_notification_sent = None
+
+    db.session.add(role)
+    db.session.commit()
+    update_role(role)
+
+    return role.api_key
+
+
+def send_api_key_expiration_notifications():
```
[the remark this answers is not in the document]
tillprochaska: This method sends notifications to users with an API key that will expire soon (in less than 7 days) or that has expired recently.

But you were probably asking because this function isn’t called anywhere right now and indeed I did forget to push one additional change that’s required to execute this daily.

aleph/logic/api_keys.py
```diff
@@ -0,0 +1,108 @@
...
+        }
+        plain = render_template(plain_template, **params)
+        html = render_template(html_template, **params)
+        email_role(role, subject, html=html, plain=plain)
+
+    query.update({Role.api_key_expiration_notification_sent: days})
+    db.session.commit()
+
+
+def reset_api_key_expiration():
```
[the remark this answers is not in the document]
tillprochaska: This is a good point! There are two aspects:

1. From a technical perspective, after rolling out this change, the API keys of all users will expire at the same time, and thus they will also all receive a notification at the same time. We’re already sending notification digests via email on a daily basis, and it usually takes approx. 1min per 1,500 users. Based on this and our total number of users, I figured it would probably be fine from a technical perspective. The only unknown here would be potential daily sending limits of our email provider.

   I had initially implemented a jitter of 0 to 20 days for the API key expiration, so that the expiration notifications where distributed across 20 days rather then sent out all on a single day. It added some complexity, and I initially didn’t feel that it’s worth it. But I will add it again just to be on the safe side.

2. From a user support perspective, I was planning to inform our internal teams that usually handle support requests from users so they are aware of this change and to display a dismissible message banner. I also tried to proactively avoid confusion by adding some context to the notification emails. Nonetheless, there will be some users who are confused by this. I’m honestly not sure if it will be helpful in this regard or cause more confusion if we send out the notification emails over a longer period or to send them out over multiple days. Do you have an opinion?

In any case, you are correct that this will be a one-time thing. Once legacy API keys have expired, users will have to manually regenerate keys if they want to keep using the API. This will automatically cause new expiration dates to vary depending on when users regenerate their keys. Also, it’s likely that only a small subset of users actually regenerates their keys.

aleph/templates/email/api_key_expired.html
```diff
@@ -0,0 +1,15 @@
...
+{% block content -%}
+  <p>
+    {% trans expires_at=(expires_at | datetimeformat) -%}
+    Your API key has expired on {{expires_at}} UTC.
+    {%- endtrans %}
+  </p>
+
+  <p>
+    {% trans settings_url=settings_url -%}
+    If you do not use the Aleph API you can ignore this email. If you use the Aleph API you will need to <a href="{{settings_url}}">regenerate your API key</a> and update any applications or scripts that use your API key.
```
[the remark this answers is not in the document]
tillprochaska: Yes, the API can still be used without authentication in order to access public data. Do you think updating the wording as follows would clarify that sufficiently?

> If you do not use the Aleph API you can ignore this email. If you use the Aleph API to access data that isn’t public you will need to regenerate your API key and update any applications or scripts that use your API key.

aleph/tests/test_role_model.py
tillprochaska: Yes, although it should probably not be replaced at that point and instead simply be updated to ensure that any keys that do not have an expiration date are always invalid or something like that. I’ll add a comment to the code and open an issue to keep track of this.

aleph/logic/api_keys.py
```diff
@@ -0,0 +1,108 @@
...
+    now = datetime.datetime.utcnow().replace(microsecond=0)
```
tillprochaska, line 95: Replace deprecated `utcnow`

aleph/views/serializers.py
```diff
@@ -118,19 +118,20 @@ def _serialize(self, obj):
...
             obj.pop("is_admin", None)
             obj.pop("is_muted", None)
             obj.pop("is_tester", None)
             obj.pop("is_blocked", None)
-            obj.pop("api_key", None)
             obj.pop("email", None)
             obj.pop("locale", None)
             obj.pop("created_at", None)
             obj.pop("updated_at", None)
         if obj["type"] != Role.USER:
```
tillprochaska, line 131: Also pop `has_api_key` and `api_key_expires_at` in this case. Probably that’s not strictly necessary, but it’s better.

aleph/logic/api_keys.py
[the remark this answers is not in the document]
tillprochaska: Done

aleph/logic/api_keys.py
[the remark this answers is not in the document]
Rosencrantz: Thanks for the response and the insights. Don't introduce complexity if you don't feel it is required. I trust your judgement on this, and it's enough for me that you've evidently thought things though.

aleph/templates/email/api_key_expired.html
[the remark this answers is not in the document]
Rosencrantz: If you do not use the Aleph API, or only use the API to access public information you can ignore this email. If you use the Aleph API to access information that is not publicly accessible then you'll need to regenerate your API key in order to keep accessing the protected information.

ui/src/components/Settings/ApiKeySettings.tsx
```diff
@@ -113,11 +118,24 @@ export default function ApiKeySettings() {
           />
         </h4>
 
+        <p>
+          {expiresAt && (
+            <FormattedMessage
+              id="settings.api_key.text_expiration"
+              defaultMessage="Your API key will expire on {date}, {time}."
```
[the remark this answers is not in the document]
Rosencrantz: Yes, I'd probably go with just the date, but I trust your judgement and I don't feel strongly one way or the other.

aleph/templates/email/api_key_expired.html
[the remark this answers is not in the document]
tillprochaska: Just as a note for my future self: I double checked one detail about the Aleph API. When sending a request with an invalid API key, we simply ignore the API key and treat the request as unauthenticated.

This isn’t super intuitive and looking at the code it doesn’t seem like it was implemented like this intentionally. Imo it would be better to respond with an authentication error if an invalid API key is supplied, even if authentication isn’t required for a request.

But I’ve kept this behavior for backwards compatibility, and it’s a lucky coincidence that it helps with the messaging in this case.

ui/src/components/Settings/ApiKeySettings.tsx
```diff
@@ -0,0 +1,231 @@
...
+      title={title}
+      isOpen={isOpen}
+      onClose={onClose}
+      canOutsideClickClose={false}
+    >
+      <DialogBody>
+        <p>
+          <FormattedMessage
+            id="settings.api_key.success"
+            defaultMessage="Your new API key has been generated. Be sure to copy it below as you won’t be able to view it again later. We recommend storing your API key in a password manager. Do not share your API key with anyone else."
```
tillprochaska, line 219: Maybe clarify that the key expires after 90 days.

ui/src/components/Settings/ApiKeySettings.tsx
```diff
@@ -0,0 +1,231 @@
...
+      <Card elevation={1}>
+        <h4 className="bp4-heading">
+          <FormattedMessage
+            id="settings.api_key.title"
+            defaultMessage="API key"
+          />
+        </h4>
+
+        <p>
+          {expiresAt && (
```
tillprochaska, line 122: This currently only handles API keys that will expire in the future. There should be a separate case that handles API keys that have expired.

**Case 1:** User doesn’t have an API key:

> You need an API key in order to access the Aleph API. Generate an API key to get started."

**Case 2:** User has an API key that hasn’t yet expired:

> Your API key will expire on {date}. You can regenerate your API key at any time. When you regenerate your API key, your old key will stop working.

**Case 3:** User has an expired API key.

> Your API key expired on {date}. Regenerate your API key to continue using the Aleph API.

ui/src/components/Settings/PasswordSettings.jsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const intl = useIntl();
+  const successMessage = intl.formatMessage({
+    id: 'settings.password.success',
+    defaultMessage: 'Your password has been updated.',
+  });
+
+  const isValidCurrentPassword = currentPassword !== '';
+  const isValidNewPassword = newPassword && newPassword.length >= 6;
+  const isValidPasswordConfirmation = newPassword == passwordConfirmation;
```
tillprochaska, line 42:
```suggestion
  const isValidPasswordConfirmation = newPassword === passwordConfirmation;
```

aleph/templates/email/api_key_generated.html
```diff
@@ -0,0 +1,13 @@
+{% extends "email/layout.html" %}
+
+{% block content -%}
+{% if event == "regenerated" -%}
+{% trans -%}
+Your API key has been regenerated. If that wasn’t you, please contact an administrator.
```
tillprochaska, line 6:
```suggestion
Your Aleph API key has been regenerated. If that wasn’t you, please contact an administrator.
```

aleph/templates/email/api_key_generated.html
```diff
@@ -0,0 +1,13 @@
...
+{%- endtrans %}
+{% else -%}
+{% trans -%}
+An API key has been generated for your account. If that wasn’t you, please contact an administrator.
```
tillprochaska, line 10:
```suggestion
An Aleph API key has been generated for your account. If that wasn’t you, please contact an administrator.
```

aleph/templates/email/api_key_generated.txt
```diff
@@ -0,0 +1,13 @@
+{% extends "email/layout.txt" %}
+
+{% block content -%}
+{% if event == "regenerated" -%}
+{% trans -%}
+Your API key has been regenerated. If that wasn’t you, please contact an administrator.
```
tillprochaska, line 6:
```suggestion
Your Aleph API key has been regenerated. If that wasn’t you, please contact an administrator.
```

aleph/templates/email/api_key_expires_soon.txt
```diff
@@ -0,0 +1,11 @@
+{% extends "email/layout.txt" %}
+
+{% block content -%}
+{% trans expires_at=(expires_at | datetimeformat) -%}
+Your API key will expire in 7 days, on {{expires_at}} UTC.
```
tillprochaska, line 5:
```suggestion
Your Aleph API key will expire in 7 days, on {{expires_at}} UTC.
```

aleph/templates/email/api_key_expires_soon.html
```diff
@@ -0,0 +1,15 @@
+{% extends "email/layout.html" %}
+
+{% block content -%}
+  <p>
+    {% trans expires_at=(expires_at | datetimeformat) -%}
+    Your API key will expire in 7 days, on {{expires_at}} UTC.
```
tillprochaska, line 6:
```suggestion
    Your Aleph API key will expire in 7 days, on {{expires_at}} UTC.
```

aleph/migrate/versions/d46fc882ec6b_api_key_expiration.py
stchris, line 1: I have some migration related weirdness when testing this branch. If I run `make upgrade` I get 

``` Error: Can't locate revision identified by '8adf50aadcb0' [flask_migrate]```

but I don't really know where that revision identifier is coming from

tillprochaska: This is the migration: https://github.com/alephdata/aleph/blob/develop/aleph/migrate/versions/8adf50aadcb0_add_last_login_at_column.py

You’ve probably run this migration before in your development environment, but it isn’t available on the branch for this PR.

Looking at the migration, in this case, you can safely set the `alembic_version` in your dev database to `c52a1f469ac7` (then reset it to `8adf50aadcb0` after your done testing).

(Alternatively, you can of course also check out the `develop` branch, downgrade the migration, then check out the PR again, but that’s a little more complicated.)

stchris: Thank you, Till! That worked!

stchris (approved):
Awesome effort, Till! Seems great to me

catileptic (approved):

PR #3168, opened 2023-06-21
Derive "safeHtml" from all "bodyHtml" values

### LINKED ISSUE #3163: BUG: `safeHtml` is derived only from first value of `bodyHtml`
**Describe the bug**
The `safeHtml` property in responses of the entities API endpoint is derived from the first value of the `bodyHtml` property only. The `safeHtml` property is used to render an HTML preview of emails and webpages in the Aleph UI.

In case of multipart emails, this leads to incomplete email previews.

**To Reproduce**
1. Navigate to example entity 1 from #3123 
2. Check `safeHtml` in the API response. It is incomplete while the `bodyHtml` property has two values.

**Expected behavior**
`safeHtml` should be derived from all values of the `bodyHtml` property.

**Aleph version**
Latest

tillprochaska:
Fixes #3163. Test is still failing because of https://github.com/alephdata/followthemoney/issues/1139.


aleph/views/entities_api.py
```diff
@@ -305,10 +305,12 @@ def view(entity_id):
     entity = get_index_entity(entity_id, request.authz.READ, excludes=excludes)
     tag_request(collection_id=entity.get("collection_id"))
     proxy = model.get_proxy(entity)
-    html = proxy.first("bodyHtml", quiet=True)
+    html = proxy.get("bodyHtml", quiet=True)
     source_url = proxy.first("sourceUrl", quiet=True)
     encoding = proxy.first("encoding", quiet=True)
-    entity["safeHtml"] = sanitize_html(html, source_url, encoding=encoding)
+    entity["safeHtml"] = [
+        sanitize_html(value, source_url, encoding=encoding) for value in html
```
Rosencrantz, line 312: This is a utility function that we wrote, or a third party created?

tillprochaska: This is a utility we wrote wrapping the Cleaner class from lxml:

* https://github.com/alephdata/aleph/blob/main/aleph/logic/html.py
* https://lxml.de/api/lxml.html.clean.Cleaner-class.html

stchris (approved):
Looks good to me!

PR #3169, opened 2023-06-22
Make code reloading work

stchris:
@monneyboi reported that the reloading feature of `flask run` wasn't properly working. I'm pretty sure it has to do with the debugger we are trying to attach. In either case [enabling earlier fails](https://flask.palletsprojects.com/en/2.1.x/server/#lazy-or-eager-loading) (instead of trying to fail lazily so a debugger can be shown) fixed the code reloading issue.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -104,7 +104,7 @@ services:
     build:
       context: .
     image: ghcr.io/alephdata/aleph:${ALEPH_TAG:-latest}
-    command: python3 -m debugpy --listen 0.0.0.0:5678 -m flask run -h 0.0.0.0 -p 5000 --with-threads --reload --debugger
+    command: python3 -m debugpy --listen 0.0.0.0:5678 -m flask run -h 0.0.0.0 -p 5000 --with-threads --eager-loading --reload --debugger
     ports:
       - "127.0.0.1:5000:5000"
       - "127.0.0.1:5678:5678"
@@ -124,6 +124,7 @@ services:
       ALEPH_PROFILE: "false"
       ALEPH_SECRET_KEY: "development"
       FLASK_APP: "aleph.wsgi"
+      FLASK_ENV: "development"
     env_file:
       - aleph.env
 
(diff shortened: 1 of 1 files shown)


docker-compose.dev.yml
```diff
@@ -124,6 +124,7 @@ services:
       ALEPH_PROFILE: "false"
       ALEPH_SECRET_KEY: "development"
       FLASK_APP: "aleph.wsgi"
+      FLASK_ENV: "development"
```
monneyboi, line 127: The fix also works without `FLASK_ENV` here. Also, `FLASK_ENV` [is deprecated](https://github.com/pallets/flask/issues/4714)

monneyboi (changes requested):

tillprochaska:
@stchris Following up on our discussion earlier today, here’s the behavior I observed:

When running `make web` / `docker compose -f ... up`:

1. Introduce a syntax error and save file
2. Flask dev server reloads code
3. Dev server exits due to the syntax error
4. Docker tries to restart the container
5. Dev server exists again
6. Infinite container restart loop

When using `docker compose run -f ... api`:

1. Introduce a syntax error and save file
2. Flask dev server reloads code
3. Dev server exists due to the syntax error
4. I need to manually restart the container

Maybe there is another solution to this problem that doesn’t caveats?

tillprochaska:
Also, is anyone here actually using debugpy? If this only happens when used together with debugpy, we might want to consider simply removing debugpy from the default development setup?

tillprochaska:
According to https://github.com/microsoft/debugpy/issues/808 disabling only Flask’s debugger might help (if we want to keep support for debugpy)?

monneyboi:
> Also, is anyone here actually using debugpy? If this only happens when used together with debugpy, we might want to consider simply removing debugpy from the default development setup?

This is exactly what i was thinking

docker-compose.dev.yml
stchris: Done

monneyboi (approved):

PR #3171, opened 2023-06-22
Add plain highlighting for documents and entities

monneyboi:
This is a alternative to https://github.com/alephdata/aleph/pull/2416

Implementation of highlights for search that uses the `plain` highlighter and does not change existing indices.

I removed some seemingly unused logic. `number_of_fragments` seems to be used by the [PdfViewer](https://github.com/alephdata/aleph/blob/62d089932b828f702209c9822926fd3879be2e2a/ui/src/viewers/PdfViewer.jsx#L288), but i'm not sure when this logic is called / if it has to stay.


aleph/search/parser.py
```diff
@@ -132,27 +132,7 @@ def __init__(self, args, authz, limit=None):
...
-        # request is bounded by the value defined in the
-        # index.highlight.max_analyzed_offset setting (1000000 by default),
-        # and when the number of characters exceeds this limit an error is
-        # returned. By setting `max_analyzed_offset` to a non-negative value
-        # lower than `index.highlight.max_analyzed_offset`, the highlighting
-        # stops at this defined maximum limit, and the rest of the text is not
-        # processed, thus not highlighted and no error is returned.
-        self.max_highlight_analyzed_offset = self.getint(
-            "max_highlight_analyzed_offset", 999999
-        )
```
tillprochaska, line 155: These settings are still relevant, even when using a different highlighter?

For example `highlight_text` is required to be able to highlight the search query in the plaintext viewer:

* Search within a multi-page document
* Go to one of the pages containing the search query
* Switch to the plaintext view, the search query is highlighted
* Reload the page
* Highlight is not visible anymore

And I guess `max_highlight_analyzed_offset` is necessary to protect against errors for large documents.

aleph/search/query.py
```diff
@@ -217,18 +216,12 @@ def get_sort(self):
         return list(reversed(sort_fields))
 
     def get_highlight(self):
-        if not self.parser.highlight:
-            return {}
```
tillprochaska, line 221: Is there a particular reason to remove the ability to explicitly enable/disable highlighting? Especially when using the Aleph API directly, it’s often not necessary to return highlights, but computing highlights can consume lots of resources. For example when streaming thousands of entities for offline processing, this could add a lot of unnecessary load on the server.

aleph/settings.py
```diff
@@ -132,9 +132,6 @@ def __init__(self) -> None:
         # Document processing pipeline
         self.INGEST_PIPELINE = env.to_list("ALEPH_INGEST_PIPELINE", ["analyze"])
 
-        # Result high-lighting
-        self.RESULT_HIGHLIGHT = env.to_bool("ALEPH_RESULT_HIGHLIGHT", True)
-
```
tillprochaska, line 137: Is there a particular reason for removing this global configuration option? Some Aleph admins might prefer to disable highlighting altogether due to high resource usage.

tillprochaska (changes requested):
Thanks for working on this (again)! Unfortunately, I wasn’t able to get this to work locally. I still can’t see highlights for documents on the search results pages.

Based on my understanding of the ES docs on highlighting, I think no matter what highlighter you use, the text needs to be retrievable from the source field for ES to be able to compute highlights?

If it works for you locally and you previously #2416, you’re index configuration might still be different?

I’ve also added a few code comments because I think some of the deletions will break other features…

tillprochaska:
I think there might be another way similar to #2416 to enable highlights for documents that works for any documents that are (re-)ingested after the change, even without rebuilding the entire index.

The reason why highlighting currently doesn’t work for documents is that we store the full document text in the `Pages:indexText` property. The `indexText` property is a special property. It’s contents are copied to the `text` field in Elasticsearch, but it’s not stored separately, i.e. the contents of the `indexText` property are searchable, but they cannot be retrieved or used for highlighting.

The reason to exclude the full contents from the source field are:

* We do not want to retrieve the full contents of a document when returning the document in a response to a request to `/entities?q=...`.
* We already store the full text contents in the `Page:bodyText` properties. (In addition to the `Pages` entity, one `Page` entity is stored per page in the document.)

#2416 fixes highlights for documents by including the `text` field (which includes the contents of the `Pages:indexText` property). This has a few side effects:

* The `text` field also includes a copy of all other entity properties in order to make them full-text searchable. Storing the `text` field verbatim in `_source` will increase disk usage (even though for most properties this is unnecessary).
* It requires a change to the index mapping which requires recreating the index to take effect (not 100% if this true for the `_source` mapping).

The `indexText` property already is slightly hacky and I appreciate that this would make it even more hacky, but we could store the full contents in a separate ES field (or FtM property) that is searchable and stored as part of the `_source` field, but is never retrieved from ES or returned in API responses.

PR #3180, opened 2023-06-27
Plain search highlights

monneyboi:
Most basic changes i can think of that lead to having search highlights for documents and entities. I tried to not remove any logic that could potentially be used.

From the previous PR:
> I wasn’t able to get this to work locally. I still can’t see highlights for documents on the search results pages.

> Based on my understanding of the ES docs on highlighting, I think no matter what highlighter you use, the text needs to be retrievable from the source field for ES to be able to compute highlights?

Yeah i was surprised by this aswell, but these changes lead to highlights on my end (i am using a newly setup development environment based on `make upgrade` and `make `web`), what could differ between our setups?

### DIFF
--- a/aleph/search/__init__.py
+++ b/aleph/search/__init__.py
@@ -62,7 +62,6 @@ class CollectionsQuery(Query):
 class EntitiesQuery(Query):
     TEXT_FIELDS = ["fingerprints.text^3", "text"]
     PREFIX_FIELD = "fingerprints.text"
-    HIGHLIGHT_FIELD = "properties.*"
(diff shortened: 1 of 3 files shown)


Rosencrantz:
@monneyboi Did you investigate further?

Rosencrantz:
@monneyboi Is this PR still alive?

Rosencrantz:
@etymono ???

monneyboi:
I missed this one somehow. No this PR can be closed, working on a alternative solution.

Issue #3161, opened 2023-06-20
BUG: ingest-file doesn’t properly handle multipart/mixed emails


tillprochaska:
**Describe the bug**
There are two different types of multipart emails:

* **multipart/mixed**: Intended for use when the body parts are independent and intended to be displayed serially
* **multipart/alternative**: Each of the parts is an "alternative" version of the same information. User agents should recognize that the content of the various parts are interchangeable. The user agent should either choose the "best" type based on the user's environment and preferences, or offer the user the available alternatives.

[RFC 1341 standard / source](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)

ingest-file basically assumes that all emails are `multipart/alternative` emails. This is however not the case and leads to incomplete previews.

**To Reproduce**
1. Navigate to the example entities 1 and 2 in https://github.com/alephdata/aleph/issues/3123.
2. Check the API response for these entities, they include all email contents. However, the contents of the HTML parts are present only in the `bodyHtml` property and the contents of the plaintext parts are present only in the `bodyText` property.
3. We need to display all parts (HTML *and* plaintext) in order. Because HTML and plaintext parts are stored in separate properties, there’s no way to reconstruct the original order in the UI.

**Expected behavior**
There should be a way to retrieve all relevant parts of an email in the correct order so we can display a complete preview in the UI.

Two possible approaches, although there might other (better) solutions:

1. When we ingest multipart/mixed emails, automatically create alternative versions in HTML or plaintext format.

   * In case of a plaintext part, append it to the `bodyText` property, but also append the HTML-encoded value to the `bodyHtml` property.
   * In case of an HTML part, append it to the `bodyHtml` property, but also strip all HTML markup and append the remaining text to the `bodyText` property.
 
   This will ensure that all contents are present in both, the `bodyHtml` and the `bodyText` property.

2. Introduce a new property (named `normalizedParts` or similar) to the `Email` FtM schema that can be used to store all parts in the original order:

   * In case of a plaintext part, append it to the `bodyText` property, but also append the HTML-encoded value to the `normalizedParts` property.
   * In case of an HTML part, append it to the `bodyHtml` and `normalizedParts` properties.

   This will ensure that all contents are present in HTML format in the `normalizedParts` property. The `bodyText` and `bodyHtml` properties would still contain only parts of the respective type as present in the original email.

**Aleph version**
Latest

tillprochaska:
[...]

PR #3186, opened 2023-06-28
Feature/3123 switch email type

tillprochaska:
This PR adds an option to switch between HTML and plaintext view. As can be seen in the screencast below, this is not a perfect solution and contents might be missing from one of the two views (which will only be resolved after fixing https://github.com/alephdata/aleph/issues/3161).

https://github.com/alephdata/aleph/assets/1512805/d6851307-cca2-48cb-9ed1-04df67de1fa0

I’m also not entirely sure about the UI, it’s not ideal that the switch is integrated into the email view/into the data. I had previously implemented a version that displayed a separate tab at the top (similar to how we display a view and plaintext tab for text documents).

Screen Shot 2023-06-28 at 22 42 19

These files can be used for testing, although you will need to change the extension from `.eml.txt` to just `.eml` (but GitHub doesn’t allow uploading `.eml` files).

- [alternative.eml.txt](https://github.com/alephdata/aleph/files/11899793/alternative.eml.txt)
- [empty.eml.txt](https://github.com/alephdata/aleph/files/11899794/empty.eml.txt)
- [html.eml.txt](https://github.com/alephdata/aleph/files/11899795/html.eml.txt)
- [mixed.eml.txt](https://github.com/alephdata/aleph/files/11899796/mixed.eml.txt)
- [nested.eml.txt](https://github.com/alephdata/aleph/files/11899797/nested.eml.txt)
- [plain.eml.txt](https://github.com/alephdata/aleph/files/11899798/plain.eml.txt)

### DIFF
--- a/contrib/email/generate_test_emails.py
+++ b/contrib/email/generate_test_emails.py
@@ -1,0 +1,116 @@
+# This script generates minimal RFC2822 email messages to test how Aleph ingests
+# and displays different types of emails
+import sys
+from email.message import EmailMessage
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+sender = "john.doe@example.org"
+recipient = "jane.doe@example.org"
+
+def plaintext_only():
+	"""Generate a plaintext-only email."""
+	msg = EmailMessage()
+	msg["From"] = sender
+	msg["To"] = recipient
+	msg["Subject"] = "Plaintext only"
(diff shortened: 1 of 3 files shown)


tillprochaska:
Closing this because a proper fix is already on its way in https://github.com/alephdata/ingest-file/pull/488. Switching between HTML and plaintext views doesn’t provide much additional value once email previews are fixed.

PR #3205, opened 2023-07-04
Clean dev environment

monneyboi:
I propose to use default ports for services, and to drop dejavu from the development setup as management software/gui would be up for personal preference.

I also removed the `hostname` params as they were the same as the defaults, plus bumped postgres to version 14.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -2,10 +2,9 @@ version: "3.2"
 
 services:
   postgres:
-    image: postgres:10.0
-    hostname: postgres
+    image: postgres:14
     ports:
-      - "127.0.0.1:15432:5432"
+      - "127.0.0.1:5432:5432"
     volumes:
       - postgres-data:/var/lib/postgresql/data
     environment:
@@ -15,7 +14,6 @@ services:
 
   elasticsearch:
     image: ghcr.io/alephdata/aleph-elasticsearch:3bb5dbed97cfdb9955324d11e5c623a5c5bbc410
-    hostname: elasticsearch
     restart: on-failure
     environment:
       - discovery.type=single-node
@@ -26,17 +24,9 @@ services:
       - "http.cors.allow-headers=*"
       - "http.cors.allow-credentials=true"
     ports:
-      - "127.0.0.1:19200:9200"
+      - "127.0.0.1:9200:9200"
     volumes:
       - elasticsearch-data:/usr/share/elasticsearch/data
-
-  dejavu:
-    image: appbaseio/dejavu:3.6.0
-    container_name: dejavu
-    ports:
-      - '1358:1358'
-    depends_on:
-      - elasticsearch
 
   redis:
     image: redis:alpine
@@ -46,7 +36,6 @@ services:
 
   ingest-file:
     image: ghcr.io/alephdata/ingest-file:3.19.1
-    hostname: ingest
     tmpfs: /tmp
     volumes:
       - archive-data:/data
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -2,7 +2,7 @@ version: "3.2"
 
 services:
   postgres:
-    image: postgres:14
+    image: postgres:15
     ports:
       - "127.0.0.1:5432:5432"
     volumes:
--- a/docs/src/pages/developers/technical-faq/index.mdx
+++ b/docs/src/pages/developers/technical-faq/index.mdx
@@ -256,16 +256,16 @@ psql $ALEPH_DATABASE_URI
 
 The same can be done if you run an instance of the `shell` container in production mode.
 
-## How can i connect to ElasticSearch directly?
+## How can I connect to ElasticSearch directly?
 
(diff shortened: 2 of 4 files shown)


docker-compose.dev.yml
```diff
@@ -2,10 +2,9 @@ version: "3.2"
 
 services:
   postgres:
-    image: postgres:10.0
-    hostname: postgres
+    image: postgres:14
     ports:
-      - "127.0.0.1:15432:5432"
+      - "127.0.0.1:5432:5432"
```
tillprochaska, line 7: I’m on macOS, but I’ve seen some people using certain Linux distros run into problems when using default ports, because they’d also have for example a postgres server running on their host machine already by default. This might be the reason why it was using the non-default ports.

docker-compose.dev.yml
```diff
@@ -26,18 +24,10 @@ services:
...
       - elasticsearch-data:/usr/share/elasticsearch/data
 
-  dejavu:
-    image: appbaseio/dejavu:3.6.0
-    container_name: dejavu
-    ports:
-      - '1358:1358'
-    depends_on:
-      - elasticsearch
-
```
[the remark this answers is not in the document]
tillprochaska: @monneyboi What would be your preferred option?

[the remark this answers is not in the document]
monneyboi: > how about just removing it from the Makefile here? 

It's already gone there in development branch.

> What would be your preferred option?

I would remove it, as i use elasticvue myself, in my opinion admin tools like management guis / profilers etc don't have a place in the compose file.

docker-compose.dev.yml
monneyboi: I don't think a lot of linux distro's have installed postgres by default. And if people are running postgres without them knowing then making them conscious of that is not a bad thing right?

stchris: I'm ok with the port change, even if I find it unnecessary.

docker-compose.dev.yml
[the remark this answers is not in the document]
stchris: I agree that this tooling should not be present by default but documented. Are you able to update the docs then?

docker-compose.dev.yml
Rosencrantz: I like the idea of utilizing standard ports as much as possible.

docker-compose.dev.yml
```diff
@@ -2,10 +2,9 @@ version: "3.2"
 
 services:
   postgres:
-    image: postgres:10.0
-    hostname: postgres
+    image: postgres:14
```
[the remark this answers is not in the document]
Rosencrantz: Ditto. 15 has been around for about 9 months, so should be fairly stable.

Rosencrantz (approved):

stchris (changes requested):
Two items pending from my point of view:

- [x] use postgres 15
- [x] update the docs with the ES UI tooling

Let me know if you want me to do those or if you need any help @monneyboi

monneyboi:
I added some documentation about how to approach elasticsearch and updated postgres to major version 15.

docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -256,6 +256,20 @@ psql $ALEPH_DATABASE_URI
 
 The same can be done if you run an instance of the `shell` container in production mode.
 
+## How can i connect to ElasticSearch directly?
```
stchris, line 259:
```suggestion
## How can I connect to ElasticSearch directly?
```

docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -256,6 +256,20 @@ psql $ALEPH_DATABASE_URI
...
+
+There's several ES GUI's out there that you can use to query ElasticSearch. You
```
stchris, line 261:
```suggestion
There are several ElasticSearch GUIs out there that you can use to query ElasticSearch. You
```

docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -256,6 +256,20 @@ psql $ALEPH_DATABASE_URI
...
+could for example use [Dejavu](https://github.com/appbaseio/dejavu):
+```bash
+docker run -p 1358:1358 -d appbaseio/dejavu
+open http://localhost:1358/
+```
+
+Where you can connect to the ElasticSearch container that has been exposed on `http://localhost:9200`.
```
stchris, line 268:
```suggestion
Then you can connect Dejavu to the ElasticSearch container that has been exposed on `http://localhost:9200`.
```

stchris (changes requested):
Left some very minor suggestion to the documentation updates. Looks good otherwise!

monneyboi:
It occurred to me that there were some references left in the docs regarding ports, fixed those. 

While doing that i noticed that there were multiple entries for connecting to database / elasticsearch. I merged the entries from the "technical FAQ" into these in the "developer setup" as they contained the same information. Of the two, developer setup seemed to be the right place.

PR #3216, opened 2023-07-17
Metrics

tillprochaska:
This PR exposes metrics in the Prometheus format. The PR also includes changes to the Aleph Helm chart to make it easy to collect these metrics, e.g. when using the prometheus-operator.

**Related PRs:**
* https://github.com/alephdata/servicelayer/pull/111
* https://github.com/alephdata/ingest-file/pull/550
* https://github.com/alephdata/servicelayer/pull/129
* https://github.com/alephdata/servicelayer/pull/130

**Implemented metrics:**
- [x] System information (Aleph version, FtM version)
- [x] API
  - [x] HTTP request metrics (duration, count by endpoint, status)
  - [x] Streamed entities (count)
- [x] Platform information (Python version)
- [ ] Users
  - [x] Users (total)
  - [ ] Blocked users (total)
  - [ ] Active users (total by period)
  - [x] Authentication attemps (failed, succesful)
  - [ ] Signups (count)
- [x] Features
  - [x] Bookmarks (total)
  - [x] Entity sets (total by entity set type)
  - [x] Users with at least one bookmark (total)
  - [x] Users with at least one entity set (total by entity set type)
  - [x] Users with at least one investigation created (total)
  - [ ] Users with at least one file uploaded (total)
- [ ] Collections
  - [x] Number of collections (total by category)
  - [ ] Collections with at least one file uploaded (total by category)
  - [ ] Collections with at least one non-file entity (total by category)
- [x] Queues
  - [x] Queued tasks (total by stage, status) 
- [x] Workers 
  - [x] Tasks processed (count by success/failed, stage, retries)
  - [x] Task duration (seconds by stage)
- [x] Ingest
  - [x] Tasks processed (count by success/failed, ingestor)
  - [x] Tasks duration (seconds by ingestor)
  - [x] Bytes processed (bytes by ingestor)
  - [x] PDF cache hits/misses
- [ ] Xref
   - [x] Number of xref'ed entities and mentions
   - [x] Number of matches per entity/matches
   - [x] Duration of ES candidates query (query processing time incl. & exlc. of network/serialization/… overhead)
   - [ ] Duration to process a single scroll batch of entities

**Testing:**
Follow these steps to test the changes locally:

* Set `PROMETHEUS_ENABLED=true` in your `aleph.env` file and restart Aleph.

* You can view metrics exposed by the API at `http://localhost:9100` (in the container).

* Run `docker compose -f docker-compose.dev.yml run --rm api bash`.

* Run `gunicorn --bind 0.0.0.0:9100 aleph.metrics.exporter:app` to start the metrics exporter.

* You can view the metrics exposed by the exporter at `http://localhost:9100` within the container you just started.

**Implementation considerations:**

* Our API service is served using Gunicorn with multiple worker processes. Prometheus requires using a special multiprocess mode for this scenario which has a few limitations, e.g. using custom collectors are not supported out of the box. We’re using custom collectors to expose metrics that are not collected during runtime of the API or workers, but instead are exposed based on data fetched from the SQL database or Redis (e.g number of tasks in queues or number of users in the database).

  There is a way to make custom collectors work while using the multiprocess mode at the same time, but it adds some complexity to the setup (see https://github.com/prometheus/client_python/discussions/943), so I eventually decided to run a Prometheus exporter as a separate sidecar service to expose custom collector metrics. This also makes it very easy to scrape API runtime metrics at a different interval than custom collector metrics.

* API metrics are exposed on a separate port (and not at the same port as the API) to ensure metrics are not exposed publicly.

* [`prometheus_flask_exporter`](https://github.com/rycus86/prometheus_flask_exporter) is a third-party exporter to expose Prometheus metrics about a Flask web application. I’m not using it though, as it’s actually quite straight forward to collect basic HTTP request metrics and I’ve found a direct integration easier when adjust to our needs and more transparent when debugging (vs. using an external extension like `prometheus_flask_exporter`).

**To do:**

- [x] Delete contents of Gunicorn multiproc directory when (re-)starting the `api` service, create directory if it doesn’t exist.
- [x] Check whether multiproc directory is set on application startup
- [x] Adjust Gunicorn config to mark process as dead when worker process exits
- [x] Opt-in to Prometheus metrics using configuration option
- [x] Tests?
- [x] Adjust Helm charts (and test the changes locally?)
- [x] Add some inline documentation for the metrics modules
- [x] Release a new servicelayer version
- [x] Upgrade servicelayer in ingest-file
- [x] Upgrade servicelayer in Aleph
- [ ] Release ingest-file
- [ ] Upgrade ingest-file in Aleph
- [x] Add xref metrics


Rosencrantz:
Thanks Till. An amazing writeup as usual. Just looking through the list of information you've managed to start collecting is really impressive. 

One thing that I'd like your input on, because I think it may be something that could be useful for us is looking at inactivity. A few examples:

- Inactive users (anyone that hasn't logged in or used Aleph for a period of time)
- Inactive datasets (a collection or investigation that hasn't be opened or used for a period of time)
- Empty datasets (anything with 0 entities or documents)

Would be interested to know your thoughts here?

tillprochaska:
> Inactive users (anyone that hasn't logged in or used Aleph for a period of time)

Prometheus is a perfect fit for exposing time-based metrics like task queue length, number of processed tasks, average task processing time, number and response time for API requests by endpoint etc. However, there is another set of metrics we’d like to collect to answer questions like “Which datasets are accessed most?” or “How many daily/monthly active users do we have?”. These metrics cannot be implemented using just Prometheus in a straightforward way – or rather, they could be implemented using just Prometheus, but that’s a bad idea. Here’s why:

In Prometheus, you can add labels to your metrics. For example, you might have a counter metric called “user_session_created_total” that is incremented every time a user signs in. You can also specify an arbitrary number of labels, e.g. you might pass a `auth_method` label that could be `oauth` or `password`. You can then also use labels when querying your metrics, e.g. to find out how many session where created using the `password` auth method you’d do something like `user_session_created_total{method="password"}`.

It might be tempting to do the same thing with user IDs to. Just add a `user_id` label to the metric and now we’re tracking how often individual users signed in, too, right?

Unfortunately, it’s a bit more complex. Under the hood, Prometheus stores every combination of label values as a separate time series in its database. So if you have an `auth_method` label that can have two values (`password` and `oauth`), that results in two time series stored in the database. But the `user_id` label could have as many values as you have users in your Aleph instance -- which could easily get slow and expensive.

Good thing is that we actually do not need to know when and how often exactly a specific user signed in. All we want to know is how many users have been active in the past 7 or 30 days etc. The solution for this is actually quite simple:

* Add a separate `last_login_at` timestamp to the table in Postgres that stores user information.
* Whenever a user signs in, update that timestamp. (We have a rather short session length by default, so the time of last sign-in should be a good enough proxy for recent activity.)
* Expose a metric `active_users` with a `period` label (that has a limited set of labels, e.g. `7d`, `30d`, `1y`). The values for this metric can be easily calculated by running a SQL query.

I actually already prototyped this, but didn’t want to include it in the initial PR to keep it easy to review.

> Inactive datasets (a collection or investigation that hasn't be opened or used for a period of time)

This is similar to the use case above, but I think the solution is a bit more complex for two reasons:

* What qualifies as an active dataset? Does that mean someone had to open the dataset? Is it enough for at least one entity from that dataset to have been included in the results for a search? At least one xref match?

* While updating the `last_login_at` timestamp for a user once per session is not a performance bottleneck, writing to the database every time a dataset is accessed would likely be a bad idea.

So this is likely going to be slightly more complex if we actually want to implement it.

> Empty datasets (anything with 0 entities or documents)

Haven’t thought about this one in detail yet. I think we might be able to leverage the collection stats we already compute and cache (but they are sometimes incorrect or out of date, so that would be reflected in the metric).

helm/charts/aleph/templates/worker.yaml
```diff
@@ -103,3 +107,36 @@ spec:
...
+metadata:
+  name: {{ .Values.global.namingPrefix }}-worker
+  labels:
+    app: {{ .Values.global.namingPrefix }}-worker
+spec:
+  selector:
+    app: {{ .Values.global.namingPrefix }}-worker
+  ports:
+    - port: 9100
+      targetPort: 9090 # TODO: Set to 9100 once servicelayer fix is released
```
tillprochaska, line 123: Note to self: Do not overlook this!

aleph/logic/xref.py
```diff
@@ -40,6 +42,36 @@
...
+
+XREF_CANDIDATES_QUERY_DURATION = Histogram(
+    "aleph_xref_candidates_query_duration_seconds",
+    "Processing duration of the candidates query (excl. network, serialization etc.)",
+)
+
+XREF_CANDIDATES_QUERY_ROUNDTRIP_DURATION = Histogram(
+    "aleph_xref_candidates_query_roundtrip_duration_seconds",
+    "Roundtrip duration of the candidates query (incl. network, serialization etc.)",
+)
```
tillprochaska, line 74: One additional metric related to xref that would be interesting is how long it takes to process a batch of entities from a single ES scroll response. This could be used to alert us when we are getting close to the scroll timeout.

I haven’t implemented this so far because the individual scroll requests are abstracted away by the `scan` helper from the ES Python client so it would consider a non-trivial amount of work and would increase complexity.

aleph/metrics/collectors.py
```diff
@@ -0,0 +1,177 @@
...
+        query = (
+            Collection.all()
+            .with_entities(Collection.category, func.count())
+            .group_by(Collection.category)
+        )
+
+        for category, count in query:
+            gauge.add_metric([category], count)
+
+        return gauge
```
tillprochaska, line 60: This metric could be extended in the future to expose number of collections by countries.

aleph/migration.py
```diff
@@ -34,7 +34,7 @@ def destroy_db():
     while len(tables):
         for table in tables:
             try:
-                table.drop(checkfirst=True)
+                table.drop(bind=db.engine, checkfirst=True)
```
tillprochaska, line 37: This isn’t strictly related to this PR, but I think it might have been overlooked during the SQLAlchemy 2 migration and it caused migrations to fail in my development environment.

aleph/views/stream_api.py
```diff
@@ -53,4 +64,21 @@ def entities(collection_id=None):
...
+        for entity in entities:
+            count += 1
+
+            if count == 1000:
+                STREAMED_ENTITIES.inc(count)
+                count = 0
+
+            yield entity
+
+        STREAMED_ENTITIES.inc(count)
```
tillprochaska, line 82: I initially thought that increasing the counter on every iteration might be a bad idea -- that’s why I incremented it only every 1k entities. But thinking about it again, this might be a premature optimization, I should double-check that.

helm/charts/aleph/templates/exporter.yaml
```diff
@@ -0,0 +1,110 @@
...
+            - name: ALEPH_SECRET_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: aleph-secrets
+                  key: ALEPH_SECRET_KEY
+            - name: SENTRY_DSN
+              valueFrom:
+                secretKeyRef:
+                  name: aleph-secrets
+                  key: SENTRY_DSN
```
tillprochaska, line 65: It might make sense to extract these (and possibly other settings that are reused across multiple templates) into a Helm partial. Right now, we duplicate this across the templates for `api`, `ingest-file`, `worker`, and `api`.

helm/charts/aleph/templates/ingest-file.yaml
```diff
@@ -103,13 +104,35 @@ spec:
...
+  selector:
+    app: {{ .Values.global.namingPrefix }}-ingest-file
+  ports:
+    - port: 9100
+      targetPort: 9100
+      protocol: "TCP"
+      name: metrics
+{{ end }}
+---
+{{ if .Values.global.serviceMonitor.enabled }}
```
tillprochaska, line 124:
```suggestion
{{ if and .Values.exporter.enabled .Values.global.serviceMonitor.enabled }}
```

This should be consistent across all templates.

helm/charts/aleph/templates/api.yaml
```diff
@@ -129,6 +138,10 @@ spec:
           secret:
             secretName: service-account-aleph
         {{ end }}
+        {{ if .Values.global.prometheus.enabled }}
+        - name: prometheus-multiproc-dir
+          emptyDir: {}
+        {{ end }}
```
tillprochaska, line 144: One thing I’m unsure about: The multiprocess mode collects metrics of multiple processes in file-based databases in this directory. The directory should be emptied on restart, but if we collect a lot of metrics and the processes run for a long time, could we run into a situation where we fill up all available disk space?

aleph/metrics/flask.py
```diff
@@ -0,0 +1,93 @@
...
+    "http_request_duration_seconds",
+    "Duration of requests to the Aleph API in seconds",
+    ["method", "status", "api_endpoint"],
+)
+
+REQUEST = Counter(
+    "http_request_total",
+    "Total number of Aleph API requests",
+    ["method", "status", "api_endpoint", "logged_in"],
+)
```
tillprochaska, line 27: Should these be prefixed with `aleph_`, too?

gunicorn.conf.py
```diff
@@ -0,0 +1,34 @@
...
+
+_PROM_ENABLED = os.environ.get("PROMETHEUS_ENABLED", False)
+_PROM_MULTIPROC_DIR = os.environ.get("PROMETHEUS_MULTIPROC_DIR", None)
+
+wsgi_app = "aleph.wsgi:app"
+bind = "0.0.0.0:8000"
+timeout = 3600
+
+if _PROM_ENABLED:
+    # Gunicorn will bind to port 8000 (the default, publicly exposed port)
```
stchris, line 13: Thanks for documenting this in such detail!

requirements.txt
```diff
@@ -36,6 +36,7 @@ requests[security] >= 2.25.1, < 3.0.0
 tabulate==0.9.0
 zipstream-new==1.1.8
 sentry-sdk[flask]==1.31.0
+prometheus-client==0.17.1
```
stchris, line 39: The latest version on PyPI seems to be `0.19.0`. I assume this is because you started on this a while ago? Not saying we'd need to update immediately.

helm/charts/aleph/templates/exporter.yaml
stchris: Good point. I created https://github.com/alephdata/aleph/issues/3496

stchris (approved):
Looks good to me, I left some comments. Feels good to have instrumentation, thanks for your hard work, @tillprochaska !

requirements.txt
tillprochaska: That’s a good point, I’ll bump it to the latest version, saves us a Dependabot PR in the future.

helm/charts/aleph/templates/api.yaml
richardjennings-occrp: The underlying storage for emptyDir will come from the Node itself. You can add a sizeLimit if you want to prevent it  from being able to get too big. To help the Scheduler insure a Node has sufficient space, you can specify limits / requests for ephemeral-storage.

requirements.txt
tillprochaska: Sooo… I pinned the exact version in servicelayer, so of course I cannot simply upgrade it here 🙃. Even though it’s still a 0.x.x, from looking at past releases, it seems that there have been almost no breaking changes so far. So I’m inclined to relax the version constraints in servicelayer to `>= 0.19.0, < 1.0.0`.

As that requires another round of releases and bumping versions, I think I’d prefer doing that separately though, as the latest release doesn’t include any changes relevant to us…

stchris: Sorry I didn't get back to you on this. I'm fine with the proposed changes in separate PRs. 👍

aleph/metrics/flask.py
tillprochaska: Added the prefix for these metrics too as it makes it much more convenient to filter out only metrics related to Aleph.

helm/charts/aleph/templates/api.yaml
tillprochaska: So I confirmed my assumption manually, everything seems to be working as expected:

* Set a very small size limit of 50 Ki for `prometheus-multiproc-dir`.
* Ran Aleph in a local Kubernetes cluster with Prometheus enabled.
* Navigated to a few routes in the API to trigger metrics collection.
* With some delay, the container was stopped and the pod replaced by a new one. Could confirm this by looking at the k8s events:

  ```
  Warning  Evicted    5m31s  kubelet            Usage of EmptyDir volume "prometheus-multiproc-dir" exceeds the limit "50Ki".
  Normal   Killing    5m31s  kubelet            Stopping container aleph
  ```

PR #3220, opened 2023-07-18
Populate type groups in ES index

### LINKED ISSUE #2315: UI not returning date histogram when document entities have dates
The UI is sometimes not returning the date histogram, even though the Document entities have date property values set. 

https://aleph.occrp.org/datasets/5707?csfacet=dates&csfacet_interval%3Adates=year&csfacet_size%3Adates=10&csfacet_total%3Adates=true&cslimit=30&csq=#mode=search

Here is the API call:

https://aleph.occrp.org/api/2/entities?facet=dates&facet_interval:dates=year&facet_size:dates=10&facet_total:dates=true&filter:collection_id=5707&filter:schemata=Thing&highlight=true&limit=30&q=

[date_facet_response.json.zip](https://github.com/alephdata/aleph/files/8859050/date_facet_response.json.zip)

### LINKED ISSUE #2500: BUG: Languages and countries filter not showing results
**Describe the bug**
When searching an investigation I see that some files have an associated Country and Language, nevertheless when I click on the Language or Country filters in the left bar it doesn't show any results.

**To Reproduce**
1. Go to an investigation with documents that have a detected language and country
2. Click on `Countries` in the left side
3. No Countries are shown

**Expected behavior**
The detected languages should be listed in order to filter out documents.

**Additional context**
I'm running 3.12.6 and I was not able to reproduce the error in OCCRP aleph's open instance

### LINKED ISSUE #2675: BUG: Date field doesn't work for non-emails - neither filtering on the side pane nor sorting
**Describe the bug**
The date field doesn't not work for non-e-mail-types such as pdf. When searching the side pane is shown where you can filter on dates using the histogram. The time line shows only the dates for e-mail documents. When adding the dates column and clicking on it for sorting, the sorting will only work for e-mails. 

**To Reproduce**
1. Go to the starting page after logging in.
2. Click on the searchfield at the top and run an example search.
3. Add the dates column field via the configure columns button. After adding the column, click on it for sorting. You will realize it does not sort non-emails. The dates field for pdf are not filled. For excel tables they are filled but the sorting doesn't work.
4. The Dates-histogram on the left side pane shows only the dates for e-mails, as well.

**Expected behavior**
The dates column should sort all dates given in that column and not only the e-mails.

**Aleph version**
3.12.6

**Screenshots**
grafik

**Additional context**
In the elasticsearch database I can see the dates is not a properties field for non-emails. Might this be the reason?

tillprochaska:
This will likely fix a few bugs related to search facets. The common root cause for all of these is that values for property type groups such as `languages`, `countries`, `dates` etc. are not populated correctly before indexing entities.

I’m not a 100% sure this is the best solution and whether it might have unintended side effects and need to do a little more research before this can be merged.

Fixes #2675, fixes #2500, fixes #2315.

**To do:**

- [x] Check whether populating type groups has unintended side effects (besides slightly increasing index size), especially with regards to XREF.
- [x] Decide how to handle `processingDate`. Make it a hidden property and ignore hidden properties when generating group values?
- [ ] Exclude properties from type groups that will likely add too much noise (`Thing:createdAt`, `Thing:retrievedAt`, `Document:processedAt`, …)
- [ ] Double-check that including non-matchable properties in type groups doesn’t affect XREF results.
- [ ] Increase number of XREF candidates???
- [ ] Add tests

### DIFF
--- a/aleph/index/entities.py
+++ b/aleph/index/entities.py
@@ -27,6 +27,16 @@ PROXY_INCLUDES = [
     "updated_at",
 ]
 ENTITY_SOURCE = {"includes": PROXY_INCLUDES}
+FACET_PROPERTY_TYPE_GROUPS = [
+    "dates",
+    "countries",
+    "languages",
+    "emails",
+    "phones",
+    "names",
+    "addresses",
+    "mimetypes",
+]
 
 
 def _source_spec(includes, excludes):
@@ -193,6 +203,12 @@ def format_proxy(proxy, collection):
     fps = set([fingerprints.generate(name) for name in names])
     fps.update(names)
     data["fingerprints"] = [fp for fp in fps if fp is not None]
+
+    # Populate property type groups used in search facets
+    for group in FACET_PROPERTY_TYPE_GROUPS:
+        values = proxy.get_type_values(registry.groups[group])
+        if values:
+            data[group] = values
 
     # Slight hack: a magic property in followthemoney that gets taken out
     # of the properties and added straight to the index text.


tillprochaska:
Hi @pudo, may I ask you for some help in order to understand the background of a specific implementation detail in Aleph? I don’t need a direct/fast answer, so feel free to ignore this if you have too much other stuff going on atm! :)

We’ve seen issues with values from certain properties not being included in dataset statistics and search facets, for example `Document:detectedLanguages`, `Document:detectedCountries`, `authoredAt` etc.

During indexing, Aleph used to call `proxy.to_full_dict(matchable=False)` to add inverted type groups. Some time ago, you changed this ([commit](https://github.com/alephdata/aleph/commit/31fa0598635aab9b48535891250d5a86acecd69c)) so it now call `proxy.to_full_dict(matchable=True)`. I’m assuming that this was done in order to reduce the size of the formatted entity in the index, because inverting all properties creates a lot of overhead (e.g. this would duplicate the `bodyText` values which is unnecessary).

Is that correct? Or was there another reason for this?

However, the inverted type groups are also used to compute the stats displayed on dataset index pages and for the default search facets. As properties like `detectedLanguages` and `authoredAt` are not matchable, right now values from these properties are not included in the inverted type groups. And consequently, they are not considered when using search facetes on these groups or computing the dataset stats.

This PR changes how we format proxies during indexing, so inverted type groups for the types relevant for search facets and dataset stats are computed for all properties, no matter whether they are matchable or not.

I’m wondering whether this could have significant disadvantages (besides slightly increasing index size) or unintended side effects?

As the inverted type groups are also used in queries to find candidates for XREF, this might increase the number of XREF candidates that in the end do not match (because the inverted type groups would now include values from non-matchable properties). Not sure this would be a huge issue, as these false positives should be discarded when computing xref scores (which only considers matchable props).

Anything else besides that?

pudo:
Hey @tillprochaska - thanks for checking in on this. For me, the `matchable` thing is not at all about saving on index size - it's about avoiding noise in the matching and UI.

The idea was that you'd have some properties - `authoredAt` or `createdAt` are good examples - that contain potentially matchable information (dates) but where doing comparison does not make any sense (the `createdAt` of two humans will always be a property of the database, not the human). Or: the industry code of a company is an `identifier` but the question of whether two companies trade steel does very very little in terms of telling us that it's the same company. Two `Things` have the same `publisherUrl`? Unimpressive. 

I think there's probably a legitimate question about whether the `detected*` props should be made matchable, my guess is that they were left out because of concerns about noise in the facets (the most common extracted name from documents used to be the name of the app generating the documents).

So: putting them in the type groups probably isn't going to kill anyone. Not sure it'll make for a prettier app, either.

PR #3234, opened 2023-07-26
Add cleaned option

ksotik:
This option allow to disable server-side values validation for all types.
Also fix "safe" option docs and default value inverse logic.

### DIFF
--- a/aleph/views/collections_api.py
+++ b/aleph/views/collections_api.py
@@ -231,10 +231,17 @@ def bulk(collection_id):
           minimum: 1
           type: integer
       - description: >-
-          This will disable checksum security measures in order to allow bulk
-          loading of document data.
+          This will disable (if safe=False) checksum security measures in order to allow bulk
+          loading of document data. Flag is only available for admins.
         in: query
-        name: unsafe
+        name: safe
+        schema:
+          type: boolean
+      - description: >-
+          This will disable (if cleaned=True) values validation for all types of all entities/properties.
+          Flag is only available for admins.
+        in: query
+        name: cleaned
         schema:
           type: boolean
       requestBody:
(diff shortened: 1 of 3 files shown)


aleph/views/collections_api.py
```diff
@@ -258,19 +265,26 @@ def bulk(collection_id):
...
+    # This will disable (if False) checksum security measures in order to allow bulk
     # loading of document data:
     safe = get_flag("safe", default=True)
     # Flag is only available for admins:
     if not request.authz.is_admin:
         safe = True
 
+    # This will disable (if True) values validation for all types of all entities / properties
+    # (will pass cleaned=True to the model.get_proxy() in the aleph/logic/processing.py)
+    cleaned = get_flag("cleaned", default=False)
```
tillprochaska, line 277: Slightly concerned about the fact that the two flags `cleaned` and `safe` do related things but behave in opposite ways when set to `true`:

* `cleaned=true`: Assume that the data **can be trusted** and contains only valid values.
* `safe=true`: Assume that the data **can NOT be trusted**, so perform additional security measures.

Not sure though if there is a better alternative and probably not a huge problem as this is mostly relevant in edge cases. A `validate` flag is used in some other endpoint. It’s has a slightly different behavior though (`validate=true` means that Aleph will return an error instead of silently discarding invalid values).

tillprochaska: Oh, maybe the flag could be `clean` instead of `cleaned` (and default to `true`)?

* `clean=true` means that the data cannot be trusted and that the data should be cleaned from invalid values.
* `safe=true` means that the data cannot be trusted and that file checksums should be removed.

ksotik: done!

Rosencrantz:
@tillprochaska Can you take another look at this, ensure that this is now up to your standards.

tillprochaska (approved):
Really sorry this took so long, but thanks a lot for opening the PR and making these changes. I’ll also merge the respective alephclient PR.

PR #3239, opened 2023-08-01
PoC OAuth2

tillprochaska:
### Do not merge this!

This is a proof of concept to allow authentication against the Aleph API using OAuth 2. It is definitely not production-ready. I’m only using this PR to document my findings for future reference.

### Why

Aleph is used in by many different organizations and individuals and some of its users (including OCCRP) have custom requirements or processes. However, features that are only relevant to a small subset of Aleph’s users or that go beyond the scope of Aleph should most likely not be part of Aleph’s core.

Also, being able to deploy experiments for new features outside of Aleph would be very useful. This would enable faster development and iterations outside of Aleph’s regular release cycle, without being limited by technical legacy in Aleph, and with limited risk of breaking things.

Some example use cases include:

* Allowing users to load spreadsheets uploaded to or entities created in Aleph into a Datasette (Light) instance.
* Deploying an experimental, new UI to edit and view timelines.
* Creating an integration with or layer on top of Jupyter notebooks (or a similar environment) to allow users to run notebooks/scripts for common tasks self-service. (At OCCRP the Data Desk has notebooks for common tasks like exporting or post-processing data from Aleph.
* Experimenting with new search interfaces (new UIs, query builders, natural language, …)
* Developing a Wiki integration to display rich previews of linked Aleph entities.
* Developing an integration with ID to suggest relevant search results based on a ticket or to store data from a ticket in Aleph.
* Building a browser extension to extract information from articles and pages on the web as Aleph entities.

Many of the these use cases can be covered by deploying standalone apps side-by-side with Aleph. However, right now there is no authentication mechanism to let these apps securely authenticate on behalf of a user in order to access data in Aleph.

### What

This PR implements a PoC for an OAuth 2 server in Aleph, allowing OAuth clients to acquire authorization from Aleph users to access their data. Upon authorization, clients can request an access token from the OAuth server that the client can then use to authenticate with the Aleph API.

If you have ever used a [third-party GitHub App](https://github.com/marketplace?type=apps), this is exactly the same auth flow. However, one notable difference is that GitHub allows anyone to create a third-party app. For Aleph, it might make sense to limit this only to admins/operators of Aleph instance.

The PR uses [authlib](https://docs.authlib.org/en/latest/) (we already use the OAuth client from authlib for integration with OAuth identity providers like Keycloak). authlib implements most of the OAuth standards, this PR implements only application-specific logic (storing client information and tokens, setting up routes to handle request).

OAuth 2 specifies multiple different auth flows. This PR implements the [authorization code](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) flow which is quite common for this use case.

This PR also includes a command for the `aleph` CLI to create OAuth clients and a small example app. The example app implements the client auth flow manually, which is obviously a bad idea for a production use case -- I did that as a learning exercise.

### Implementation details

* authlib comes with many integrations, including integrations for Flask and SQLAlchemy. This means that creating a SA model to store client information in the database is super simple: https://github.com/alephdata/aleph/blob/feature/oauth-poc/aleph/oauth2/model.py#L20-L22

* While the SA integration also comes with model mixins for access tokens, this implementation doesn’t make use of them and instead stores them in Redis. This is what we already do for "normal" access tokens issued when authenticating in the UI and removes the need for a database roundtrip on every request. This requires some additional custom implementation to store, read, and delete tokens in Redis, something that might actually not be necessary/worth it.

* When logging in to the Aleph UI, the user’s authentication token is stored in local storage. This means that API requests always have to explicitly include the token in the request headers. However, the OAuth Authorization Code flow relies on a redirect to the `/oauth2/authorize` API endpoint (after the user has logged in). As the authentication token is stored in local storage, it isn’t included in the request to the `/oauth2/authorize` endpoint.

  To work around this, I have updated Aleph’s session handling to store the authentication token in an HTTP-only cookie (instead of local storage). This might be a good idea anyways (as it makes it impossible for JavaScript code to extract the authentication token), but it might also have unintended side effects?

### Ideas

* We use a custom auth flow for the UI which requires that the Aleph API knows about implementation details of the UI. Could we decouple this and simply implement the UI as a generic OAuth client, removing the need for custom auth flows in Aleph?

* Support public clients without a secret (currently client secrets are required) and PKCE or PKCE in addition to client secret.

* This PoC issues access tokens that always have access to everything the user has access to. Depending on use cases, using OAuth scopes to limit access might make sense, for example to limit access to specific datasets/investigations, actions (e.g. read only) or endpoints (e.g. user/role info only, no access to datasets/investigations).

* This PoC currently uses short access token lifetimes (60 mins), but doesn’t issue refresh tokens. authlib can easily be configured to generate refresh tokens, but this would require a small change to also persist the refresh tokens (potentially in a more permanent store than Redis).

* It might make sense to permanently store access tokens (even after they expire) for auditing purposes.

### Demo

I’ve pushed a very simple example application. It let’s users authenticate via Aleph and fetches some basic information about the user to display a personalized greeting. Nothing fancy. Here’s how to run it:

Make sure to run database migrations after checking out this branch, then start Aleph:

```
make upgrade
make web
```

Open the Aleph UI in your browser and sign in. (You need to be sign in, otherwise the auth flow will fail. Usually in an OAuth flow, when a user is not signed in, you’d redirect them to the sign in page of your app and then continue the flow afterwards. This would require a few small changes to the Aleph UI, but I haven’t implemented them in this PoC.)

Next, create a new OAuth client using the `aleph` CLI:

```
make shell
aleph oauth-client create --name "My Aleph App" --redirect-uri "http://localhost:1234/callback"
```

Create `oauth-example-app/.env` based on the template and add the client ID and secret from the previous step.

```
cp ./oauth-example-app/.env.template ./oauth-example-app/.env
```

Now start the example app:

```
docker compose -f docker-compose.dev.yml up oauth-example-app
```

Open http://localhost:1234 in your browser. Click on the "Login" link. This will start the OAuth flow. You will be redirected back to http://localhost:1234 on successful completion of the flow. The page should then display "Hi {name}!" where `{name}` is the user’s name requested from the Aleph API. While this is a very simple example, the client app can access other API endpoint too, for example to fetch entity data (as long as the user has access to the entities).

### DIFF
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -7,6 +7,7 @@ from aleph.core import db, cache
 from aleph.settings import SETTINGS
 from aleph.model import Collection, Role, Permission
 from aleph.model.common import make_token
+from aleph.oauth2.logic import load_token as load_oauth2_token
 
 log = logging.getLogger(__name__)
 
@@ -165,6 +166,23 @@ class Authz(object):
             is_admin=state.get("is_admin"),
             token_id=token_id,
         )
+
+    @classmethod
+    def from_oauth2_token(cls, access_token):
+        token = load_oauth2_token(access_token)
+
+        if not token or token.is_expired():
+            raise Unauthorized()
+
+        if not token.user_id:
+            raise Unauthorized()
+
+        role = Role.by_id(token.user_id)
+
+        if not role:
+            raise Unauthorized()
+
+        return cls.from_role(role)
 
     @classmethod
     def flush(cls):
--- a/aleph/core.py
+++ b/aleph/core.py
@@ -120,6 +120,11 @@ def create_app(config=None):
 
     mount_app_blueprints(app)
 
+    # TODO Fix circular dependency?
+    from aleph.oauth2.server import configure_oauth2_server
+
+    configure_oauth2_server(app)
+
     # This executes all registered init-time plugins so that other
     # applications can register their behaviour.
     for plugin in get_extensions("aleph.init"):
--- a/aleph/manage.py
+++ b/aleph/manage.py
@@ -43,6 +43,7 @@ from aleph.logic.permissions import update_permission
 from aleph.util import JSONEncoder
 from aleph.index.collections import get_collection as _get_index_collection
 from aleph.index.entities import get_entity as _get_index_entity
+from aleph.oauth2.logic import create_oauth_client
 
 log = logging.getLogger("aleph")
 
@@ -517,3 +518,21 @@ def evilshit():
     delete_index()
     destroy_db()
     upgrade()
+
+
+@cli.group(help="Manage OAuth clients")
+def oauth_client():
+    pass
+
+
+@oauth_client.command("create")
+@click.option("-n", "--name", help="Name of the OAuth client", required=True)
+@click.option(
+    "-r", "--redirect-uri", help="OAuth redirect URI", required=True, multiple=True
+)
+def create_app(name, redirect_uri):
+    client = create_oauth_client(name, redirect_uri)
+
+    print(f"OAuth 2 client created.")
+    print(f"Client ID: {client.client_id}")
+    print(f"Client secret: {client.client_secret}")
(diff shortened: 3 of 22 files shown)


tillprochaska:
Closing this PR as it is a proof of concept that was never intended to be merged (at least not in its current state).

PR #3276, opened 2023-08-18
Add "Add to..." button to entity preview

adryd325:
Perhaps the "Add to..." button should be seperated into its own component


tillprochaska:
Hi @adryd325, thanks a lot for opening your first PR in this repository. I think adding this button to the entity preview is a good idea, but I will need to take a closer look which will unfortunately take some time. Sorry about that! :)

Rosencrantz:
@adryd325 Just to let you know we've not forgotten about this issue.

Rosencrantz:
@adryd325 Just something we noticed. You've branched from our main branch. Can you update this work to come from the develop branch instead? If you check our dev guidelines you'll see that all new work needs to be branched from the develop branch: https://github.com/alephdata/aleph/blob/develop/CONTRIBUTING.md#branches

tillprochaska (commented):
Hi! Sorry it took so long to review this. I like the idea and implementation, but I agree with you that it might be a good idea to extract this into a separate collection.

In addition to `EntityToolbar` we already use a button like this in two other places – and the implementations have already slightly diverged (my fault!). 

https://github.com/alephdata/aleph/blob/develop/ui/src/components/EntityTable/EntityTable.jsx#L263
https://github.com/alephdata/aleph/blob/develop/ui/src/screens/EntityScreen/EntityScreen.jsx#L113

Let me know if you’d like to extract the button into a separate component yourself. Otherwise, I can also take care of that before merging :)

adryd325:
> Let me know if you’d like to extract the button into a separate component yourself. Otherwise, I can also take care of that before merging :)

I'll take a look at that now

> Can you update this work to come from the develop branch instead?

I've changed the base to develop and caused a mess, i'll try fix that up or re-create the PR against develop

adryd325:
I've managed to move the branch over to develop, and I've moved the Add to button to its own element, I've not had the chance to test yet.

tillprochaska:
@adryd325 Cool, thanks a lot. Just ping me when you’ve had a chance to test it and I’ll take another look!

Issue #3275, opened 2023-08-17
BUG: Highlights not working for more than 1 token but I am completly lost here


simonwoerpel:
**Describe the bug**
When searching for "lorem", in the highlight context preview the token "lorem" is highlighted. When searching for "lorem ipsum", nothing is displayed.

**To Reproduce**
1. Check out the current `main` or `develop` branch (both affected) on a local machine
2. Expose port `9200` from elasticserch to localhost for debugging
3. Don't change anything else, just start a fresh aleph with `docker compose up`
4. Create a new investigation
5. Upload a text document `highlight.txt` with this content:
`Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod
tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At
vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren,
no sea takimata sanctus est Lorem ipsum dolor sit amet.`
6. In the Aleph UI, search for "lorem ipsum" vs. "lorem" (refresh the browser between searches as highlights might be cached in the UI)
7. check the highlight query aleph is producing against elastic:

```bash
curl -X GET "localhost:9200/aleph-entity-plaintext-v1/_search?pretty" -H 'Content-Type: application/json' -d'{"query":{"bool":{"should":[{"query_string":{"query":"lorem ipsum","lenient":true,"fields":["fingerprints.text^3","text"],"default_operator":"AND","minimum_should_match":"66%"}}],"must":[],"must_not":[],"filter":[],"minimum_should_match":1}},"post_filter":{"bool":{"filter":[]}},"from":0,"size":30,"aggregations":{},"sort":[],"highlight":{"encoder":"html","fields":{"properties.*":{"highlight_query":{"query_string":{"query":"lorem ipsum","lenient":true,"fields":["properties.*"],"default_operator":"AND","minimum_should_match":"66%"}},"require_field_match":false,"number_of_fragments":3,"fragment_size":120,"max_analyzed_offset":999999}}},"_source":{"includes":["schema","properties","collection_id","profile_id","role_id","mutable","created_at","updated_at"]}}'
 ```

-> no highlights for "lorem ipsum", but for "lorem".

**Expected behavior**
When searching for "lorem ipsum", the two tokens "lorem ipsum" should be highlighted.

**Aleph version**
[...]

PR #3278, opened 2023-08-29
Hotfix: multi token highlights #3275

simonwoerpel:
This fixes the issue discussed here: https://github.com/alephdata/aleph/issues/3275

As discussed, the origin of this bug and why this one-liner fixes it are not fully investigated, nor are some side-effects.

First experiences in a running production environment are positive.

### DIFF
--- a/aleph/search/query.py
+++ b/aleph/search/query.py
@@ -220,6 +220,7 @@ class Query(object):
         if not self.parser.highlight:
             return {}
         query = query_string_query(self.HIGHLIGHT_FIELD, self.parser.highlight_text)
+        query["query_string"].pop("fields")
         return {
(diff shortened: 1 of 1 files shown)


monneyboi:
I propose removing the `query_string_query`, see https://github.com/alephdata/aleph/pull/3280

Seems more clean to me compared to first creating the `fields` array and then removing it again.

stchris:
I like the suggestion from #3280 a little bit better as well, as it makes it a bit more obvious what is going on. My suggestion is to merge that and see how it works in our staging environment.

monneyboi:
Fields array has been removed from the highlight queries now as per #3280

PR #3279, opened 2023-08-29
Flask defaults, some reduced coupling & cleaning up

monneyboi:
This PR refactors some of the view logic to more closely follow the flask documentation / defaults.

It looks like a lot, but not much is going on:

- [`require`](https://github.com/alephdata/aleph/blob/eac98ec386944a82183ed5fd3816958e4ed5da17/aleph/views/util.py#L25) was dropped in favor of [`abort`](https://flask.palletsprojects.com/en/2.3.x/api/#flask.abort)
- Some HTTP 403 codes were changed to HTTP 403 where that made sense.
- [`jsonify`](https://github.com/alephdata/aleph/blob/eac98ec386944a82183ed5fd3816958e4ed5da17/aleph/views/util.py#L144) and related calls were dropped in favor of a [`JSONProvider`](https://flask.palletsprojects.com/en/2.2.x/api/#flask.json.provider.JSONProvider)
- A test was added for JSONP functionality
- Code was localized from general utils to local files where possible
- Unused code was removed

This was done with the intention of reducing coupling, side-effects, and making the code more understandable for newcomers.

### DIFF
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -77,6 +77,7 @@ class Authz(object):
             return False
         return collection in self.collections(action)
 
+    @property
     def can_bulk_import(self):
         if not self.can_browse_anonymous or not self.session_write:
             return False
@@ -97,6 +98,7 @@ class Authz(object):
             return True
         return int(role_id) in self.roles
 
+    @property
     def can_register(self):
         if self.logged_in or SETTINGS.MAINTENANCE or not SETTINGS.PASSWORD_LOGIN:
             return False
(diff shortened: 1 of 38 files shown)


stchris:
I started reviewing this but I find it really hard with all the things going on. I will need some more time. Perhaps we can focus on potentially just doing one thing per PR in the future, @monneyboi ?

Rosencrantz:
@monneyboi If you'd be willing to slit this up it'll making merging and releasing it a lot faster/easier

Rosencrantz:
ping @monneyboi

PR #3280, opened 2023-08-29
Remove query_string_query function, remove fields from highlight query

monneyboi:
As passing the `fields` parameter to the ES highlight query breaks highlights https://github.com/alephdata/aleph/issues/3275, i propose removing this `query_string_query` function completely, as these dictionaries differ a bit, only occur in 3 locations, and just creating the query_string dictionaries on the spot is more understandable.

### DIFF
--- a/aleph/index/util.py
+++ b/aleph/index/util.py
@@ -116,19 +116,6 @@ def none_query(query=None):
     return query
 
 
-def query_string_query(field, query):
-    """Default config for querying the entity text."""
-    return {
-        "query_string": {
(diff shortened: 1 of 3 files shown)


monneyboi:
I propose removing the `query_string_query`, see https://github.com/alephdata/aleph/pull/3280

Seems more clean to me compared to first creating the `fields` array and then removing it again.

stchris (approved):
Happy to give this a go, as per the discussion in #3278

PR #3281, opened 2023-08-30
discover `make format`, run that

monneyboi:
Saw that the build failed for my PR. After looking into why, i discovered `make
format`.

I'll add the `black` formatter as a plugin to my editor so we don't have to go
through this again.

### DIFF
--- a/aleph/logic/alerts.py
+++ b/aleph/logic/alerts.py
@@ -74,15 +74,17 @@ def alert_query(alert, authz):
         "_source": {"includes": ["collection_id"]},
         "query": {
             "bool": {
-                "should": [{
-                    "query_string": {
-                        "query": alert.query,
-                        "lenient": True,
-                        "fields": ["text"],
-                        "default_operator": "AND",
-                        "minimum_should_match": "66%",
+                "should": [
+                    {
+                        "query_string": {
+                            "query": alert.query,
+                            "lenient": True,
+                            "fields": ["text"],
+                            "default_operator": "AND",
+                            "minimum_should_match": "66%",
+                        }
                     }
-                }],
+                ],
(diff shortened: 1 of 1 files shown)


monneyboi:
Maybe the `make format-check`  could be added as a PR check instead of being part of the build process?

monneyboi:
I mean the [status checks](https://github.com/alephdata/aleph/pull/3281/checks) for pull requests on github. I think you could add a workflow that simply runs `make format-check` on `pull_request` events. See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#checks

This would prevent anyone from merging unformatted code.

stchris:
I'm confused. [We do this](https://github.com/alephdata/aleph/blob/27e7c2cd6c72d77ce93de1b18de6dfff899ea365/.github/workflows/build.yml#L39). Where did this (not) fail exactly?

stchris (approved):

stchris:
The build actually failed as expected here: https://github.com/alephdata/aleph/actions/runs/6023919524/job/16341591649

monneyboi:
>The build actually failed as expected

Yes, that is why i created this pull request. To fix the formatting and have it pass.

> I'm confused. [We do this](https://github.com/alephdata/aleph/blob/27e7c2cd6c72d77ce93de1b18de6dfff899ea365/.github/workflows/build.yml#L39). Where did this (not) fail exactly?

You trigger a build on `push` events. I mean the [status checks](https://github.com/alephdata/aleph/pull/3281/checks) for pull requests on github. Those are something else:
image

A status check that checks for formatting could have prevented me from merging code that would produce a failing build.

PR #3282, opened 2023-08-31
Add `make format-check` as pull request "check"

monneyboi:
This adds a pull-request "check" that will run `make format-check` to check if code from the PR is formatted correctly.

See https://github.com/alephdata/aleph/pull/3281

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,20 @@
+name: Check formatting
+
+on:
+  pull_request:
+    paths-ignore:
+      - "ui/**"
+      - "docs/**"
+
+jobs:
+  check-formatting:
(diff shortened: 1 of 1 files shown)


monneyboi:
I've not removed the `make format-check` from the [build workflow](https://github.com/alephdata/aleph/blob/5c9dabfbee8db17b79a10935898247378cfa0e5d/.github/workflows/build.yml#L39). It could be argued that is has no place there.

stchris:
> Though it could be argued that is has no place there.

I'm not sure I follow, why is the new check better? In my mind checking the formatting should be part of every commit and it should stop building if it encounters errors. Would you agree?

stchris (approved):

Issue #3334, opened 2023-09-26
BUG: Som web pages don't render


monneyboi:
Some HTML documents uploaded to Aleph don't render in the GUI. Not all sites seem to be affected by this. A couple examples:

- www.politico.eu
- www.euractiv.com

**To Reproduce**
1. Safe any article from one of the examples 
3. Upload the HTML file in Aleph
4. Open the HTML file in Aleph
5. See nothing

**Expected behavior**
I expect to see a safe render of the HTML document.

**Aleph version**
This happens in the OCCRP Aleph instance as well in ours. Aleph on our end is `3.15.1`

**Screenshots**
![image](https://github.com/alephdata/aleph/assets/86304937/26cbe1e3-8c17-4d47-aaef-d6b0772eaced)

**Additional context**
The problem seems to be caused by the `safeHtml` property not containing all html content. The `bodyHtml` property does contain everything.

monneyboi:
The PR for this issue has been waiting to be accepted for quite some time now, if that gets merged this bug can be closed.

PR #3335, opened 2023-09-26
Move safeHtml to React component.

monneyboi:
As a fix for #3334, i propose to move the safe html rendering to React for a couple of reasons:

- `bodyHtml` is being sent over the wire anyway.
- It's ultimately presentation logic

I would love to have used the [HTML sanitizer API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API) for this, but that's not yet supported in all major browsers. So i used [DOMPurify](https://github.com/cure53/DOMPurify) instead, as proposed by Mozilla in their [documentation on safely inserting HTML content in browser extensions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Safely_inserting_external_content_into_a_page).

This move also allowed me to drop the LXML python requirement.


ui/src/viewers/HtmlViewer.jsx
```diff
@@ -1,27 +1,82 @@
 import React, { Component } from 'react';
+import DOMPurify from 'dompurify'
 
 import { Skeleton } from 'components/common';
 
 import './HtmlViewer.scss';
 
-class HtmlViewer extends Component {
+const FORBID_TAGS = [
+  "meta",
```
tillprochaska, line 9: `meta`, `base`, `bgsound`, `embed`, `frame`, `frameset`, `iframe`, `link`, `meta`, `object`, `plaintext`, `svg` are not in DOMPurify’s default allow list, so it’s probably redundant to specify them here, right?

https://github.com/cure53/DOMPurify/blob/main/src/tags.js

tillprochaska (changes requested):
Thanks! A few minor notes:

1) `components/common` is the location where we usually keep components like `SafeHtmlDocument` viewer that are included in multiple other components. Could you extract the component into its own file in that directory?

2) Could you add tests (that cover at least the existing Python test cases)? There are few tests at the moment for the frontend codebase, but we try to add them for new stuff and it would be valuable especially for critical stuff like HTML sanitization imo. You can take a look at this test suite for the testing boilerplate: https://github.com/alephdata/aleph/blob/develop/ui/src/components/Timeline/util.test.ts

3) There’s a reference to `safeHtml` in `ui/src/selectors.js` that can be removed, too.

monneyboi:
I resolved the proposed changes.

> components/common is the location where we usually keep components like SafeHtmlDocument viewer that are included in multiple other components. Could you extract the component into its own file in that directory?

I would argue that this `SafeHtmlDocument` is not a common component. Common components in my mind would be things that get used throughout the codebase. 

After looking into the different "viewers", i would instead argue that this `SafeHtmlDocument` should actually be the `HtmlViewer` component, and that the `Skeleton` logic that's found in most "viewers" should ideally be moved higher up the tree to the `DocumentViewMode` component instead of being implemented in every viewer itself.

I was looking into refactoring that right away, but i couldn't quickly make sense of where the `isPending`  stuff came from, as it is found all throughout the source.

Therefore i just moved `SafeHtmlComponent` into it's own `common` file.

monneyboi:
The merge conflict seems to be caused by the `package-lock.json` file not having been updated on the `develop` branch.

monneyboi:
Merge conflict has been fixed, can this get merged?

tillprochaska:
Hi @monneyboi, sorry I haven’t replied earlier.

I took another look at this PR and noticed that it doesn’t sanitize some HTML markup properly that can trigger HTTP requests to third parties, e.g. `<div background="https://example.org"></div>` would trigger an HTTP request to `Https://example.org` when the preview is displayed.

Also, HTML documents that contain forms seem to be sanitized differently, although I don’t think this would be a security issue.

Although I’m not against moving the HTML sanitization to the client-side per se, unfortunately I currently have limited bandwith to consider all the possible implications that this change might have.

I have however opened a small PR that fixes a bug causing incorrect previews for some web pages. This should fix preview e.g. for politico.eu: #3575

PR #3379, opened 2023-10-03
added ignored settings test

brassy-endomorph:
This test was `.gitignore`'d on PR #2538. This fixes that.


catileptic:
Thanks for this PR! We try to stick to the rule of having [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits). Would you be able to sign your commit before we merge it?

PR #3380, opened 2023-10-03
Add note to signup guide

tillprochaska:
I have added notes at the top of the following two pages as they contain information specific to the Aleph instances operated by OCCRP. All other documentation pages should be generic.

Screen Shot 2023-10-03 at 11 02 24

Screen Shot 2023-10-04 at 10 36 38


Rosencrantz (approved):
Nice addition. Hopefully it will help prevent confusion.

PR #3395, opened 2023-10-08
Limit concurrency of the deploy job only

tillprochaska:
We actually want individual docs PRs to trigger the build job. And we don’t care if there are multiple build jobs running at the same time. But we want there to be only one deploy job at the same time (in addition to the deploy job only being triggered manually or by pushes to the main branch).


stchris (approved):
Agreed. Thanks for the change!

PR #3400, opened 2023-10-08
Remove ignored versions from Dependabot config

tillprochaska:
Most of the ignored items are either not relevant anymore (because that dependency is already on a newer version) or it might make sense to reevaluate whether we should defer upgrading permanently.


stchris (approved):
LGTM

Issue #2809, opened 2023-01-13
Clean up dependabot ignores


tillprochaska:
`dependabot.yml` lists quite a few package versions (pip and npm) to be ignored. I think that was necessary when migrating from the legacy version of Dependabot after GitHub integrated it.

Most of the ignored versions listed are now irrelevant, because we’re already on newer versions, so they can be removed. For the remaining ignores, we may also want to check if they are still necessary.

tillprochaska:
Closed in #3400

PR #3401, opened 2023-10-08
Set up grouped Dependabot updates

tillprochaska:
Merge #2809 first!

***

This PR adds grouped Dependabot updates for npm dependencies. That means that Dependabot will try to update certain dependencies (e.g. all Blueprint dependencies) in a single PR rather than opening separate PRs for every dependency.

This will hopefully make it easier to manage dependency updates and reduce the noise created by Dependabot updates.

I haven’t set up groups for Python dependencies so far, because noise from Dependabot PRs seems to be a much smaller problem for Python dependencies.


stchris (approved):
Looking forward to seeing this in action! Thanks!

PR #3402, opened 2023-10-08
Remove urllib3 dependency

tillprochaska:
I’m not really sure if we still need this as a direct dependency. Looking at the git history, it might have been added (years ago) in order to fix an incorrect dependency constraint in a dependency (probably the Elasticsearch Python client).

Removing it doesn’t seem to cause any problems and the latest version of `urllib3<2` gets installed as a transitive dependency (because it is required by `elasticsearch`.


tillprochaska:
We're not actually loading urllib though, just setting the log level, right?

But let me check if we do need to set the log level explicitly or if we can set it for the ES client directly in some way.

stchris (approved):

Issue #3403, opened 2023-10-10
BUG: unable to upload a large amount of files through the UI


stchris:
**Describe the bug**
Uploading a large amount of files through the UI can make an Aleph installation freeze and stop processing. 

**To Reproduce**
1. Go to an investigation
2. Click on 'Upload documents'
3. Select a large amount of files (thousands)
4. Aleph will choke and stop processing them.

**Expected behavior**
Aleph is either able to show and enforce a limit to the number of files being uploaded or it is able to queue the files so that they are processed correctly. 

**Aleph version**
3.15.1

**Screenshots**
-

**Additional context**
The current workaround is to use [alephclient](https://docs.aleph.occrp.org/developers/alephclient/#importing-all-files-from-a-directory)

andresmrm:
Thanks for helping us with this issue, Chris!
Extra context can be found in https://github.com/alephdata/aleph/issues/3383 .
"Uploading a folder with 8K files worked, but not another with 300K files"

Rosencrantz:
Options here:

1. Document a limit (1,000 - 5,000?) files max, possibly size could have an impact.
2. Try to fix the UI to handle larger file sets (harder, more time consuming).
3. Limit uploads to single files at a time

andresmrm:
If 2 is too complex for now, I would prefer 1. Including a warning in
the upload dialog about this limitation and a link to crawldir doc or
tutorial as an alternative.

PR #3483, opened 2023-11-07
Suggest alephclient for uploading lots of files instead of UI

stchris:
Relates to #3403 (first step measure)

### DIFF
--- a/ui/src/dialogs/DocumentUploadDialog/DocumentUploadForm.jsx
+++ b/ui/src/dialogs/DocumentUploadDialog/DocumentUploadForm.jsx
@@ -86,7 +86,18 @@ export class DocumentUploadForm extends PureComponent {
             }}
           />
         </p>
-      </div>
+        <p className="DocumentUploadForm__info">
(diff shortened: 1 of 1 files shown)


ui/src/dialogs/DocumentUploadDialog/DocumentUploadForm.jsx
```diff
@@ -86,6 +86,19 @@ export class DocumentUploadForm extends PureComponent {
             }}
           />
         </p>
+        <p className="DocumentUploadForm__info">
```
tillprochaska, line 89: Is the class name actually used anywhere? Think it might be leftover from when you had custom styles applied to this element?

stchris: I wasn't using it, I thought this was the way to do it, perhaps for accesibility, but I see how that doesn't make any sense now. Removing it :)

tillprochaska (approved):

Issue #3384, opened 2023-10-05
BUG: Documentation for configuring Aleph workers is lacking


stchris:
**Describe the bug**
It is not immediately obvious that aleph and ingest-file workers are using "number of cpus" threads, which can cause scaling issues as seen in numerous reports, the last of which is #3383 .

**To Reproduce**
1. Follow the installation / deployment docs
2. Massively scale up ingest-file, workers etc.
3. Run out of postgres connections

**Expected behavior**
Have better control over the number of workers running. Make it visible that we are using threads by default. Explain what the [servicelayer workers default to](https://github.com/alephdata/servicelayer/blob/b657e3b4039e13158d80d653bdeb394a47be8ae1/servicelayer/settings.py#L17) and how to disable that.

**Aleph version**
3.15.1

**Screenshots**
-

**Additional context**
-

tillprochaska:
Related #3064

Rosencrantz:
Would be good to set the limit to 0, and disable threads entirely. Also update the documentation to talk more about how we (don't) use this feature and what we recommend other people do.

PR #3491, opened 2023-11-08
Explicitly disable servicelayer worker threads in Helm chart & env

stchris:
This is part of #3384

### DIFF
--- a/aleph.env.tmpl
+++ b/aleph.env.tmpl
@@ -125,3 +125,6 @@ SENTRY_DSN=
 
 # Sentry environment description, see https://docs.sentry.io/platforms/python/configuration/environments/
 SENTRY_ENVIRONMENT=
+
+# Number of threads to use in servicelayer workers, 0 = disabled.
+WORKER_THREADS=0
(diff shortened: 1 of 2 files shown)


aleph.env.tmpl
```diff
@@ -125,3 +125,6 @@ SENTRY_DSN=
 
 # Sentry environment description, see https://docs.sentry.io/platforms/python/configuration/environments/
 SENTRY_ENVIRONMENT=
+
+# Number of threads to use in servicelayer workers, 0 = disabled.
```
Rosencrantz, line 129: Maybe provide a bit more context here. I'd like to see some information on why we have this set to zero, or a link to docs that explain how to determine the number of servicelayer workers?

stchris: Agreed, will improve this.

tillprochaska (approved):
Looks good to me, thanks! Might be worth noting the new default in the release notes for the Helm chart, although in practice I doubt it would have significant impact for anyone.

PR #3493, opened 2023-11-09
Update K8s example

tillprochaska:
The previous example for running Aleph in a local Kubernetes cluster was based on kind. This simplifies the example and default configuration, making it easier to run a minimal Aleph instance in a local Kubernetes cluster provided by Docker Desktop.

Additionally, this PR includes two changes to the Aleph Helm chart:

* Sets the default `ingest-file` version to `3.19.3`, the latest release (the default version in the chart was out of sync with the latest release).

* Marks the `SENTRY_DSN`key ref as optional. This way, starting containers will succeed, even if the `SENTRY_DSN` key does not exist in the `aleph-secrets` secret. Previously, starting containers that reference the `SENTRY_DSN` key would fail (requiring users to manually create an empty `SENTRY_DSN` key in the `aleph-secrets` secret).


helm/examples/dev/templates/ingress.yaml
```diff
@@ -1,11 +1,13 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: aleph-ingress-dev
-  namespace: dev
+  name: aleph-ingress
+  namespace: default
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
```
stchris, line 7: Perhaps we can fix this warning:

```
annotation "kubernetes.io/ingress.class" is deprecated, please use 'spec.ingressClassName' instead
```

helm/examples/dev/secrets/minio/secretKey
stchris, line 1: my minio pod fails to start with 

```
Error: couldn't find key secretkey in Secret default/minio-secrets
```

Perhaps the filename needs to be `secretkey`?

helm/examples/dev/Chart.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v2
+type: application
+name: aleph-dev
+version: 0.1.0
+dependencies: 
+  - name: aleph
+    repository: file://../../charts/aleph
+    version: 3.15.4
```
stchris, line 8: I'm not sure if I did something wrong, but I get aleph `3.15.1` :/

helm/examples/dev/secrets/minio/secretKey
tillprochaska: Yep, you’re 100% right. I made a few changes to the structure of secrets etc. but didn’t bother to fully test them afterwards. Sorry! I’ll push a fix…

stchris: Wow ...

helm/examples/dev/Chart.yaml
tillprochaska: Hm, that’s weird… I just reset my cluster, removed local images, and set up everything again from scratch. But looking at the about page and at the version specified inside of the container I see `3.15.4`. We could look into it together tomorrow if you’d like :)

helm/examples/dev/templates/ingress.yaml
tillprochaska: That happens when you blindly follow StackOverflow answers. Thanks for catching it, this is now fixed!

helm/examples/dev/Chart.yaml
tillprochaska: Only other reason for this I could imagine is that this branch used to be based off of 3.15.1, and I later rebased it and force pushed to be up to date with develop. But that was a few days before you looked at it, I think…

PR #3564, opened 2024-01-11
Bump FTM version and ingest-file img version

catileptic:


docker-compose.dev.yml
```diff
@@ -30,12 +30,12 @@ services:
 
   redis:
     image: redis:alpine
-    command: [ "redis-server", "--save", "3600", "10" ]
+    command: ["redis-server", "--save", "3600", "10"]
     volumes:
       - redis-data:/data
 
   ingest-file:
-    image: ghcr.io/alephdata/ingest-file:3.19.3
```
stchris, line 38: I wonder if we should have a final ingest-file release before we bump this in Aleph?

stchris (approved):
LGTM

PR #3573, opened 2024-01-23
Don't rename existing collections when running crawldir

adryd325:
Prevents existing collections from being renamed to the folder name when running crawldir. I haven't had the chance to test this but I've read over it and I don't think it will cause issues if merged


stchris:
Thanks for this, @adryd325 and congratulations on your first contribution to Aleph! 👏

PR #3575, opened 2024-01-25
Stringify sanitized HTML as HTML (not XML)

tillprochaska:
This fixes a bug where Aleph doesn’t correctly preview HTML pages (see #3334).

The `etree.ElementTree.tostring` method stringifies as valid XML by default. While XML allows self-closing tags by default, only some tags can be self-closing in HTML.

This causes problems when rendering the sanitized HTML. Consider this example:

```html
<noscript><script src="..."></script></noscript>
<h1>Page title</h1>
<p>Some text</p>
```

When converting the sanitized element tree to XML, this would result in something like this:

```html
<noscript />
<h1>Page title</h1>
<p>Some text</p>
```

`noscript` is not a valid void element which results in browsers treating all following elements as children of the `noscript` tag, parsing the HTML into an effective DOM structure like this:

```html
<noscript>
  <h1>Page title</h1>
  <p>Some text</p>
</noscript>
```

As a result the actual page contents are hidden when rendered in a browser (if JavaScript is enabled).

### DIFF
--- a/aleph/logic/html.py
+++ b/aleph/logic/html.py
@@ -57,7 +57,7 @@ def sanitize_html(html_text, base_url, encoding=None):
             if el.tag == "a":
                 el.set("target", "_blank")
                 el.set("rel", "nofollow noreferrer external noopener")
-        return tostring(doc)
+        return tostring(doc, method="html", encoding="unicode")
     except Exception as exc:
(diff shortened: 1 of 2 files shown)


stchris (approved):
Brilliant investigation, thanks for documenting this in detail! And for fixing it, of course!

PR #3580, opened 2024-01-30
Fix instructions on how to enable Prometheus in local Kubernetes cluster

tillprochaska:
The previous instructions simply don’t make sense 🙃


stchris (approved):
Works for me!

PR #3607, opened 2024-02-04
Fix link target in user guide

tillprochaska:
Fixes a broken link in the user guide on this page: https://docs.aleph.occrp.org/users/faq/account/. We do automatically check for broken internal links in CI. While this link was supposed to be targeting another documentation page, it was an absolute URL and thus not detected by the broken link checker.


stchris (approved):
👍

PR #3608, opened 2024-02-04
Fix default Dockerfile command

### LINKED ISSUE #3606: BUG:  'aleph.manage' has no attribute 'app'
Just downloaded Aleph, and changed the configuration file according to the installation guide 
And start the container with docker-compose up -d

The problem is with the API container where I get the error of  'aleph.manage' has no attribute 'app' and here after it shutsdown.

**Aleph version**
3.15.5

**Additional context**
Traceback (most recent call last):
File "/usr/local/lib/python3.8/dist-packages/gunicorn/util.py", line 414, in import_app
app = getattr(mod, name)
AttributeError: module 'aleph.manage' has no attribute 'app'
Failed to find attribute 'app' in 'aleph.manage'.
[2024-02-03 08:09:01 +0000] [9] [INFO] Worker exiting (pid: 9)

tillprochaska:
I broke the default Dockerfile command in `7cfcb5`. While the Dockerfile did always specify the WSGI app `aleph.manage:app`, this has been incorrect for some time and it should have always been `aleph.wsgi:app`. Previously we never actually ran a container with the default command though and instead specified the command in the `docker-compose.yml` file. I removed the "duplicate" command in the `docker-compose.yml` in the referenced commit and thus the default command in the Dockerfile is now used.

I’ve now changed the default command specified in `Dockerfile` to be in line with the command previously set in `docker-compose.yml` (https://github.com/alephdata/aleph/blob/3.15.5/docker-compose.yml#L76).

* The WSGI app is now `aleph.wsgi:app` (this is specified in `gunicorn.py` not via a CLI option).
* Uses 6 workers by default (instead of 5).
* Log level set to `debug` (instead of `info`).

Additionally, the installation guide in the docs referenced the configuration templates on the `develop` branch instead of `main`. We recently renamed `main` to `master` and changed the default branch to `develop`. GitHub now redirects any links that still use the old `master` branch to the default branch. I’ve fixed this too.

Closes #3606

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,4 +49,4 @@ ENV ALEPH_ELASTICSEARCH_URI=http://elasticsearch:9200/ \
 RUN mkdir /run/prometheus
 
 # Run the green unicorn
-CMD gunicorn --config /aleph/gunicorn.conf.py --workers 5 --log-level info --log-file - aleph.manage:app
+CMD gunicorn --config /aleph/gunicorn.conf.py --workers 6 --log-level debug --log-file -
--- a/docs/src/pages/developers/alephclient.mdx
+++ b/docs/src/pages/developers/alephclient.mdx
@@ -54,7 +54,7 @@ Using the `write-entities` command, users can load JSON-formatted entities forma
 A typical use might look this:
 
 ```bash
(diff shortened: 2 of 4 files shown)


stchris (approved):
Thanks for the fix(es)!

PR #3615, opened 2024-02-28
Add mimetype inclusion/exclusion flags to reingest command

stchris:
# Problem statement

Reingesting collections is sometimes necessary, either for data consistency reasons or for the sake of using newer versions of the processing agent (typically `ingest-file`). This operation can take a considerable amount of time and it can also do a lot of unnecessary work, because only some of the files should be reingested.

# Proposal

The `aleph reingest` command gets these two new flags:

```
  --include-mimetype TEXT  Only include documents with these mimetypes
                           (multiple mentions possible)
  --exclude-mimetype TEXT  Exclude documents with these mimetypes (multiple
                           mentions possible)
```

which can be mixed and reused to narrow down documents to be reingest by their mimetype.

As an example one would reingest only images with:

```
aleph reingest foreign_id --include-mimetype "image/*"
```

or exclude pdf files with:

```
aleph reingest foreign_id --exclude-mimetype "application/pdf"
```

# Known limitations

* one can shoot themselves in their limbs by not being careful mixing these. For instance 

```
aleph reingest foreign_id --include-mimetype "image/jpeg" --exclude-mimetype "image/*"
```

does **not** reingest only JPEG images, but rather excludes all images :/

### DIFF
--- a/aleph/manage.py
+++ b/aleph/manage.py
@@ -205,11 +205,37 @@ def reindex_casefiles(flush=False):
 @click.option("--index", is_flag=True, default=False)
 @click.option("--include_ingest", is_flag=True, default=False)
 @click.option("--flush/--no-flush", default=True)
-def reingest(foreign_id, index=False, flush=True, include_ingest=False):
+@click.option(
+    "--include-mimetype",
+    "include_mimetypes",
+    multiple=True,
+    default=[],
+    help="Only include documents with these mimetypes (multiple mentions possible)",
+)
+@click.option(
+    "--exclude-mimetype",
+    "exclude_mimetypes",
+    multiple=True,
+    default=[],
+    help="Exclude documents with these mimetypes (multiple mentions possible)",
+)
+def reingest(
+    foreign_id,
+    index=False,
+    flush=True,
+    include_ingest=False,
+    include_mimetypes=[],
+    exclude_mimetypes=[],
+):
     """Process documents and database entities and index them."""
     collection = get_collection(foreign_id)
     reingest_collection(
-        collection, index=index, flush=flush, include_ingest=include_ingest
+        collection,
+        index=index,
+        flush=flush,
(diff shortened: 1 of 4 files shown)


aleph/manage.py
```diff
@@ -205,11 +205,37 @@ def reindex_casefiles(flush=False):
...
+):
     """Process documents and database entities and index them."""
     collection = get_collection(foreign_id)
     reingest_collection(
-        collection, index=index, flush=flush, include_ingest=include_ingest
+        collection,
+        index=index,
+        flush=flush,
+        include_ingest=include_ingest,
+        include_mimetypes=include_mimetypes,
```
catileptic, line 237: should we import `from pantomime import normalize_mimetype` and call `normalize_mimetype`before passing these user-provided arguments on?

stchris: Great suggestion, I'll add that!

catileptic: it's what we do in [ingest-fle](https://github.com/alephdata/ingest-file/blob/main/ingestors/ingestor.py#L2)

tillprochaska (changes requested):
I added a few comments, and you do not need to necessarily address all of them. However, it would be good if you could check if this will reliably work even for files that have been ingested with alephclient or as part of another file (e.g. zip archives, email attachments, …)

aleph/model/document.py
```diff
@@ -178,6 +179,21 @@ def by_collection(cls, collection_id=None):
         q = q.yield_per(5000)
         return q
 
+    @classmethod
+    def by_collection_and_mimetype(
+        cls, collection_id=None, include_mimetypes=[], exclude_mimetypes=[]
+    ):
```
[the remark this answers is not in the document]
stchris: Thanks for this. Sounds like my approach doesn't work as expected, then. I will go back to the drawing board with this.

PR #3670, opened 2024-04-02
Update workflows to use `docker compose` instead of `docker-compose`

tillprochaska:
It seems like GitHub Actions runners do not ship docker-compose v1 anymore, so builds started failing. Fixing this is as easy as replacing `docker-compose` with `docker compose` (and a good idea anyway).


tillprochaska:
The change in the GHA runner images was reverted and this PR is missing a few files, most importantly probably the `Makefile`. I’m going to close this for now and until it becomes relevant again.

PR #3673, opened 2024-04-04
Use pytest instead of 👃

stchris:


tillprochaska (approved):
Ah, didn’t know it’s as easy as replacing the command! Thanks, this is very helpful!

Issue #3680, opened 2024-04-15
FEATURE: Remove references to convert-document from Technical FAQ


friendly-wolfbat:
**Is your feature request related to a problem? Please describe.**
The Technical FAQ in [docs/src/pages/developers/technical-faq/index.mdx](https://github.com/alephdata/aleph/blob/develop/docs/src/pages/developers/technical-faq/index.mdx) and https://docs.aleph.occrp.org/developers/technical-faq/ mentions `convert-document`, which has been removed per https://github.com/alephdata/aleph/pull/2755. References to this should be removed because right now, users reading the documentation will be confused and unable to apply the steps from the FAQ.

**Describe the solution you'd like**
1. Remove references to the `convert-document` image
2. Clarify on how many workers and ingest files to create relative to the number of cores. As of right now, the implication is that, given an eight core system, there should be eight `ingest-file` containers, four `convert-document` containers and two `workers`. Depending on how `convert-document` was deprecated, the documentation should say that only four or two workers are recommended in an eight core environment
3. The "document convert service keeps crashing on startup" question must be updated.

**Describe alternatives you've considered**
The alternative is to leave the documentation as is. Doing so will leave it inconsistent with the current state of the stable version of Aleph, negatively impacting user experience. 

**Additional context**
See pull request https://github.com/alephdata/aleph/pull/2755

tillprochaska:
Hi @friendly-wolfbat, thanks for opening this issue. You are correct, the documentation isn’t up to date in this regard. We do plan to address this as part of a larger overhaul of the technical documentation. But this may take some time, so we should probably address the issue you brought up earlier than that.

friendly-wolfbat:
Hi @tillprochaska, thanks for responding. If you can clarify on points 2 and 3 briefly (or link me to some code where I can find the answer), I can put together a pull request for this particular issue to correct the documentation.

stchris:
Hi @friendly-wolfbat , thanks for noticing and for offering to fix this. 

1. You're totally right, we missed a few spots where `convert-document` is still mentioned. 
2. We have added a section on scaling workers in general [here](https://docs.aleph.occrp.org/developers/installation/#scaling-workers), so it'd be best to link to that.
3. That's right, that section should stay, because it provides a useful workaround, but the references to the "convert-document service" should probably should just mention "document convert operations" (in `ingest-file`).

We're very happy to review your contribution as a PR, but otherwise I can take care of these changes, just let me know. Thanks either way!

[...]

PR #3681, opened 2024-04-16
Remove references to convert-document

friendly-wolfbat:
Per https://github.com/alephdata/aleph/issues/3680#issuecomment-2057288614, I am attempting a pull request to remove references to the old `convert-document` container, which has since been consolidated into `ingest-file` (see https://github.com/alephdata/aleph/pull/2755).

This is my first pull request and I am open to learning! Please let me know if anything is wrong or if I need to make any adjustments.

### DIFF
--- a/docs/src/pages/developers/technical-faq/index.mdx
+++ b/docs/src/pages/developers/technical-faq/index.mdx
@@ -121,7 +121,15 @@ If you're encountering this issue in production mode, try to check the worker lo
 
 ## How can I make imports run faster?
 
-Please see the [Scaling Workers](/developers/installation.mdx#scaling-workers) section in the installation guide.
+The included `docker-compose` configuration for production mode has no understanding of how powerful your server is. It will run just a single instance of the services involved in data imports, `worker` and `ingest-file`. While by default, both workers will create as many threads as Python's [multiprocessing.cpu_count()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count) reports, this may not achieve the speed you desire.
+
+The easiest way to speed up processing is to scale up those services. Make a shell script to start docker-compose with a set of arguments like this:
+
+```bash
+docker-compose up --scale ingest-file=8 --scale worker=4
+```
+
+The number of `ingest-file` and `worker` processes could be the number of CPUs in your machine. Since performance and scaling depends a lot on types of workloads you are seeing you may need to review and adjust these settings accordingly. Please see the [Scaling Workers](/developers/installation.mdx#scaling-workers) section in the installation guide.
 
 ## ElasticSearch will not start. What's wrong?
 
--- a/docs/src/pages/developers/technical-faq/index.mdx
+++ b/docs/src/pages/developers/technical-faq/index.mdx
@@ -121,15 +121,7 @@ If you're encountering this issue in production mode, try to check the worker lo
 
 ## How can I make imports run faster?
 
-The included `docker-compose` configuration for production mode has no understanding of how powerful your server is. It will run just a single instance of the services involved in data imports, `worker` , `ingest-file` and `convert-document`.
-
-The easiest way to speed up processing is to scale up those services. Make a shell script to start docker-compose with a set of arguments like this:
-
-```bash
-docker-compose up --scale ingest-file=8 --scale convert-document=4 --scale worker=2
-```
-
-The number of `ingest-file` processes could be the number of CPUs in your machine, and `convert-document` needs to be scaled up for imports with many office documents, but never higher than `ingest-file`.
+Please see the [Scaling Workers](/developers/installation.mdx#scaling-workers) section in the installation guide.
 
 ## ElasticSearch will not start. What's wrong?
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -121,15 +121,7 @@ If you're encountering this issue in production mode, try to check the worker lo
 
 ## How can I make imports run faster?
 
-The included `docker-compose` configuration for production mode has no understanding of how powerful your server is. It will run just a single instance of the services involved in data imports, `worker` , `ingest-file` and `convert-document`.
```
stchris, line 124: I'd leave this section in, because it still holds true. Just the reference to `convert-document` needs to go.

docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -121,15 +121,7 @@ If you're encountering this issue in production mode, try to check the worker lo
...
-
-The easiest way to speed up processing is to scale up those services. Make a shell script to start docker-compose with a set of arguments like this:
-
-```bash
-docker-compose up --scale ingest-file=8 --scale convert-document=4 --scale worker=2
```
stchris, line 129: `convert-document` needs to go here as well, but the rest can stay

docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -270,12 +262,12 @@ All of this said, we'd really love to hear about any experiments regarding this.
   We have well-defined graph semantics for FollowTheMoney data and you can export any data \(including documents like emails\) in Aleph [into various graph formats](/developers/followthemoney/ftm#exporting-data-to-a-network-graph) \(RDF, Neo4J, and GEXF for Gephi\).
 </Callout>
 
-## The document converter service keeps crashing on startup, what's wrong?
+## The `ingest-file` container keeps crashing on startup, what's wrong?
```
stchris, line 265: The changes in this section look good. Thanks!

docs/src/pages/developers/technical-faq/index.mdx
```diff
@@ -121,15 +121,7 @@ If you're encountering this issue in production mode, try to check the worker lo
...
 
-The included `docker-compose` configuration for production mode has no understanding of how powerful your server is. It will run just a single instance of the services involved in data imports, `worker` , `ingest-file` and `convert-document`.
-
-The easiest way to speed up processing is to scale up those services. Make a shell script to start docker-compose with a set of arguments like this:
-
-```bash
-docker-compose up --scale ingest-file=8 --scale convert-document=4 --scale worker=2
-```
-
-The number of `ingest-file` processes could be the number of CPUs in your machine, and `convert-document` needs to be scaled up for imports with many office documents, but never higher than `ingest-file`.
```
stchris, line 132: Suggestion:

```
The number of `ingest-file` processes could be the number of CPUs in your machine. Since performance and scaling depends a lot on types of workloads you are seeing you may need to review and adjust these settings accordingly.
```

followed by what you wrote:

```
Please also see the [Scaling Workers](/developers/installation.mdx#scaling-workers) section in the installation guide.
```

stchris (changes requested):
I left a few remarks, thanks for cleaning this up!

docs/src/pages/developers/technical-faq/index.mdx
friendly-wolfbat: Left it back in!

friendly-wolfbat:
Thanks so much for reviewing. I believe I made all the changes requested. 

A couple of unrelated things I noticed:
1. Looks like the documentation is referring to `docker compose` as `docker-compose`. The latter is for version 1 of Docker Compose, which has since been [deprecated](https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/). Should I make a separate pull request to remove the dash from the `docker compose` commands?
2. In the installation.mdx file, there are references to `aleph-worker` containers; they are also referred to as `worker`. Should they be consistent, or are they different? Which should the documentation be using?

stchris:
Thanks for your changes, I will review them shortly!

>     1. Looks like the documentation is referring to `docker compose` as `docker-compose`. The latter is for version 1 of Docker Compose, which has since been [deprecated](https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/). Should I make a separate pull request to remove the dash from the `docker compose` commands?

Ah yes, good observation. I believe for consistency we should do this in a separate PR across all of our documentation.

>     2. In the installation.mdx file, there are references to `aleph-worker` containers; they are also referred to as `worker`. Should they be consistent, or are they different? Which should the documentation be using?

Another good observation. So if one uses our Helm chart everything would be prefixed with `aleph-`, but since we are not referring to `api` or `ingest-file` containers with that prefix we should be consistent here as well and call it `worker`. And that should ideally be a PR of its own as well.

stchris (approved):

stchris:
Thanks for your contribution, @friendly-wolfbat

friendly-wolfbat:
@stchris Hey, I'm sorry, it looks like the hyperlink I had to the installation page is wrong, when looking at the "Checks" tab. Should I have put a slash before the hashtag?

PR #3701, opened 2024-04-23
Fix broken link in docs

stchris:
- fixes a broken link from #3681
- enables running the docs workflow when PRs make changes to docs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,6 +6,8 @@ on:
       - "docs/**"
     branches:
       - "**"
+    pr:
+      - "docs/**"
   workflow_dispatch: {}
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -6,6 +6,8 @@ on:
       - "docs/**"
     branches:
       - "**"
+    pr:
+      - "docs/**"
```
tillprochaska, line 10: I’m not entirely sure this is working as expected. I tried [opening a PR from a fork](https://github.com/alephdata/aleph/pull/3703) but the workflow isn’t triggered.

I think maybe all that’s required is to switch from the `push` trigger to the `pull_request` trigger?

```yml
on:
  pull_request:
    paths:
      - "docs/**"
  workflow_dispatch: {}
```

If I understand the docs correctly, this should trigger the workflow whenever a pull request is opened, updated etc. and includes changes in the `docs` subdirectory.

In addition, we could add a trigger to also run the workflow after changes have been merged into the `main` or `develop` branch:


```yml
on:
  pull_request:
    paths:
      - "docs/**"
  push:
    paths:
      - "docs/**"
    branches:
      - "develop"
      - "main"
  workflow_dispatch: {}
```

PR #3724, opened 2024-05-07
Use default language when Accept-Language header is '*'

### LINKED ISSUE #3704: BUG: Error messages in Russian when Accept-Language header is *
**Describe the bug**
When making an HTTP request to the Aleph API with the Accept-Language header set to `*` (as seems to be the default in the node.js implementation of `fetch()`), any error messages will be in Russian.

**To Reproduce**
Assuming Aleph is running on port 8080:

```sh
curl -X "POST" "http://localhost:8080/api/2/collections"
# {"status": "error", "message": "You are not authorized to do this.", "roles": [1]}

curl -X "POST" "http://localhost:8080/api/2/collections" -H 'Accept-Language: *'
# {"status": "error", "message": "\u0423 \u0432\u0430\u0441 \u043d\u0435\u0434\u043e\u0441\u0442\u0430\u0442\u043e\u0447\u043d\u043e \u043f\u0440\u0430\u0432.", "roles": [1]}
# (That’s "У вас недостаточно прав." if you don’t speak fluent Unicode)
```

**Expected behavior**
Requests with `Accept-Language: *` should be treated as if the header was not set at all.

(You could argue that the semantics of `*` are “Any language will do” while a missing header indicates “Give me the default”, but since this is caused by the default behaviour of node.js’s standard library, I imagine I won’t be the only one who is going to find themselves unexpectedly told off in Russian.)

**Aleph version**
3.15.5

stchris:
Fixes #3704

### DIFF
--- a/aleph/core.py
+++ b/aleph/core.py
@@ -46,7 +46,13 @@ prometheus = PrometheusExtension()
 def determine_locale():
     try:
         options = SETTINGS.UI_LANGUAGES
-        locale = request.accept_languages.best_match(options)
+        if str(request.accept_languages) == "*":
+            # For #3704: if any language is fine pick the default one
+            locale = SETTINGS.DEFAULT_LANGUAGE
+        else:
+            locale = request.accept_languages.best_match(
+                options, default=SETTINGS.DEFAULT_LANGUAGE
+            )
         locale = locale or str(babel.default_locale)
     except RuntimeError:
         locale = str(babel.default_locale)


aleph/core.py
```diff
@@ -46,7 +46,11 @@
 def determine_locale():
     try:
         options = SETTINGS.UI_LANGUAGES
-        locale = request.accept_languages.best_match(options)
+        if str(request.accept_languages) == "*":
```
tillprochaska, line 49: Just a quick note: I haven’t tested this, but I saw that `best_match` has a default argument which is returned if none of the options matches the `Accept-Language` header. If that covers this case as well, it might be a little bit more robust as it would also cover cases where `Accept-Language` contains a list of languages but none of them matches one of the options.

https://github.com/pallets/werkzeug/blob/main/src/werkzeug/datastructures/accept.py#L148

stchris: Unfortunately it does not cover this case, because it immediately exits with the first found value. But I can add the default language to the `best_match` invocation for good measure.

stchris: (done)

tillprochaska: Nice, thanks for checking!

tillprochaska (approved):
Thanks for the quick fix, works perfectly! :)

pbock:
Fantastic, thank you!

PR #3728, opened 2024-05-08
Automatically post releases to Discourse

stchris:
This is based on https://github.com/roots/discourse-topic-github-release-action

### DIFF
--- a/.github/workflows/discourse.yml
+++ b/.github/workflows/discourse.yml
@@ -1,0 +1,18 @@
+name: Post release topic on Discourse
+
+on:
+  release:
+    types: [published]
+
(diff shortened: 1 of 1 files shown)


tillprochaska (approved):
I haven’t tested this myself, but it sounds like a good way to automate a tedious manual task!

One thing I’d note is that it might be sensible to pin GitHub Actions from third parties to a commit hash [as recommended here](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). (We don’t do this consistently yet everywhere, but I think it would be a good thing to start doing.)

PR #3734, opened 2024-05-20
Topic exchange

stchris:


catileptic:
Decided not to implement it this way.

PR #3746, opened 2024-05-23
Fix end-to-end tests in CI

tillprochaska:
The `docker compose run` command doesn’t support the `--build-arg` argument. Depending on the Docker Compose version installed, using the `--build-arg` argument would either exit with an error code or simply print the command help.

To achieve the desired outcome, we can first build the image with the build argument, then start the container. This has the additional advantage of separating the build logs from the logs of the actual command execution in the GitHub Actions UI.

Tested these changes: https://github.com/alephdata/aleph/actions/runs/9205541034/job/25321430312

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -30,6 +30,9 @@ jobs:
           docker --version
           docker-compose --version
 
+      - name: Pull and start services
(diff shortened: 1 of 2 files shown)


stchris (approved):
Nice, thanks for this!

PR #3752, opened 2024-05-28
Remove bookmarks migration

### LINKED ISSUE #2864: Remove support for local bookmarks/migration
Do not implement this before 2023-12-31.

When we first implemented bookmarks as an experimental feature, they were stored in local storage. People started using the feature so we decided to continue working on bookmarks to build it out into a mature feature.

Most importantly, this included storing bookmarks on the server-side in order to support growing numbers of bookmarks per user and to prevent users from losing their bookmarks when clearing browsing data/switching devices.

We also needed to migrate bookmarks that were currently stored locally to the server side. That means the code base now has some code that only handles migrating the "legacy" local bookmarks to the server.

We should remove code related to local bookmarks and the migration after X months on 2023-12-31.

tillprochaska:
When we first implemented bookmarks as an experimental feature, they were stored in local storage. People started using the feature so we decided to continue working on bookmarks to build it out into a mature feature.

Most importantly, this included storing bookmarks on the server-side in order to support growing numbers of bookmarks per user and to prevent users from losing their bookmarks when clearing browsing data/switching devices.

We also needed to migrate bookmarks that were currently stored locally to the server side. That means the code base had quite some code that only handled migrating the "legacy" local bookmarks to the server. As it’s been a while since we started migrating to server-side bookmarks, it should be safe to remove that migration logic now.

Fix #2864


Rosencrantz (commented):
This is the most baller pull request I have ever seen

Rosencrantz (approved):

catileptic (approved):
LGTM :) great work, thank you for returning to this code and removing it

PR #3778, opened 2024-06-14
download actual files instead of HTML wrapped configuration

vsessink:
refer to alephdata/aleph/raw/ to get actual files


tillprochaska (approved):
Thanks a lot for your PR!

PR #3779, opened 2024-06-17
Order of commands made more clean

vsessink:
"shell aleph upgrade" is mandatory for 1st time installs.


tillprochaska (approved):
Thank you, this makes it indeed much clearer!

Issue #3787, opened 2024-06-27
BUG: Information in JSON for an ingesting dataset does not seem to be entirely correct


Rosencrantz:
**Describe the bug**
From the status page when reviewing active ingest jobs the start_time,  end_time, and last_updated don't seem to reflect what I would expect for a dataset.

**Expected behavior**
I would expect start_time to be a static time from when a dataset started being ingested. end_time should show the point at which a dataset completed. last_updated should show the last time a file was ingested, analysed, or indexed.

**Aleph version**
4.0.0 (aleph.occrp.org)

tillprochaska:
Fixed in https://github.com/alephdata/servicelayer/pull/190

PR #3788, opened 2024-06-28
Display start and last updated timestamp on system status page

tillprochaska:
This exposes the two timestamps (that are already returned per collection by the status API) in the UI. Times are relative (for example "45 seconds ago"), but hovering over them displays the full date/time (for example "Fri, Jun 28, 2024, 5:05 PM").

Screen Shot 2024-06-28 at 17 06 20

**Todos:**

- [x] Figure out if there’s an issue with the start time, possibly related to #3787.

### DIFF
--- a/aleph/worker.py
+++ b/aleph/worker.py
@@ -138,7 +138,6 @@ class AlephWorker(Worker):
                     self.often.update()
                     log.info("Self-check...")
                     compute_collections()
-                    Dataset.cleanup_dataset_status(kv)
 
                 if self.daily.check():
                     self.daily.update()
--- a/aleph/worker.py
+++ b/aleph/worker.py
@@ -8,7 +8,7 @@ import copy
 from typing import Dict, Callable
 
 import structlog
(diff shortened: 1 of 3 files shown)


Rosencrantz (approved):
Such a small change!

tillprochaska:
@Rosencrantz I’ve pushed a small update. When testing this with a few larger datasets, I found that the relative times would add a lot of noise. Specifically:

* The rendered relative times update constantly even if the actual values doesn’t change. For example, even if there has been no update to a collection, "last updated now" will become "last updated 3 seconds ago", …
* The length of the rendered values changes frequently which can cause line breaks and layout shifts. This makes it more difficult to visually track a specific row/value, as it’s position on the page changes.

Instead, I’m now displaying just the absolute time (if the date is today) or otherwise the date and time:

Screen Shot 2024-07-04 at 11 19 02

tillprochaska:
This can be merged as soon as https://github.com/alephdata/servicelayer/pull/190 is released

Rosencrantz:
Nioce update. I approve!

Issue #3811, opened 2024-07-11
BUG: Error during OAuth callback / missing constraints for `role_membership` table


tillprochaska:
**Describe the bug**
Some users experience a 500 error when signing in using OAuth.

The `role_membership` table stores which groups a user is part of. It has two columns, `group_id` and `role_id` ([migration](https://github.com/alephdata/aleph/blob/main/aleph/migrate/versions/af9b37868cf3_init.py#L135-L147)). However, it doesn’t define any constraints (composite primary key or a unique constraint, so it is possible to add duplicate rows to the table.

Duplicate rows in this table can cause errors during the OAuth callback. After authenticating a user, Aleph "syncs" the groups the user is part of. It does this by first deleting all rows in the `role_membership` for the user, then adding new rows for all of the groups the users is currently part of. Groups are encoded in the OAuth token returned by the identity provider, e.g. Keycloak.

In case of duplicate rows, there’s a mismatch between the number of rows SQLAlchemy expects to delete and the number of rows that are actually deleted. This results in the following error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/dist-packages/flask_cors/extension.py", line 176, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/aleph/aleph/views/sessions_api.py", line 123, in oauth_callback
    role = handle_oauth(oauth.provider, oauth_token)
  File "/aleph/aleph/oauth.py", line 100, in handle_oauth
    role.clear_roles()
  File "/aleph/aleph/model/role.py", line 109, in clear_roles
    db.session.flush()
  File "/usr/local/lib/python3.8/dist-packages/sqlalchemy/orm/scoping.py", line 897, in flush
    return self._proxied.flush(objects=objects)
[...]

PR #3815, opened 2024-07-17
Do not allow duplicate role memberships

tillprochaska:
Aleph users can be part of user groups. Users and groups are both modeled as "roles" in Aleph. To add a user to a group, a row is inserted into the `role_membership` table.

We’ve occasionally had issues with duplicate rows in this table, likely due to a race condition when Aleph handles parallel OAuth requests for the same user (https://github.com/alephdata/aleph/issues/3811#issuecomment-2232953036), but we haven’t been able to reproduce or pinpoint the exact root cause.

The effect of a duplicate role membership is that users are permanently blocked from authenticating until the duplicates are removed.

This PR adds a primary key constraint to the `role_membership` table. While this does not address the root cause of the problem, it prevents duplicates from being inserted into the table. In practice, this means that in rare cases, OAuth requests may fail when the constraint is violated, but users can simply retry the authentication in that case.

***

In addition, I’ve implemented two related changes:

* I have added tests covering the group sync during OAuth authentication as this was previously completely untested.
* I have added logging in case duplicate role memberships are added explicitly. While it’s unlikely this is the root cause of the issue, it’s helpful as it allows excluding this possibility with certainty if we see the issue again.

***

> [!IMPORTANT]  
> This PR contains a migration that adds a primary key constraint to an existing table. The migration will fail if the existing table violates the constraint. It is necessary to manually check and remove whether duplicates exist in this table before applying the migration. See https://github.com/alephdata/aleph/issues/3811#issuecomment-2223355962 for a query that returns duplicates.

### DIFF
--- a/aleph/model/role.py
+++ b/aleph/model/role.py
@@ -118,8 +118,11 @@ class Role(db.Model, IdModel, SoftDeleteModel):
 
     def add_role(self, role):
         """Adds an existing role as a membership of a group."""
-        self.roles.append(role)
-        db.session.add(role)
+        if role not in self.roles:
+            self.roles.append(role)
+            db.session.add(role)
+        else:
+            log.warning(f"Tried to add duplicate role membership: {self} {role}")
         db.session.add(self)
         self.updated_at = datetime.utcnow()
 
--- a/aleph/tests/test_sessions_api.py
+++ b/aleph/tests/test_sessions_api.py
@@ -102,7 +102,7 @@ class SessionsApiTestCase(TestCase):
 
 class SessionsApiOAuthTestCase(TestCase):
     def setUp(self):
-        super().setUpClass()
+        super().setUp()
 
         SETTINGS.OAUTH = True
         SETTINGS.OAUTH_HANDLER = "test-oidc"
--- a/aleph/tests/test_sessions_api.py
+++ b/aleph/tests/test_sessions_api.py
@@ -2,6 +2,7 @@ from contextlib import contextmanager
 import unittest.mock as mock
 from urllib.parse import urlparse, parse_qs
 from requests import Response
+from typing import List
 
 from aleph.core import db
 from aleph.settings import SETTINGS
@@ -236,8 +237,8 @@ def mock_oauth_token_exchange(name: str, email: str):
 
         # https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
         parse_mock.return_value = {
-            "name": "John Doe",
-            "email": "john.doe@example.org",
+            "name": name,
+            "email": email,
         }
 
         yield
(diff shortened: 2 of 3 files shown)


aleph/model/role.py
```diff
@@ -118,8 +118,11 @@ def clear_roles(self):
 
     def add_role(self, role):
         """Adds an existing role as a membership of a group."""
-        self.roles.append(role)
-        db.session.add(role)
+        if role not in self.roles:
```
stchris, line 121: I wonder if catching https://docs.sqlalchemy.org/en/20/core/exceptions.html#sqlalchemy.exc.IntegrityError here and reacting on it (log.warn) would be more efficient than what I assume is a SELECT followed by an INSERT?

aleph/tests/test_sessions_api.py
```diff
@@ -220,9 +221,78 @@ def test_oauth_callback_blocked_user(self):
...
+            headers={"Authorization": f"Token {auth_token}"},
+        )
+        assert len(res.json["results"]) == 2
+        assert res.json["results"][0]["name"] == "project-b"
+        assert res.json["results"][1]["name"] == "project-c"
+
 
 @contextmanager
-def mock_oauth_token_exchange(name: str, email: str):
+def mock_oauth_token_exchange(name: str, email: str, groups: List[str] = []):
```
stchris, line 295: High praise for adding type hints 🏅

aleph/tests/test_sessions_api.py
```diff
@@ -220,9 +221,78 @@ def test_oauth_callback_blocked_user(self):
         assert query["status"] == ["error"]
         assert query["code"] == ["403"]
 
+    def test_oauth_callback_sync_groups(self):
```
stchris, line 224: The test case looks good, thanks for making it so straight-forward!

stchris (approved):
I'm ok with this implementation, but left a few side questions.

aleph/model/role.py
tillprochaska: This operates on the ORM object only, i.e., no requests are made at this time. Only after flushing the SQLAlchemy session and comitting the transaction statements to update the role memberships would be executed (and any errors would be raised only then).

tillprochaska: The primary goal of this log message was to confirm that there is no logic flaw that can insert duplicate role memberships within a single request. I’m pretty sure there is no such flaw (instead the issue is a race condition with parallel requests) and thus we should never see this log message in practice. But I’m not 100% sure.

tillprochaska:
I suggest we keep this PR open and merge it when we are ready to release/deploy the changes (after a stable 4.0 release), to ensure we do not forget about the breaking changes/manual checks required.

PR #3821, opened 2024-08-01
Fix user auto-suggest input / do not explicitly exclude group IDs 

### LINKED ISSUE #3820: BUG: UI constructs excessively long URLs
**Describe the bug**
The auto-suggest input in the collection access control dialog uses the `/roles/_suggest` API. To not suggest users that already have access to the collection, the UI [passes the IDs](https://github.com/alephdata/aleph/blob/develop/ui/src/dialogs/CollectionAccessDialog/CollectionAccessDialog.jsx#L149) of all users and groups that already have access as a query parameter:

Given a collection that is already shared with users 1, 2, and 3, the API URL would look something like this:

```
/roles/_suggest?prefix=jane.doe&exclude:id=1&exclude_id=2&exclude:id=3
```

**To Reproduce**
1. Create a large number of groups (250+) and add yourself to these groups. You can also temporarily reduce the [Gunicorn request line limit](https://docs.gunicorn.org/en/latest/settings.html#limit-request-line) and will experience the issue with a lower number of groups.
2. Navigate to a collection and open the sharing settings from the settings dropdown.
3. Click on "Choose a user" and start typing"
4. Observe the API requests the UI sends in the browser developer tools network tab. You should see requests to `/api/2/roles/_suggest` with one `exclude:id` query parameter for every group your user is part of. If the total length of the request URI exceeds the Gunicorn limits, the request will fail.

**Expected behavior**
The auto-suggest input should work even for users that are members of many groups.

**Aleph version**
3.15.7, 3.4.0-rc*

**Screenshots**
image

**Additional context**
* This error occurs as long as a user is a member of many groups, no matter whether a collection is actually shared with many/all of them.
* This error would also occur when sharing an investigation with a large number of users, but this a) is something that users have to do explicitly, and b) I don’t think this was ever an issue before.
* Need to double check, but I think `/roles/_suggest` will only ever return users and never groups, so it isn’t necessary to explicitly exclude group IDs.

tillprochaska:
The reason behind excluding specific IDs is to not suggest users and groups that are already associated with the collection. However, the `/roles/_suggest` endpoint returns only users, never groups, so there is no point in explicitly excluding groups. Fixes #3820.

In case a collection is shared with a large (200+) number of individual users, this may cause the same error as reported in #3820. However, this would require explicitly sharing a collection with such a large number of users whereas #3820 affected any collection, even if it wasn’t shared with anyone. Also, I don’t think there has been a use case for sharing a collection with so many individual users, and probably using groups would be more appropriate in such a situation.

You can verify this by adding yourself to a few user groups:

```
aleph creategroup test1
aleph creategroup test2
aleph creategroup test3
aleph useradd test1 user@example.org
aleph useradd test2 user@example.org
aleph useradd test3 user@example.org
```

* Log out and in to recreate your session.
* Create a new investigation and open the share dialog.
* Open the network tab in the browser developer tools.
* Type something into the auto-suggest input.
* You should be able to see a request to `/roles/_suggest?exclude:id=123&prefix=test` (where `123` is the ID of your user). There shouldn’t be any additional `exclude:id` params.

### DIFF
--- a/ui/src/dialogs/CollectionAccessDialog/CollectionAccessDialog.jsx
+++ b/ui/src/dialogs/CollectionAccessDialog/CollectionAccessDialog.jsx
@@ -146,10 +146,11 @@ class CollectionAccessDialog extends Component {
       return null;
     }
 
-    const exclude = permissions.map((perm) => perm.role.id);
     const systemRoles = this.filterPermissions('system');
     const groupRoles = this.filterPermissions('group');
     const userRoles = this.filterPermissions('user');
+    const exclude = userRoles.map((perm) => perm.role.id);
+
     return (
       <FormDialog
         processing={blocking}


stchris (approved):
Thanks for the investigation! The fix makes total sense to me.

PR #3841, opened 2024-08-05
Improve date formats on status page

tillprochaska:
* Previously, only the time was displayed if the start time/last update was on the same day. This could be confusing, so we now display "today" in these cases.
* The default US English short date format is `MM/DD/YY` which can be confusing, especially for users outside the US. I've replace this with a less ambiguous date format: Aug 5, 2024.

Screen Shot 2024-08-05 at 11 45 33


Rosencrantz (approved):
No notes

PR #3844, opened 2024-08-12
New and updated Prometheus metrics

tillprochaska:
This PR adds a few new metrics and extends existing metrics based on feedback. I’d recommend taking a look at the individual commits when reviewing this PR.

Corresponding documentation changes are in #3845 because the `release/4.0.0` branch doesn’t yet have the recently updated tech docs from `develop`.

### Users

I’ve added new `active` and `locale` labels to the `aleph_users` metric. This allows querying for the number of active users (e.g. within last 24h, 7d, 30d, …) and the number of users that use a specific UI locale.

```
# Number of all users
aleph_users{active="all"}

# Number of all users using the (default) English localization
aleph_users{locale="en", active="all"}

# Number of monthly active users using the German localization
aleph_users{active="30d", locale="de"}
```

### Collections

Previously, there was a single `aleph_collections` metric that allowed querying for the number of collections by category (e.g. "casefile", "leak", "gazette", …). I have renamed this metric to be consistent with two new metrics that expose the number of collections by country and language.

```
# Number of leaks
aleph_collection_categories{category="leak"}

# Number of datasets or investigations
aleph_collection_categories{type="dataset"}
aleph_collection_categories{type="investigation"}

# Number of datasets or investigations by country
aleph_collection_countries{type="dataset", country="fr"}
aleph_collection_countries{type="investigation", country="fr"}

# Number of datasets or investigations by language
aleph_collection_languages{type="dataset", language="fra"}
aleph_collection_languages{type="investigation", language="fra"}
```

### Entities

I’ve added a new metric to track the number of entities per schema type. This re-uses collection statistics that are pre-computed and cached in Redis, i.e. we do not send any (potentially expensive) ES aggregation queries during metrics collection.

```
# Number of `Person` entities
aleph_entities{schema="Person"}
```
### API HTTP requests

I have added a new label `auth_method` to the `aleph_http_requests_total`. This allows us to query for the number of API requests that used an API key vs. a UI session token.

```
# Number of requests to `/api/2/entities` using an API key
aleph_http_requests_total{api_endpoint="entities.index", auth_method="api_key"}
```

### DIFF
--- a/aleph/authz.py
+++ b/aleph/authz.py
@@ -26,6 +26,7 @@ class Authz(object):
     def __init__(self, role_id, roles, is_admin=False, token_id=None, expire=None):
         self.id = role_id
         self.logged_in = role_id is not None
+        self.auth_method = None
         self.roles = set(roles)
         self.is_admin = is_admin
         self.token_id = token_id
--- a/aleph/metrics/collectors.py
+++ b/aleph/metrics/collectors.py
@@ -1,3 +1,5 @@
+import datetime
+
 from sqlalchemy import func
 from prometheus_client.core import GaugeMetricFamily, InfoMetricFamily
 from prometheus_client.registry import Collector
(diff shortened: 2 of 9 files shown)


tillprochaska:
@catileptic This is now ready for review, even though one test is still failing. As discussed, this test fails because of an update of the `phonenumbers` Python package. The new version of the package has validation changes, and as a result phone numbers used in that particular test are determined to be invalid which causes the test to fail.

I haven’t yet understood completely whether `phonenumbers` is actually right about thes test phone numbers and the numbers are indeed invalid or if this is a regression and they should be treated as valid. At first glance they look like a reasonable German mobile phone number, but phone number prefixes in Germany are weird.

I’ll keep the test as is until I’ve found a definitive answer, but that shouldn’t block a review of the actual changes proposed in this PR.

catileptic (approved):
This looks good to me. Really appreciate adding the tests! Let's get this merged.

PR #3852, opened 2024-08-21
Fix active users metric

tillprochaska:
When implementing the active users metric (#3844), I noticed that Aleph currently updates the `role.updated_at` column every time a user logs in – so I took the shortcut and simply used that to count the number of users that logged in at least once within a specific period of time. But it turns out that `updated_at` is only updated on login when using password auth. It isn’t updated when signing in via OAuth, and consequently the active users metric is incorrect when OAuth is used for authentication.

I’m fixing this by adding a separate timestamp column that stores when a user last logged in. It’s probably a good idea to make the intent clear (Just reading the name updated_at, most people probably wouldn’t expect that timestamp to be set every time a user logs in.)

Note: This currently doesn’t include a test for the expected behavior when logging in via OAuth. There have been lots of changes to authentication testing on the `develop` branch that are not present in `release/4.0.0`, so any changes would lead to annoying merge conflicts later on. I have created a follow-up task to add a separate test for OAuth logins once the 4.0.0 release is out. You can manually test the behavior by [configuring your development environment to use an OAuth provider](https://docs.aleph.occrp.org/developers/how-to/development/identity-provider/).

### DIFF
--- a/aleph/metrics/collectors.py
+++ b/aleph/metrics/collectors.py
@@ -72,7 +72,7 @@ class DatabaseCollector(Collector):
 
             if delta:
                 limit = now - delta
-                query = query.filter(Role.updated_at >= limit)
+                query = query.filter(Role.last_login_at >= limit)
             else:
                 label = "all"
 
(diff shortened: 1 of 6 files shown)


catileptic (approved):
Looks good to me :) LGIM. Thanks for circling back to this!

PR #3855, opened 2024-08-25
Implement infinite scrolling for document search

tillprochaska:
When searching a large document in a dataset which has a significant number of results, only the first set of results are shown.

I’ve fixed this by implementing infinite scrolling in the same way we already implement this for other views (such as the main search view). Now, when a user has scrolled to the end of the search results list and if there are more results, the next page of results is automatically loaded.

In addition, I’ve used the opportunity to do some cleanup and remove dead code from the component responsible for displaying document search results.


stchris (approved):
Works smoothly 👍

PR #3858, opened 2024-08-28
aleph CLI command to downgrade the postgres DB

stchris:


aleph/logic/notifications.py
```diff
@@ -24,7 +24,7 @@
 
 def channel_tag(obj, clazz=None):
     clazz = clazz or type(obj)
-    if clazz == str:
+    if clazz is str:
```
stchris, line 27: This is not related, but my local ruff installation complained about it 🤷

aleph/manage.py
```diff
@@ -504,6 +504,19 @@ def upgrade():
...
+@click.option(
+    "--revision",
+    "-r",
+    default="-1",
+    show_default=True,
+    help="target revision number. Can also be 'head' or '-1'",
+)
+def downgrade_database(revision):
+    """Downgrade the database."""
+    downgrade_db(revision)
```
tillprochaska, line 517: Running database downgrades can be quite scary as they can delete data. The core alembic CLI has a dry-run option and I just checked and Flask-Migrate exposes that, too.

```suggestion
@cli.command()
@click.option(
    "--revision",
    "-r",
    default="-1",
    show_default=True,
    help="target revision number. Can also be 'head' or '-1'",
)
@click.option(
    "--sql",
    is_flag=False,
    is_flag=True,
    help="...",
)
def downgrade_database(revision, sql):
    """Downgrade the database."""
    downgrade_db(revision, sql)
```

Something like this should print the SQL statements the migration would execute without actually executing it, as well as the names of the migrations that will be rolled back.

This would allow us to double check that the downgrade is doing what we expect before actually executing it.

tillprochaska (commented):
Thank you, works as expected :) Added one suggestion about a dry-run option below.

aleph/manage.py
tillprochaska: When I tested this locally, I accidentally rolled back a different migration than I expected… 😄

stchris: I made these changes:

- the default is a dry run, which prints a warning that this doesn't actually execute, the migration and the SQL statements and quits.
- you now have to pass `--execute` to actually run the migrations.

catileptic (approved):

PR #3863, opened 2024-09-09
Fix ES index upgrades when using index aliases

tillprochaska:
ES supports fetching the index config for multiple indices in a single request. We don’t make use of this, but for this reason the ES responses are an object with a key for each requested index. For example:

```json
{
  "aleph-entity-company-v1": {
    "settings": { ... },
    "mappings: { ... }
  }
}
```

In case an index is requested using an alias, the key will be the original index name and *not* the alias, i.e. we cannot rely on the key being the name that was used to request the index, as this would fail in case aliases are used.

However, as we’re always requesting only a single index at a time, the response should always contain data for one index, so we can extract that without relying on the name of the index.

### DIFF
--- a/aleph/index/util.py
+++ b/aleph/index/util.py
@@ -289,7 +289,27 @@ def configure_index(index, mapping, settings):
             "timeout": MAX_TIMEOUT,
             "master_timeout": MAX_TIMEOUT,
         }
-        config = es.indices.get(index=index).get(index, {})
+        res = es.indices.get(index=index)
+
+        if len(res) == 0:
+            # This case should never occur.
+            raise AlephOperationalException(
+                f"ES response does not contain data for index {index}"
+            )
+
+        if len(res) > 1:
+            # This case should never occur.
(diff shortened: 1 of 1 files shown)


aleph/index/util.py
```diff
@@ -289,7 +289,27 @@ def configure_index(index, mapping, settings):
...
-        config = es.indices.get(index=index).get(index, {})
+        res = es.indices.get(index=index)
+
+        if len(res) == 0:
+            # This case should never occur.
+            raise AlephOperationalException(
+                f"ES response does not contain data for index {index}"
+            )
+
+        if len(res) > 1:
```
stchris, line 300: out of curiosity: how do you feel about merging these checks into a `if len != 1`?

tillprochaska: Fair point, I’ve pushed this.

stchris (approved):
Looks and works great, thanks for the fix!

PR #3865, opened 2024-09-17
Allow sharing of investigations

tillprochaska:
This enables sharing of investigations with individual users by providing their full email address. Main changes are in 7cced50eaa0ff09eca17dba5c977c9b6814ac6c7.

If you want to manually test cases including OAuth user groups, you can set up Keycloak in your development environment like this: https://docs.aleph.occrp.org/developers/how-to/development/identity-provider/

### DIFF
--- a/aleph/tests/test_permissions_api.py
+++ b/aleph/tests/test_permissions_api.py
@@ -1,0 +1,103 @@
+from aleph.tests.util import TestCase
+from aleph.logic.roles import create_group
+
+
+class PermissionsApiTestCase(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.role, self.headers = self.login(
+            foreign_id="john",
+            name="John Doe",
+            email="john.doe@example.org",
+        )
+        self.col = self.create_collection(creator=self.role)
+
+    def test_update(self):
+        jane = self.create_user(
+            foreign_id="jane",
+            name="Jane Doe",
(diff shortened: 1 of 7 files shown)


aleph/tests/test_permissions_api.py
```diff
@@ -0,0 +1,103 @@
+from aleph.tests.util import TestCase
+from aleph.logic.roles import create_group
+
+
+class PermissionsApiTestCase(TestCase):
```
stchris, line 5: Thanks for this! 🙏

aleph/views/roles_api.py
```diff
@@ -51,12 +54,26 @@ def suggest():
       - Role
     """
     require(request.authz.logged_in)
-    return jsonify(
-        {
-            "results": [],
-            "total": 0,
-        }
+    parser = QueryParser(request.args, request.authz, limit=10)
+    if parser.prefix is None or len(parser.prefix) < 3:
```
stchris, line 58: I guess we could raise minimum length to 5 or 6 here, but it's a minor thing.

stchris (approved):
LGTM, thanks for this! Left some minor remarks which are fine to be considered later (or not at all)

aleph/views/roles_api.py
tillprochaska: Done!

PR #3869, opened 2024-09-25
feat: Optimize helm templates to expose multiple secret values as env variables

ps-occrp:
Optimize helm templates to expose multiple secret values as environment variables, this has side effect that all values in `aleph-secrets` are exposed to deployment


stchris:
Closing this as discussed @ps-occrp .

PR #3870, opened 2024-09-26
Merge main into release/4.0.0

tillprochaska:


.github/workflows/build.yml
```diff
@@ -34,6 +34,8 @@ jobs:
 
       - name: Install development dependencies
         run: make dev
+        env:
+          PIP_BREAK_SYSTEM_PACKAGES: 1
```
tillprochaska, line 38: @stchris @catileptic  I started seeing CI runs fail with "error: externally-managed-environment".

While in general it’s bad practice to install packages system-wide as this may interfere with system-managed packages, this is much less problematic in the one-off runners in CI.

As a workaround, I’ve disabled the warning when installing dependencies in CI.

PR #3875, opened 2024-10-15
Fix edge case in metrics collector

tillprochaska:
For some reason, the default/fallback value returned by `get_collection_stats` is an empty list, not an empty object. This can trigger an edge case if Prometheus metrics are scraped after a collection has been created, but the stats for the new collection haven’t been computed yet.

To prevent unintended side effects, I haven’t changed the default/fallback value (as other parts of the code base might rely on it) and have instead explicitly handled empty values.


stchris (commented):
Looks good to me! Thanks for the fix!

PR #3878, opened 2024-10-15
Add test to cover `last_login_at` timestamp in OAuth workflow

tillprochaska:
This is a follow-up to https://github.com/alephdata/aleph/pull/3852. This PR above didn’t include a test for the expected behavior when logging in via OAuth. There have been a lot of changes to authentication testing on the develop branch that were not present in the 4.0 release branch, so any changes would have lead to annoying merge conflicts down the line.

But now that the 4.0 release is integrated, it’s time to add a proper test :)


stchris (approved):
Ah, thanks for coming back to add this. Much appreciated, Till!

PR #3879, opened 2024-10-17
Make document search available in entity preview

tillprochaska:
We frequently receive feedback about the search flow requiring too many clicks in the following use case:

1. User searches for a term.
2. User finds a document (e.g. PDF, Word) that might be relevant in the search results.
3. They click on the search results and the preview is displayed.
4. In case of documents that are longer than a few pages, most users don’t want to go over the entire document, instead they want to skim the relevant sections.
5. In order to find relevant sections, users currently need to click on the "Expand" button in the entity preview. After clicking expand, the document search is available.
7. Users then need to enter the search term again in the document search input.

This PR simplifies this flow as the document search is now available from within the entity previews. That means users can click on a search results, search within the document inside of the preview, and can easily return to the search results.

In a follow-up PR, I plan to prefill the search input for the document search with the value of the previous search or provide another easy way to reuse the search query so users don’t have to retype the entire query.

https://github.com/user-attachments/assets/b17f0712-5c0e-439d-a980-989a2141a3b1

From an implementation perspective, the main changes are in https://github.com/alephdata/aleph/pull/3879/commits/52c706c3a45ccac662a8601c86d6bd701fbb6fe7. In addition to these changes, I’ve used the opportunity to refactor the `EntityViews` component as it was quite big and slightly confusing.

For this reason, I’d recommend reviewing the commits individually. The commit messages also provide additional details.

### DIFF
--- a/ui/src/components/Entity/EntityViews.jsx
+++ b/ui/src/components/Entity/EntityViews.jsx
@@ -1,13 +1,6 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
-import {
-  Callout,
-  Classes,
-  Tab,
-  Tabs,
-  Icon,
-  InputGroup,
-} from '@blueprintjs/core';
+import { Callout, Classes, Tab, Tabs, InputGroup } from '@blueprintjs/core';
 import queryString from 'query-string';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
@@ -168,14 +161,8 @@ class EntityViews extends React.Component {
     return (
       <Tab
         id="info"
-        title={
-          <>
-            <Icon icon="info" className="left-icon" />
-            <span className="tab-padding">
-              <FormattedMessage id="entity.info.info" defaultMessage="Info" />
-            </span>
-          </>
-        }
+        icon="info"
+        title={<FormattedMessage id="entity.info.info" defaultMessage="Info" />}
         panel={<EntityProperties entity={entity} />}
       />
     );
@@ -198,12 +185,8 @@ class EntityViews extends React.Component {
     return (
       <Tab
         id="view"
-        title={
-          <>
-            <Icon icon="documentation" className="left-icon" />
-            <FormattedMessage id="entity.info.view" defaultMessage="View" />
-          </>
-        }
+        icon="documentation"
+        title={<FormattedMessage id="entity.info.view" defaultMessage="View" />}
         panel={<DocumentViewMode document={entity} activeMode={activeMode} />}
       />
     );
@@ -219,12 +202,8 @@ class EntityViews extends React.Component {
     return (
       <Tab
         id="text"
-        title={
-          <>
-            <Icon icon="plaintext" className="left-icon" />
-            <FormattedMessage id="entity.info.text" defaultMessage="Text" />
-          </>
-        }
+        icon="plaintext"
+        title={<FormattedMessage id="entity.info.text" defaultMessage="Text" />}
         panel={<DocumentViewMode document={entity} activeMode={activeMode} />}
       />
     );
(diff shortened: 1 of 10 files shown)


ui/src/viewers/PdfViewerSearch.jsx
```diff
@@ -32,16 +38,19 @@ class PdfViewerSearch extends Component {
   }
 
   getResultLink(result) {
-    const { activeMode, query } = this.props;
-    const page = result.getProperty('index').toString();
+    const { activeMode, query, document, location, isPreview } = this.props;
```
tillprochaska, line 41: This is unused and should be removed

ui/src/components/Entity/EntityViews.jsx
```diff
@@ -57,215 +59,345 @@ class EntityViews extends React.Component {
...
+
+    return (
+      <Tab
+        id="search"
+        title={<span className="visually-hidden">Search results</span>}
+        panel={
+          <PdfViewerSearch
+            isPreview={isPreview}
+            document={entity}
+            activeMode={activeMode}
```
tillprochaska, line 369: This is unused and should be removed.

ui/src/viewers/PdfViewerSearch.jsx
```diff
@@ -70,13 +83,20 @@ class PdfViewerSearch extends Component {
...
               <p dir={dir}>
-                <Link to={this.getResultLink(res)}>
+                <a href={this.getResultLink(res)}>
                   <span
-                    className={c(Classes.ICON, `${Classes.ICON}-document`)}
+                    className={c(
+                      Classes.ICON,
+                      `${Classes.ICON}-document`,
+                      'PdfViewerSearch__icon'
+                    )}
```
tillprochaska, line 99: Why doesn’t this use the `Icon` component?

ui/src/components/Entity/EntityViews.jsx
```diff
@@ -57,215 +59,345 @@ class EntityViews extends React.Component {
...
+  }
+
+  renderReferenceModes() {
+    const { entity, references, reference, referenceQuery, activeMode } =
+      this.props;
+
+    if (!references.total && references.isPending) {
+      return <Tab id="loading" title={<TextLoading loading={true} />} />;
+    }
+
```
tillprochaska, line 250: I think this is unnecessary as this is never rendered if references are still being loaded.

ui/src/components/Entity/EntityViews.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
-import { Callout, Tab, Tabs, Icon } from '@blueprintjs/core';
+import { Callout, Classes, Tab, Tabs, InputGroup } from '@blueprintjs/core';
```
tillprochaska, line 3: `Classes` is unused

ui/src/viewers/PdfViewerSearch.jsx
tillprochaska: For future reference:

The `document` icon from Blueprint looks like this: a single page. That’s what we want in this case.

Screenshot 2024-10-23 at 12 53 28

However, we also have a custom set of icons, e.g. for the different FtM schemata which can be used with the `Icon` component in addition to the default Blueprint icons. As there is an FtM schema called `Document`, when using `<Icon icon="document" />` you get this custom icon which is a multi-page document:

Screenshot 2024-10-23 at 12 53 03

Because both the custom schema icon and the Blueprint icon are named `document`, this seems to be a workaround to get the default single-page icon from Blueprint instead of the custom multi-page icon.

PR #3948, opened 2024-10-22
Fix infinite request loop if metadata request fails

### LINKED ISSUE #3812: BUG: Metadata API Request Error Handling
**Describe the bug**
The UI does not utilise a retry limit or backoff for retrying Metadata API requests when the API is not available.

**To Reproduce**
1. Start the UI without an Aleph API available

**Expected behaviour**
The UI should retry a number of times delayed by some form of backoff. 

If a maximum number of retries is reached there should be some button to try again (refresh).

**Aleph version**
Current

**Screenshots**
network-error-spinning

**Additional context**

tillprochaska:
Fixes #3812. Previously, the UI would retry the metadata request indefinitely and without a delay. Now, if the metadata request fails, an error message and a button to reload the page are displayed.

I haven’t implemented automatic retries with back-offs for now, as this adds complexity and I’m not sure it’s necessary.

Screen Shot 2024-10-22 at 11 38 59

### DIFF
--- a/ui/src/app/Router.jsx
+++ b/ui/src/app/Router.jsx
@@ -1,7 +1,8 @@
 import React, { Component, Suspense } from 'react';
 import { Route, Routes } from 'react-router-dom';
 import { connect } from 'react-redux';
-import { Spinner, Classes } from '@blueprintjs/core';
+import { Button, Spinner, Classes } from '@blueprintjs/core';
+import { FormattedMessage, injectIntl } from 'react-intl';
 
 import { fetchMetadata, fetchMessages, dismissMessage } from 'actions';
 import {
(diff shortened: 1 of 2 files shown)


ui/src/app/Router.jsx
```diff
@@ -60,11 +61,11 @@ class Router extends Component {
...
 
-    if (metadata.shouldLoad) {
+    if (metadata.shouldLoad && !metadata.isError) {
       this.props.fetchMetadata();
     }
 
-    if (messages.shouldLoad) {
+    if (messages.shouldLoad && !messages.isError) {
       this.fetchMessages();
     }
```
stchris, line 70: Noob question, but would it be possible to merge these two if clauses? They look identical to me.

stchris: I misread that, sorry. 😞

stchris (approved):
Very nice, I like the pragmatic change with the Reload button.

ui/src/app/Router.jsx
tillprochaska: No problem, the two terms are indeed very similar :)

PR #3951, opened 2024-10-23
Use pg_isready in make upgrade instead of sleep

stchris:
Small improvement, but I need this! No more sleep! 😳

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,10 @@ format-check-ui:
 
 upgrade: build
 	$(COMPOSE) up -d postgres elasticsearch
-	sleep 10
+	# wait for postgres to be available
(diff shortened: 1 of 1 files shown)


tillprochaska (approved):
Nice, that’s great! :)

arp242:
What I did in Alfred is this in the compose.yaml file:

```
  elasticsearch:
    [...]
    healthcheck:
      test:     "exit $( ( curl -s 'http://localhost:9200/_cat/indices' || echo red ) | grep '^red' | wc -l )"
      interval: '5s'
      retries:  10
```

And then `docker compose up --wait`.

Maybe that /_cat/health?h=st is better for this though. But the advantage of putting it in the compose means it works everywhere, rather than just for your upgrade make target.

PR #3996, opened 2024-11-06
Update Prometheus exporter probes so they do not actually generate all metrics

tillprochaska:
Generating all metrics can take multiple seconds for large Aleph instances, which can make the liveness/readiness probe fail. The `name[]` query parameter can be used to select a subset of metrics the exporter should return (or no metrics as all in this case). That way, we can check that the exporter is up and running without actually generating any metrics.


stchris (approved):
Sounds sensible to me. Thanks for figuring it out and fixing it! 🚀

catileptic (approved):

PR #4004, opened 2024-11-20
Prepopulate document search with query string

tillprochaska:
This is a small enhancement and follow-up #3879. It automatically prepopulates the document-scope search input with the query string of the global search. If for example a user searches for "Elizabeth Bennet" and they click on one of the PDF results, the document-scope search input will be prepopulated with "Elizabeth Bennet". If they want to find pages relevant to their search term, they don’t have to retype the search query and can simply press enter.

As an alternative I’ve considered adding a small banner below the tab bar ("Do you want to search for 'Elizabeth Bennet' in this document?") which would probably be more obvious. Decided to try this option first as it adds less clutter to the UI. If it turns out the functionality isn’t clear to users, we can always iterate.


tillprochaska:
Just for future reference, there’s one edge case where this doesn’t work properly:

If you use the ES query string syntax, e.g. `elizabeth bennet properties.mimeType:application/pdf`, the document search input will be prefilled with `elizabet bennnet properties.mimeType:application/pdf` and that will always return zero results.

However, before attempting to reliably parse ES query syntax on the client side, I think it’s best to see if this actually annoys users in practice.

catileptic (approved):

stchris (approved):
This looks great, thanks @tillprochaska !

PR #4011, opened 2024-11-25
Always enable bookmarks

arp242:
It's been an "experimental feature" since 2021. And it seems simple enough? So just always enable it.

### DIFF
--- a/aleph/settings.py
+++ b/aleph/settings.py
@@ -301,12 +301,6 @@ class Settings:
         self.XREF_MODEL = env.get("FTM_COMPARE_MODEL", None)
 
         ###############################################################################
-        # Feature flags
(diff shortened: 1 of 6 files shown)


catileptic:
Thanks for opening this PR :). 

The bookmarks feature was never enabled by default and it isn't a widely-used feature in Aleph. Making it a default feature would increase the code complexity for maintainers of self-deployed instances. It would also widen the set of on-by-default features we have to support. 

I would like to keep this PR open for now, because good work went into it. We may merge it down the line. For now, we're not seeing a need to make it a core feature.

arp242:
How does removing code "increase code complexity?"

It's just one more thing I would rather not do for Alfred, it makes merging the ui from Aleph easier, and I don't see how it burdens Aleph.

PR #4021, opened 2024-11-28
House cleaning

tillprochaska:
This PR resolves all ESLint warnings, fixes the formatting issues on develop, and updates the ESLint configuration to fail in case there are warnings. (Currently, ESLint will succeed if there are warnings, but no errors.)


arp242:
Another warning I had to fix is: https://github.com/occrp/alfred/commit/5148af4

I think there's also some more eslint errors/warnings I fixed: https://github.com/occrp/alfred/commit/a2afdda

PR #4101, opened 2025-01-10
Source API_RATE_WINDOW from environment

riotbib:
The API_RATE_WINDOW variable was hardcoded to 15 minutes and not sourced from environment. This commit fixes this bug.


stchris:
Good find. Thanks for your contribution, @riotbib !

PR #4104, opened 2025-01-20
Add additional validation checks

tillprochaska:
These additional checks aren't necessary in theory (and we do already have tests covering these edge cases), but I think they still make sense to make them explicit and to be more resilient against potential future changes.


stchris (approved):
Looks good, always apreciated!

PR #4105, opened 2025-01-20
Do not delete API key to enable rollbacks to previous version.

tillprochaska:
The `api_key` column will be removed in the next version at which point all plaintext API keys will be deleted.


stchris (approved):
👍

PR #4215, opened 2025-05-13
Fix rendering of dataset summaries

tillprochaska:
This fixes two issues with summaries/descriptions of datasets and entity sets:

1. Long summaries weren’t properly truncated, the overflow wasn’t hidden. I introduced this regression in #4057.

3. In listings of datasets and entity sets, Markdown was rendered (including block-level elements such as lists). This is a regression that was introduced a long time ago with a dependency update of `react-markdown`.


stchris (approved):
Thanks for the fix! 🙏

